PR #234, opened 2017-08-29
Fix turning tween

guzzard:
Fixes angle jumping/popping.
Never rotates 180+ degrees, always rotate the shortest path.


luboslenco:
Cool, thanks!

PR #297, opened 2017-10-23
Support characters without animations/actions

guzzard:


luboslenco:
Yay! More love for character traits definitely handy.

PR #300, opened 2017-10-25
Upgrade props

yulbryn:
Added type of props displayed in Blender as: prop(Type). If type is String, auto add quotes in exporter.


luboslenco:
Amazing!

PR #389, opened 2017-12-02
Add flake8 tests in Python 2.7 and 3.6

cclauss:
I have taken this one as far as I can but would need help to complete it.

The four issues that remain should be looked at:
* https://travis-ci.org/armory3d/armory/jobs/310544629#L561-L573


cclauss:
The four issues that remain should be looked at:
* https://travis-ci.org/armory3d/armory/jobs/310544629

luboslenco:
👍 Those `exporter_gp/sdf.py` were taken from the main `exporter.py` as being temporarily unused, will fix!

PR #407, opened 2017-12-07
Implement LUT-based color-grading

Naxela:
3D LUT color grading. Maybe it's only necessary to only have the StringProperty like lensflare in order to not clutter the UI.

Todo: LUT size handling (resolution dependent handling +  UE4/Unity LUT compatibility), currently only 512x512 LUTS.

### DIFF
--- a/Shaders/compositor_pass/compositor_pass.frag.glsl
+++ b/Shaders/compositor_pass/compositor_pass.frag.glsl
@@ -20,6 +20,10 @@ uniform sampler2D gbuffer1;
 uniform sampler2D lensTexture;
 #endif
 
+#ifdef _CLUT
+uniform sampler2D lutTexture;
+#endif
+
 #ifdef _Hist
 uniform sampler2D histogram;
 #endif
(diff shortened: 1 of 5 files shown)


luboslenco:
Yay! I will clean up the way armory passes stuff into the compositor pass, so it's also easier to do this in the future.

luboslenco:
It's awesome.:)
no lut / some random lut
untitled-1

Naxela:
Nice! I also had some ideas for some other effects, though the ones I have in mind will probably need their own passes, but that shouldn't be too much of a problem considering from what I've looked through the new render path system, it's soooo much easier to see through and get an overview of than the old one 👍 - Now, my main concern is only if the UI would become too cluttered - especially if more advanced color-grading stuff are implemented and taking up panel space - Do you know if B.2.8 have managed to make UI's more customizable?

luboslenco:
Oh yeah, excited the new one feels better! Been thinking a lot on how to overhaul it to make it simpler while keeping it fully programmable. Will write a proper doc on it and expose more hooks to easily add custom passes. 

Writing a minimal render is now just:
```hx
var path = new iron.RenderPath();
path.commands = function() {
	path.setTarget(""); // Draw to framebuffer
	path.clearTarget(0xffffffff, 1.0); // Clear color & depth
	path.drawMeshes("mesh"); // Draw all visible meshes
};
iron.RenderPath.setActive(path);
```

Will slowly rework some of the remaining systems like this compositor extending stuff. Overall armory is still quite small and simple despite getting more powerful. 😃

Nothing comes to mind on the b2.8 ui customization. Perhaps a separate 'Armory Color Grading' panel would do?

Issue #497, opened 2018-02-11
RenderPathDeferred.hx drawTranslucency() failed after 7bef6fa


MLHX:
This error happens after 7bef6fa:
(1) Create a scene containing object using Armory PBR node with opacity <1.0
(2) Tried to run it, failed with the following error:
![screenshot from 2018-02-11 09-37-32](https://user-images.githubusercontent.com/17811339/36068406-d604df42-0f0f-11e8-94ad-86cb3bde91dd.png)

/Milo

luboslenco:
Fix merged!

PR #505, opened 2018-02-13
Fix for #497

Naxela:
Fix for [https://github.com/armory3d/armory/issues/497
](https://github.com/armory3d/armory/issues/497
)Opacity should work now with the fix


luboslenco:
👍

PR #517, opened 2018-02-18
Move on local axis node

Simonrazer:
I hope I did everything right, would be a first if I managed it on first try 🙂 
Trying to add this node to main armory, since it is really necessary and annoying to live without. Tell me if I should try to add some other nodes from logic_pack, thinking about "Inverse" and "To Bool", since they are general purpose and really handy to have.


Simonrazer:
Good dangit one space to much

luboslenco:
Cool, thanks! Will also add "Inverse" and "To Bool" then.

Issue #390, opened 2017-12-02
[Feature Request] "and" Node


Simonrazer:
a "and" node would be really handy, I think about a node in which you plug in 2/more red inputs, like the "merge" node, only with the difference that it outputs the red thing only when all inputs are active. Currently I am using a workaround in which booleans are set to true or false if a red thing is active or not, and then stack "is true" nodes.

Devination:
I've opened a PR #529 that should give you these features. Check it out 👍

luboslenco:
It's in!

PR #529, opened 2018-02-26
LogicNodes: Compare and Multi-value And/Or Gate

Devination:
Logic Nodes: Added ability to add an arbitrary number of inputs to Gate nodes, but only in the "And" or "Or" mode. This addresses #390.

Also added a Compare Node that can compare 2 inputs and return a bool. (==, !=, <, <=, >, >=)

image

### DIFF
--- a/blender/arm/logicnode/logic_gate.py
+++ b/blender/arm/logicnode/logic_gate.py
@@ -44,7 +44,7 @@ class GateNode(Node, ArmLogicTreeNode):
             row = layout.row(align=True)
             op = row.operator('arm.node_add_input', text='New', icon='PLUS', emboss=True)
             op.node_index = str(id(self))
-            op.socket_type = 'ArmNodeSocketAction'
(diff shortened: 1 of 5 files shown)


blender/arm/logicnode/logic_gate.py
```diff
@@ -3,6 +3,13 @@
 from bpy.types import Node, NodeSocket
 from arm.logicnode.arm_nodes import *
 
+def remove_extra_inputs(self, context):
+    # if not any(p == self.property0 for p in ['Or', 'And']):
+    #     if len(self.inputs) > self.min_inputs:
+    #         #FIXME: str(id(self)) gives us a key error here
```
Devination: I'm not super familiar with Blender's API. If someone more experienced could help figure out how to clean up the extra input sockets when switching to a non-And/Or mode that'd be much appreciated.

luboslenco:
😮 👍 
Thanks! Will try to also check on the `remove_extra_inputs`.

PR #530, opened 2018-02-26
Gate Node: Remove extra inputs when no longer on "Or" or "And"

Devination:
Found the solution easily after a good night's sleep


luboslenco:
Awesome!

PR #548, opened 2018-03-08
FollowCamera: use LateUpdate to avoid sync issues

kevinresol:
Currently this could happen:

1. FollowCamera updates first, picked the old position of target
1. Target update its position then
1. Camera appears lagging one frame behind.

Using LateUpdate should fix it.

### DIFF
--- a/Sources/armory/trait/FollowCamera.hx
+++ b/Sources/armory/trait/FollowCamera.hx
@@ -40,7 +40,7 @@ class FollowCamera extends iron.Trait {
 			}
 		});
 
-		notifyOnUpdate(function() {
+		notifyOnLateUpdate(function() {
 			if(!disabled){
 				if(targetObj != null) {
 					if(lerp){


kevinresol:
But I wonder why this can't be used on a camera itself.
What I understand is that this trait should be a general "follow object" one with some options, e.g lerp, offsets, etc.

kevinresol:
cc @guzzard

guzzard:
@kevinresol, you could of course have the script in the child camera and set the position of the parent object, but I just implemented similar to how @luboslenco did the ArcBall script, so they would work in a similar way.

kevinresol:
What I mean is that this trait is too specific to be bundled in the Armory SDK. What I mean is that this class should be renamed to `FollowObject` and remove the `Cannot be used on CameraObject restriction`. In this case, the "follow camera tutorial" would still work, but at the same time this trait gets generalized and can be used in many other scenarios.

guzzard:
@kevinresol, sure that sounds good 👍

PR #591, opened 2018-04-13
Adding some new nodes and length() to VectorMath node

anadin:
@luboslenco theres a few bits in here which may not be ideal? maybe I need to split this into multiple PR's

I added Length to the VectorMath node (straight forward)

I needed a GetVelocity node which led me a little deeper into rigidbody.hx I saw a comment that it was not compiling, for me on macOS it is fine. On this node I think I should out in a extra check to make sure the rb is valid..

Also a new ApplyImpulseAtLocation node as I saw people asking for it.

Finally a VectorClampToSize node (using it to limit velocity on a rigidbody) BUT it requires updated Vec2,3 and 4 in Iron (will submit seperate PR for that)

Anyway, hope this is OK this way let me know

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.pyc
+*.DS_Store
--- a/Sources/armory/logicnode/ApplyImpulseNodeAtLocation.hx
+++ b/Sources/armory/logicnode/ApplyImpulseNodeAtLocation.hx
@@ -1,0 +1,27 @@
+package armory.logicnode;
+
+import iron.object.Object;
+import iron.math.Vec4;
+import armory.trait.physics.RigidBody;
+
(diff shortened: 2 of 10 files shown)


anadin:
Actually added the Get Velocity node + added null checks and a null return when no physics

luboslenco:
Awesome, thanks a lot!

anadin:
No prob, thanks for fixing my formatting, will try to stick better to it - also, didn't know I could change the vector return type on the rigidbody, good to know. I had originally tried that but then switched to handling with the btvector when it didn't work :)

luboslenco:
Sorry that was my fault on the vector return type, we can support multiple physics engines but I exposed the Bullet vector there.. now it's properly self-contained.:)

PR #598, opened 2018-04-16
Add #if to WasmScript - was breaking iOS builds

anadin:
I got latest this morning and after refreshing the gate node I could not build for iOS, I saw all the other WASM checking where wrapped in #if js


luboslenco:
👍 👍 
Also updated travis to auto-detect this.

PR #601, opened 2018-04-18
Add ApplyForceNodes

anadin:
Think it would be good to add a global/local bool?

Also needs haxebullet PR
https://github.com/armory3d/haxebullet/pull/21

### DIFF
--- a/Sources/armory/logicnode/ApplyForceAtLocationNode.hx
+++ b/Sources/armory/logicnode/ApplyForceAtLocationNode.hx
@@ -1,0 +1,27 @@
+package armory.logicnode;
+
+import iron.object.Object;
+import iron.math.Vec4;
+import armory.trait.physics.RigidBody;
+
+class ApplyForceAtLocationNode extends LogicNode {
+
+	public function new(tree:LogicTree) {
+		super(tree);
+	}
+
+	override function run() {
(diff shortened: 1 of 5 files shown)


luboslenco:
👍 👍 
Should be okay, easier to 'read' the node tree this way.

Simonrazer:
Questions, what is the difference between the "Force" and "Impulse" node in use other than that the "Force" node need ca 55 times higher Force-Value? (which is good, needing to always divide by some amount when using "Apply Impuls" is kinda annoying)

anadin:
Yeah so Applying Force (which appears to be the bullet name for it) is basically adding force

`m_totalForce += force*m_linearFactor;`

So if you are falling with force 10(rounded gravity) and you apply(add) a force of .5, you add .5 to the that force which is being applied to the body so now you falling with a force of 10.5 (then drag etc, does its physics thing) and you will still come to a halt when you hit the ground etc.

Applying an Impulse takes current velocity and adds a force to it. Its a single shot, more explosive thing - especially like when a grenade explodes etc.

Applying force is a gentle push, Applying Impulse is exactly that. I think Applying Force should be called Add Force but its not called that in the BulletPhysics system to kept the naming convention

Issue #602, opened 2018-04-18
Adding Material node to graph causes crash


MadMinstrel:
I've encountered a situation where just adding a Material node to a graph, connected or not, causes the game to crash. Sorry for not providing a more minimal file, but I couldn't figure out what exactly the problem is to isolate it.

To reproduce:
1. Open attached file and press P
2. Right-Click the green field to spawn some blue spheres on it. Notice they scale up slightly on mouse over. This is the way it's meant to work.
3. Now, add a Material node to the "highlight" graph, which should be open in the node editor. No need to connect it to anything.
4. Press P and try to spawn some spheres. Notice the game is frozen and there's an error in the console.

[material_node_fail_2.zip](https://github.com/armory3d/armory/files/1926504/material_node_fail_2.zip)

ubaldy:
When I first open the project, MyCanvas trait on the scene is red. I overcame this by creating an empty canvas for it and running the project. Right clicking does not do anything though. 

I added the Material node and did not get the crash. 

Hope my input helps.

MadMinstrel:
Thanks for checking it out. I'm a silly duck, never noticed the canvas is stored outside the .blend file. Sorry about that. I've updated the file.

ubaldy:
Confirmed. That is strange. I will export it and debug it and see what jumps out :)

ubaldy:
Here is my first dive. The only way I know how to debug this is with exporting to C++. I don't know how to debug Krom or else I would :) and exporting HTML fails complaining that it cannot find a Pen.

In the function MaterialData_obj::parse

HXLINE(  59)			 ::Dynamic raw = ::iron::data::Data_obj::getMaterialRawByName(( (::Array< ::Dynamic>)(format->__Field(HX_("material_datas",31,2a,02,22),hx::paccDynamic)) ),name);

The value of raw after this line is NULL which then gets dereferenced later on causing the exception. This gives me a pretty good idea of where to look in the actual haxe code so I will do that next!

MadMinstrel:
The issue is now fixed. Thank you!
Closing.

PR #605, opened 2018-04-21
Stop Trait Node stopping compilation & Get/Set Gravity Nodes

anadin:
The Trait node is still not actually returning a trait but with this you can still build/play with the node in the logic tree.

Also added Get and Set Gravity Nodes


anadin:
8398900 is a fix for https://github.com/armory3d/armory/issues/602

luboslenco:
Wohoo

Issue #636, opened 2018-05-28
"Clear Parent" node does not work properly


donalffons:
Here is a test scene:
![image](https://user-images.githubusercontent.com/33040347/40631730-9c96ab4a-62df-11e8-868f-25931519533d.png)

The idea is that the conveyor should move the sphere to the crane. Then the crane moves it back to the conveyor and so on. This does not work as expected with the node setup.

I think I have a fix for that, so I will post a pull request in a second.

[assemblyline.zip](https://github.com/armory3d/armory/files/2046314/assemblyline.zip)

PR #637, opened 2018-05-28
Fixing issue, where "Clear Parent" node breaks execution in successive calls to "Set Parent"

donalffons:
Hi Lubos,

see this issue: https://github.com/armory3d/armory/issues/636

this simple fix solves the problem in this case. Let me know if this fix works in your opinion. And once again, thanks for your great work!


luboslenco:
Oho sweet. Thank you for the fix! 👌

PR #663, opened 2018-06-08
adjusted ObjectNode to work with SetVariable Node

zaethan:
The objectName was never ever set anywhere so the line
`value = objectName != "" ? iron.Scene.active.getChild(objectName) : tree.object;
`
would always return tree.object.
Thus using the set method had no effect at all as it set the value, which was then overridden because objectName was not set.
One might also consider to remove the value variable completely as I don't see it being used anywhere in thi script.

### DIFF
--- a/Sources/armory/logicnode/ObjectNode.hx
+++ b/Sources/armory/logicnode/ObjectNode.hx
@@ -12,7 +12,7 @@ class ObjectNode extends LogicNode {
 		super(tree);
 	}
 
-	override function get(from:Int):Dynamic { 
+	override function get(from:Int):Dynamic {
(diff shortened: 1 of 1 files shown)


luboslenco:
Thanks! Not sure I follow on the rest, `objectName` is set when the node is constructed.

zaethan:
I just looked up default values in function parameters in haxe, I now get that the constructor actually sets the ´objectName´ if it is given. My bad, I asssumed it would always be set to "".
The last sentence was about the value variable always getting overridden before actually being used, but I've seen that you have taken care of that too now.

luboslenco:
Aha, sounds good then! 👍

PR #766, opened 2018-07-11
add send object event node

KTakahiro1729:
This is a pull request for an addirion for "send object event" node. This will send a event to a specific object.

Compared with send event node, this will send events to a different object.
Compared with send global event node, this will not corrupt the "Global namespace".

sample file: 
[send_obj_node.zip](https://github.com/armory3d/armory/files/2186152/send_obj_node.zip)


luboslenco:
Thanks!

PR #780, opened 2018-07-21
Activate process on play on macOS

poke1024:
On macOS, when you press "Play" in Blender, the Krom process will start, but stay in the background, which is annoying. This PR activates the process, bringing it to the front, so that when you press "Play" you get the player process as front window.

### DIFF
--- a/blender/arm/mac_activate_proc.py
+++ b/blender/arm/mac_activate_proc.py
@@ -1,0 +1,8 @@
+#!/usr/bin/python
+
+import sys
+from Cocoa import NSRunningApplication, NSApplicationActivateIgnoringOtherApps
+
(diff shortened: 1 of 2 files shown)


luboslenco:
Thanks! Perhaps we would be able to put this right into the window creation process? The window on macOS is created along these lines:
https://github.com/Kode/Kore/blob/master/Backends/System/macOS/Sources/Kore/System.mm#L110

poke1024:
Opened a PR in Kore: https://github.com/Kode/Kore/pull/290. Closing this.

luboslenco:
That is awesome! Will include latest Kroms in armory 0.5.

PR #812, opened 2018-08-12
Changed HueSaturationValue node to work with all of its inputs

zaethan:
This pull request enables the HueSaturationValue Node to function properly based on its color and factor inputs. Previously it only used its hue, saturation and value inputs to generate a color. It was not possible to use it to modify an incoming color or texture (at least in my testing).

To achieve this, this pull request adds a new function in `cycles_functions.py` to convert an incoming rgb color to hsv, apply the modifications of the hsv node and return the result as an rgb color, mixed with the original color as defined by the factor input. It also changes the `cycles.py` file to read these factor/color inputs (they were already in place but commented out) and then call the new function.
I also added some comments to the function. If you don't like that, feel free to remove these comments.

Demonstration:

The setup contains two plains. The right plane has a very simple material. It just connects an image texture to a diffuse shader which then connects to the surface. The left plane has almost the same material, it only has an additional HueSaturationValue node between the texture and the diffuse shader.

Right Material: 
matright

Left Material: 
matleft

I just played around with the input values a little. The sample texture I used was 4 pixels I colored in gimp.

Before:
before

After:
after_changes

### DIFF
--- a/blender/arm/material/cycles.py
+++ b/blender/arm/material/cycles.py
@@ -571,13 +571,16 @@ def parse_vector(node, socket):
         return 'pow({0}, vec3({1}))'.format(out_col, gamma)
 
     elif node.type == 'HUE_SAT':
+        curshader.add_function(c_functions.str_rgb_to_hsv)
         curshader.add_function(c_functions.str_hsv_to_rgb)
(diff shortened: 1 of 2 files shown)


luboslenco:
Yay!

PR #815, opened 2018-08-13
Lerp Node

Simonrazer:
Added a lerp node, as requested by Monte_Drebenstedt.

### DIFF
--- a/Sources/armory/logicnode/LerpVectorsNode.hx
+++ b/Sources/armory/logicnode/LerpVectorsNode.hx
@@ -1,0 +1,23 @@
+package armory.logicnode;
+
+import iron.math.Vec4;
+
+class LerpVectorsNode extends LogicNode {
+
+	var v = new Vec4();
+
+	public function new(tree:LogicTree) {
+		super(tree);
+	}
+
+	override function run() {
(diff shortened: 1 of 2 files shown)


luboslenco:
Yo, is there a difference compared to `Vector Mix` node?

Simonrazer:
@luboslenco if you know how to use the Vector Mix node properly maybe not, but I can't figure it out and it seems like [others don't too](http://forums.armory3d.org/t/call-for-a-lerp-node-please/1629). This new one is way simpler constructed and to use, but basically has no options.

luboslenco:
Hm I see. If the `Vector Mix` node is confusing we should rework it (or throw it away), so we do not end up with duplicated functionality.

Is it the naming of sockets in `Vector Mix` which is not good? Or is the functionality broken? It should mimic the existing `Mix` node in Cycles and is set to linear interpolation/lerp by default.

These should be fully equal:

capture

capture1

luboslenco:
A little cleaner now, just shows the `Linear` by default.
capture

Simonrazer:
Ok, I now understand how to use it but your first two pictures confuse me, how does it work using "Time" and not "Delta Time"? Or are the ports swapped in location?

PR #880, opened 2018-09-17
Update SoftBody.hx

Amir-Arsalan:
Improve soft body simulation quality


luboslenco:
We need to expose properties for this, because this way everyone who is 'ok' with the current setting will get slower performance. Maybe binding it to the `Rigid Body World - Solver Iterations` in Blender would do.

luboslenco:
`Rigid Body World - Solver Iterations` property is now exported which controls this. https://github.com/armory3d/armory/commit/9c9a289aea01ed812614847ab7f6ef9839026882

We could also hook `Cloth - Quality` property eventually.:)

PR #969, opened 2018-10-20
Modding Workflow

### LINKED ISSUE #748: Add Modding Support
I know this is kind of a large topic, but it would be awesome if armory could support a modding workflow that would allow you to create blends and build them as a mod that could be loaded into an Armory game.

I'm assuming that assets could very easily be packed, just like they already are, and then they could be loaded dynamically into a game, but I don't know how you would handle mod code. You would probably have to be able to write Haxe code and build it to native libraries ( like .so, .dll ), or use an extension language ( probably better, or at least easier ) like Lua. Lua could probably be integrated similar to how Bullet physics were integrated, with a Haxe library wrapping both a native C and and a ported Javascript library.

Another idea for the extension language could be to bundle V8 for evaluating Javascript on the non web/krom targets. That could be cool too. :slightly_smiling_face:

EDIT: Web assembly is yet another option I forgot about. Because Javascript and web assembly are already supported on the web/krom platforms, it might be better to try to support it on the native targets than it would be to try to support something like Lua on the web targets.

zicklag:
A working implementation of modding for Armory! See [forum topic](http://forums.armory3d.org/t/im-adding-modding-to-armory/2031?u=zicklag).

Requires corresponding changes in armory3d/iron#49.

Fixes #748.

## Known Issues
* You cannot call `trace()` directly in mods. You have to call `Log.trace()`. I don't know exactly why this is. I will look into fixing it some time.
* When using the Forward renderer, objects spawned from mods render shadeless. I don't know much about the RenderPath system so I don't know how to go about fixing this one. It might just go away once Armory 0.6 gets more stable.

### DIFF
--- a/Sources/armory/logicnode/GetTraitNode.hx
+++ b/Sources/armory/logicnode/GetTraitNode.hx
@@ -16,8 +16,12 @@ class GetTraitNode extends LogicNode {
 
 		if (object == null) return null;
 
+		if (cname == null) cname = cast Type.resolveClass(name);
+		// When building mods the Main class doesn't exist
+		#if !arm_modding_mod
 		if (cname == null) cname = cast Type.resolveClass(Main.projectPackage + "." + name);
 		if (cname == null) cname = cast Type.resolveClass(Main.projectPackage + ".node." + name);
+		#end
 		return object.getTrait(cname);
 	}
 }
--- a/Sources/armory/system/Modding.hx
+++ b/Sources/armory/system/Modding.hx
@@ -18,8 +18,9 @@ class Modding {
      * 
      * @param modListFile The path to the file containing the mod list
      * @param done Callback for when mods are finished
+	 * @param onload Callback called for each mod loaded
      */
-    static public function loadModsFromListFile(modListFile:String, done:() -> Void) {
+    static public function loadModsFromListFile(modListFile:String, done:() -> Void, onLoad:(modName:String) -> Void = null) {
 		iron.data.Data.getBlob(modListFile, (blob:kha.Blob) -> {
 			var modList = blob.toString().split("\n");
 			for (mod in modList) {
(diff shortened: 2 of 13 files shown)


zicklag:
@luboslenco Is there anything that you would like me to clarify or update with this? If you just haven't had the time to test it, that's fine, I know you're busy doing awesome stuff. :wink:

I have been running these changes with the latest version of Armory for the last month and it doen't seem to break anything else in Armory. I have an [example](https://github.com/katharostech/armory_examples/tree/modding-example/modding) that you can use to test the setup.

Some things may still need to be implemented, like speakers and some other special types of objects, but the basic functionality works and it doesn't ruin anything that isn't loaded as a mod. I also need a sys implementation in Krom before I can load the mod list dynamically, but hopefully I will find some time to attempt that in the near future and it is not essential.

It is not necessarily urgent that this get merged in, but if it *is* something that you are good with going in, I just want to make sure that it gets merged before `iron.data.Scene` gets refactored or something and I end up having to rework to account for the changes. :slightly_smiling_face:

nanjizal:
zicklag

In regard to sys support for loading there is c++ sys code and krom equivalent in the zui filebrowser code, would that provide some of the functionality maybe?
   https://github.com/armory3d/zui/blob/master/Sources/zui/Ext.hx#L115
You can see demo here of his filebrowser running here.
   https://github.com/armory3d/zui/blob/master/examples/FileBrowser/Sources/Elements.hx
Or maybe you need more sys stuff, but thought would reference incase.

zicklag:
@nanjizal Thanks for the tip. :slightly_smiling_face: It looks like we might be getting a sys implementation soon. ( see this [topic](http://forum.kode.tech/topic/228/is-there-a-way-to-list-folders-and-files-in-krom/8) ).

luboslenco:
Sorry it did not get in in the end.

Maybe helpful / related:
- Using wasm: https://github.com/armory3d/armory/wiki/wasm
- Using js: https://github.com/armory3d/armory/wiki/js
- ArmorPaint plugins: https://github.com/armory3d/armorpaint/blob/master/Sources/arm/Plugin.hx

PR #977, opened 2018-10-24
Fixes for 2.8 addon, interface, and exporter

TriVoxel:
Fixed icons for new 2.8 version with icon updates. Fixed an issue preventing logic nodes from loading. Fixed exporter issue. Fixed addon not loading. Fixed all `.layers` to `.view_layers`.


luboslenco:
Included now, thanks for hints!

PR #982, opened 2018-10-29
Allow Arbitrary Category Names for Nodes

zicklag:
Logic node categories are now created dynamically. This allows nodes be added to any arbitrarily named categories without having to maintain a global category list. This is useful because it allows custom logic nodes to be added to custom categories.


luboslenco:
Awesome!

PR #992, opened 2018-11-01
Make the Output of the Look At Node Euler Rotation

zicklag:
The Look At node previously output a quat for the rotation, which is
inconsistent and incompatible with the other logic nodes such as
Set Rotation.


luboslenco:
Thanks once again!

PR #995, opened 2018-11-04
Add Functions and Function Calls to Logic Nodes

zicklag:
* Renamed Call Haxe node to Call Function because it will now be able
to call node functions as well.
 * Added ability to pass arguments to the Call Function node.
 * Created Function and Function output nodes that allow you to create
functions in node trees. Node functions can be called from other nodes
and from Haxe.

**Calling a Function:**
image

**Defining a Function:**
image


luboslenco:
👍 Ideally we would turn functions into node groups, any thoughts on that? There already is a barebones `Call Node Group` node, which could maybe be coupled with this.

PR #1012, opened 2018-11-18
Add a "From Scene" Option to the SpawnObject Node

zicklag:
Allow spawning objects from other scenes with the SpawnObject node. This requires armory3d/iron#50.

### DIFF
--- a/Sources/armory/logicnode/SpawnObjectNode.hx
+++ b/Sources/armory/logicnode/SpawnObjectNode.hx
@@ -23,6 +23,7 @@ class SpawnObjectNode extends LogicNode {
 		var m:Mat4 = inputs[2].get();
 		matrices.push(m != null ? m.clone() : null);
 		var spawnChildren:Bool = inputs.length > 3 ? inputs[3].get() : true; // TODO
+		var sceneName:String = inputs[4].get();
 
 		iron.Scene.active.spawnObject(objectName, null, function(o:Object) {
 			object = o;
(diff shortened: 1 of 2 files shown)


luboslenco:
Thanks! I am hoping we can solve this using the collections feature in b2.8, so we don't have to parse a new scene file when spawning an object. Will take a look at it asap as we migrate to b2.8 beta (hopefully that begins next week).

zicklag:
Are we going to still support b2.79 when b2.8 comes out?

luboslenco:
We can discuss together on forums/discord once beta is out to evaluate it. But ideally we drop b2.79 asap to develop faster. 🙂

Bugsbane:
Any word on this now that the 0.6 beta with Blender 2.8 is out?

jeske:
Spawning from collections is not the same feature. 

My app really needs the ability to load assets from separate files at runtim3.. as there are community developed and distributed assets.

zicklag:
That was actually the reason that I did this, it was to support my modding integration effort. @luboslenco I think it is still valid to be able to spawn from other scenes, and this is tiny change anyway, so it shoudln't be a big deal, right?

luboslenco:
I have been thinking about this for a long time but still fear the perf issues down the road, sorry. Nonetheless in cases where this is necessary:
- I have added a new [example](https://github.com/armory3d/armory_examples/tree/master/spawn_from_scene) showing how to manually load the scene file and then spawn single object. (use latest git armory)
- To spawn all scene objects into existing scene, [Scene.addScene()](https://github.com/armory3d/iron/blob/master/Sources/iron/Scene.hx#L296) can be used, logic nodes example [here](https://github.com/armory3d/armory_examples/blob/master/logic_examples/scenes.blend).

This should give full control over scene handling. Whenever possible, pre-reference objects in the scene.

PR #1024, opened 2018-11-26
Update to support rigid body that need special compound shape handling 

danilko:
Update to support rigid body that need special compound shape handling 
Example include vehicle body

It is a compond rigid body, but as it is not a supported rigid body type and need special physics update (through btVehicle).

Need to modify rigidbody as aromry's physic world contacts are store using armory RigidBody list, if the ojbect is not an Armory RigidBody, then RigidBody.contacts will not get them if through they are contacted.

Following is one way vehicle body can be added as rigidbody

```
	function createRigidBody(mass:Float, shape:BtCompoundShapePointer):BtRigidBodyPointer {
		var localInertia = BtVector3.create(0, 0, 0);
		shape.calculateLocalInertia(mass, localInertia);

		var centerOfMassOffset = BtTransform.create();
		centerOfMassOffset.setIdentity();

		var startTransform = BtTransform.create();
		startTransform.setIdentity();
		startTransform.setOrigin(BtVector3.create(transform.loc.x, transform.loc.y, transform.loc.z));
		startTransform.setRotation(BtQuaternion.create(transform.rot.x, transform.rot.y, transform.rot.z, transform.rot.w));

		var myMotionState = BtDefaultMotionState.create(startTransform, centerOfMassOffset);
		var cInfo = BtRigidBodyConstructionInfo.create(mass, myMotionState, shape, localInertia);

		var body = BtRigidBody.create(cInfo);
		
		body.setLinearVelocity(BtVector3.create(0, 0, 0));
		body.setAngularVelocity(BtVector3.create(0, 0, 0));

		//physics.world.addRigidBody(body);

		var rigidBody = new armory.trait.physics.bullet.RigidBody(1.0, 8, 0.5, 0.0, 0.0,
						0.04, 0.1, false,
						null, null,1, false, null, false, true, body);
		rigidBody.object = this.object;
		rigidBody.init();
		
		return rigidBody.body;
	}
```

Instead of just return the BtRigidBody pointer, with this patch, one can then above to pass in the created rigidBody and set the object as the object itself to get the controller through the btRigidBody with benefit of Armory RigidBody and rest of integration.

One reason above is necessary is for example, if there is a damage calculation such as something is throwing at the vehicle and vehicle need rigid body to calculate the damage, or vehicle passed through the finished line physic block.

There are only few lines change, the most major block is because of intend as they are in the if statement

Result implementation can be found at
https://youtu.be/MNTmtQ_nrE4

with the time attack example.

There are other way to bypass this such as using another rigidbody on top of vehicle, but when test, it is harder to manage, and also had a conflict where the result object is still not the parent vehicle object.

### DIFF
--- a/Sources/armory/trait/physics/bullet/RigidBody.hx
+++ b/Sources/armory/trait/physics/bullet/RigidBody.hx
@@ -31,6 +31,7 @@ class RigidBody extends iron.Trait {
 	var angularFactors:Array<Float>;
 	var deactivationParams:Array<Float>;
 	var ccd = false; // Continuous collision detection
+	var alternatePhysicUpdate = false;
 	public var group = 1;
 	public var trigger = false;
 	var bodyScaleX:Float; // Transform scale at creation time
@@ -61,7 +62,8 @@ class RigidBody extends iron.Trait {
 	public function new(mass = 1.0, shape = Shape.Box, friction = 0.5, restitution = 0.0, collisionMargin = 0.0,
 						linearDamping = 0.04, angularDamping = 0.1, animated = false,
 						linearFactors:Array<Float> = null, angularFactors:Array<Float> = null,
-						group = 1, trigger = false, deactivationParams:Array<Float> = null, ccd = false) {
+						group = 1, trigger = false, deactivationParams:Array<Float> = null, ccd = false,
+						alternatePhysicUpdate = false, body:BtRigidBodyPointer = null) {
 		super();
 
 		if (nullvec) {
@@ -88,6 +90,8 @@ class RigidBody extends iron.Trait {
 		this.trigger = trigger;
 		this.deactivationParams = deactivationParams;
 		this.ccd = ccd;
+		this.body = body;
+		this.alternatePhysicUpdate = alternatePhysicUpdate;
 
 		notifyOnAdd(init);
 	}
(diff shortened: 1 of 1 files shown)


danilko:
Forgot to mention, the other use cases include the need for other traits that has rigidbody (such as KineticController/StaticController(which seem like trigger option)), where they are using rigidbody, but will not have same physics update as normal RigidBody.

The other options is create a `CollisonObjectTrait` and extend `RigidBody` from there so others can also extend. However, it seem trait/extend another trait seem to cause the cast to be problematic as the `PhysicWorld.getContacts` seem to be mostly depend on `RigidBody` object. Not sure if we are able to break the API, if we can, one way is add `getCollisionContacts` and deprecate `getContact` will return `CollisionObjectTriat`, the `rb` will be `col` to track all collision objects (even if they are not trait rigidbody). Then the user will be responsible for upcast the collision object to RigidBody trait to find the necessary rigidbody, or use collisionobject.object.name to find target object.

mistajuliax:
Hi, @danilko, have you check the new version of armory. I think bullet bindings have been updated and some fixes have been made. Is your PR ready to merge or does it need some rework ?? 
Regards. Julien.

danilko:
@mistajuliax ,

Thanks for contacting, after reviewing the latest armory code, it actually includes functions from this patches already (and actually more functions also :) ).

The `BtRigidBody` can now be passed as part of creation in with the master branch (and seem the master also include other useful parameters). I believe with current master code, this PR is no longer needed.

Thanks again for help

danilko:
@mistajuliax ,

Sorry, I confused my fork  with the current master. It seem current master still not support compound shape yet. But may fix seem not a good solution (as it is more like a hack). So will close the PR and find some other approach later.

mistajuliax:
Ok, it sounds good to me. Thanks for reply.

PR #1057, opened 2018-12-20
Allow Multiple Contact Points Between Two Physics Objects

zicklag:
Tell me if there is anything wrong with the way I did this. The only thing that I can think of is, because I removed the `distance < 0` check, that the Get Contacts node, or similar code would have to perform that check themselves. I can update the node do that check if that is a possible issue.

Other than that, I did rename the `ContactPair.nor` to `ContactPair.normOnB` because it was more intuitive, which would require a change for any code depending on it. The only class that appears to use it is the `PhysicsBreak` trait, but other API changes have broken that class in the dev branch and I didn't know how to fix them.

### DIFF
--- a/Sources/armory/trait/physics/bullet/PhysicsWorld.hx
+++ b/Sources/armory/trait/physics/bullet/PhysicsWorld.hx
@@ -13,8 +13,9 @@ class ContactPair {
 	public var b:Int;
 	public var posA:Vec4;
 	public var posB:Vec4;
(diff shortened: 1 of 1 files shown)


luboslenco:
Cool! What does the removed `distance < 0` check implies? Do we want contacts with distance above 0 in some cases?

luboslenco:
Thanks! 🙂

PR #1060, opened 2018-12-21
Fix Node Trait Selection Issue

### LINKED ISSUE #1059: Blender 2.8: The Fake User Indication "F" is Getting Inserted In Node Tree Names and Breaking Build
When selecting Node traits in the Traits panel, the "F" for the fake user indication gets put in the Node tree name and the Armory build fails because it can't find the class name. I can't tell, but this might actually be a Blender 2.8 bug.

image

zicklag:
fixes #1059

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -2496,9 +2496,9 @@ class ArmoryExporter:
                 if t.enabled_prop == False:
                     continue
                 x = {}
-                if t.type_prop == 'Logic Nodes' and t.nodes_name_prop != '':
+                if t.type_prop == 'Logic Nodes' and t.node_tree_prop != None and t.node_tree_prop.name != '':
                     x['type'] = 'Script'
-                    group_name = arm.utils.safesrc(t.nodes_name_prop[0].upper() + t.nodes_name_prop[1:])
+                    group_name = arm.utils.safesrc(t.node_tree_prop.name[0].upper() + t.node_tree_prop.name[1:])
                     x['class_name'] = arm.utils.safestr(bpy.data.worlds['Arm'].arm_project_package) + '.node.' + group_name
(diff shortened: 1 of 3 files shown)


zicklag:
Because this PR changes the data name for the NodeTree references in the trait panel, it will break games written in older versions of Armory. I added a small bit of logic that can pull auto-set the node references again from the previous blend, but it only does it when the trait panel is rendered, which means that you have to click on every object that has a Node trait and view the traits panel before it will update the traits. I'll see if maybe there is a better way to do that.

blender/arm/exporter.py
```diff
@@ -2496,9 +2496,9 @@ def export_traits(self, bobject, o):
                 if t.enabled_prop == False:
                     continue
                 x = {}
-                if t.type_prop == 'Logic Nodes' and t.nodes_name_prop != '':
+                if t.type_prop == 'Logic Nodes' and t.node_tree_prop != None and t.node_tree_prop.name != '':
```
zicklag: If we want to allow for seamless upgrade from earlier versions of armory we could reference `t.name` here instead of `t.node_tree_prop.name`. Earlier versions of Armory have the `name` property which is set from the node tree name, so it will work on old and new versions without any user interaction. It is up to you @luboslenco what you want to do with that.

zicklag:
If we want to keep compatibility with older versions of Armory check my [comment](https://github.com/armory3d/armory/pull/1060/files). I found an easy way to do it, if it isn't against what we want to do for code design in general.

zicklag:
I just found out this PR has a neat bonus. If you rename a node tree, now you will not have to go to the objects that use that trait and re-select it in the trait panel. The object will continue to reference the node tree even when it is renamed.

I didn't realize this would help with that. :laughing:

luboslenco:
Yay, sounds cool!

PR #1064, opened 2018-12-23
Changed user_preferences to preferences to adapt with latest 2.8 api changes

BlackGoku36:

PR #1070, opened 2018-12-26
Update armory.py

Amir-Arsalan:
Stop crashing when starting up Armory in Blender 2.8


luboslenco:
Linking to https://github.com/armory3d/armory/pull/1064.

PR #1071, opened 2018-12-26
Update utils.py

Amir-Arsalan:
Stop crashing when starting up Armory in Blender 2.8 as `user_preferences` has been replaced with `preferences` in Blender 2.8.

### DIFF
--- a/blender/arm/utils.py
+++ b/blender/arm/utils.py
@@ -12,6 +12,11 @@ import arm.lib.armpack
 import arm.make_state as state
 import arm.log as log
 
+if blenderVersion.find('2.7') != -1:
+    user_preferences = bpy.context.user_preferences
+elif blenderVersion.find('2.8') != -1:
+    user_preferences = bpy.context.preferences
+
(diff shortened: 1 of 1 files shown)


BlackGoku36:
@Amir-Arsalan I think armory plan to support only 2.8 now and 2.79 is going to be ditched.
there is already PR opened here -> https://github.com/armory3d/armory/pull/1064
Your PR would not work, because the reason is that all python files are changed so that it is adapted to 2.8 and these changes would not work in 2.79, you can try this repo in 2.79 and it would not work.

Amir-Arsalan:
@BlackGoku36 Oh right ... . I didn't notice your PR; sorry about that. However, I think Blender 2.79 should still be supported by Armory as Blender 2.8 is still in Beta mode.

luboslenco:
Linking to https://github.com/armory3d/armory/pull/1064.

Yes, 0.6 is coded for b2.8. The reason is that supporting both at once turned out to be very problematic & time consuming given our resources.

PR #1089, opened 2019-01-07
Fix Armory Bake Python Error

zicklag:
From forum topic:
http://forums.armory3d.org/t/cnt-bake-traceback-error-comes-always/2451?u=zicklag

In [this](https://github.com/armory3d/armory/compare/master...katharostech:bug_armory-bake?diff=split&expand=1#diff-6c28ad04114826c2ba032cdea128c144L1037) line I removed the scene context for the Armory bake object search, which means that objects in the current or any scene will be included in the object search. I did this because there seems to be some kind of bug where you can't select the object from the dropdown when you use the `bpy.context.scene` as the search object.


luboslenco:
👍 👍

PR #1090, opened 2019-01-08
Update Make Proxy script for Blender 2.8

Leonix:
Hey!

I noticed Make Proxy button fails with a stack trace in armory 0.6 alpha (archive downloaded from site) in `armsdk/armory/blender/arm/proxy.py` on line 17.

Looks like `bpy.context.scene.objects.active` is [no longer available](https://docs.blender.org/api/blender2.8/bpy.types.SceneObjects.html#bpy.types.SceneObjects) in Blender 2.8

After a quick search I found people use `context.view_layer.objects.active` [for the same purpose](https://blender.stackexchange.com/questions/126577/blender-2-8-api-python-set-active-object).

I never did extensive testing or research, but this simple fix seems to work for me.

### DIFF
--- a/blender/arm/proxy.py
+++ b/blender/arm/proxy.py
@@ -14,16 +14,16 @@ def traverse(obj, is_parent=False):
 
     override = bpy.context.copy()
     override['object'] = obj
-    bpy.context.scene.objects.active = obj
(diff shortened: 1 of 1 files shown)


luboslenco:
Thanks, lets do it! 🙂

PR #1092, opened 2019-01-09
Add Migration Notes

zicklag:
Added a couple of notes about migration steps when upgrading to Armory 0.6.


luboslenco:
Logic node traits should get auto-updated in https://github.com/armory3d/armory/commit/103c30704d086f900a85ab544ea6a4fb05f7a14b, will look into bake panel.:)

zicklag:
Oh, cool. If I make other changes like that in PR's I'll try to include upgrade logic like you did there. There should be no need for these notes then. :+1:

PR #1105, opened 2019-01-17
Get Contacts Node Returns Each Contact Once

zicklag:
Make sure that the Get Contacts node doesn't return the same object multiple
times.

**Edit:** Actually wait to merge this. I just realized that I should be putting the "don't return same object twice" logic in the PhysicsWorld, not in the logic node.


zicklag:
OK this is ready now.

luboslenco:
🎉

PR #1115, opened 2019-01-23
Add Late Update and Physics Pre-Update To The On Update Node

zicklag:
I added a dropdown to the On Update node to let you select between On Update, On Late Update, and On Physics Pre-Update. If you would rather these be separate nodes, just tell me and I'll split it out and update the PR.


luboslenco:
Many thanks!

zicklag:
Glad I can help!

PR #1124, opened 2019-01-27
Merge the Keyboard and Mouse nodes

zaethan:
So far, the difference between the yellow and the red output of the __On Keyboard/Mouse__ and __Keyboard/Mouse State__ nodes has created quite a lot of confusion especially for new users.

So this is why I would suggest to join these almost identical nodes into a single one (simply called keyboard/mouse) with two outputs.
image

One could take this a step further and merge the mouse coords node as well, but since it is easy to tell the difference between that node and the other mouse related nodes, it is probably not necessary.

The same could be done with the surface, gamepad and virtual button nodes.

I also did not perform a check on the (from:Int) in the get method, since there is only one information to get.

Would it be possible to check whether something is connected to an output in haxe? In that case, tree.notifyOnUpdate(update) in the constructor could be called only when something is connected to the run output, which would save a few cycles every frame if it is not used (especially if you have a lot of these nodes and only use their state output).

One last thing to think about is compatibility with existing node trees. As of this PR, the "new" nodes extend the Keyboard/Mouse State node with a run output. This will probably cause existing nodes to try running the run method of whatever they are connected to. Ideally the __On Keyboard/Mouse__ nodes would be removed, since they are no longer needed, but this would break existing trees.


luboslenco:
Let's do this for all the input nodes then, would you mind including gamepad and others here?

> Would it be possible to check whether something is connected to an output in haxe?

Will check on that, we should do that.:)

zaethan:
So, I merged the gamepad, surface and virtual nodes aswell.
I was only able to test the gamepad node though, since I don't know about publishing to android for testing surface and I don't know exactly how the virtual input is working. However they did not lead to a build failure when used in this way:
image

From the above setup, I got the following trace messages in that order:
* \<function\>
* false
* 0

So the first one is coming from surface... A function? Looking at the code, I noticed that sometimes it uses surface.down() and surface.down without the paranthesis. I guess this is a typo? At least I couldn't make out any specific pattern.

The second one is false, that sounds good.

And the third one is 0 (or 1 if I press the button). This is most likely to take the analog triggers into consideration which is totally fine. One might argue however that having a simple boolean output would be a better idea because it would work exactly the same like the other input nodes (although this could be easily achieved by comparing the outputted number with some low threshold using some other node). And then have an additional output for analog readings or move the analog readings to the gamepad coords node.
(Interestingly some of the buttons don't work for me at all or are mixed up (L1 is the left button, L2 is the back button and L3 the right sticks click), but that might be because I'm running on Linux and something is not working right, or because I use an Xbox360 Controller, I will look into it later).

zaethan:
Okay, I will split it in two sockets then.

We could also create new nodes  instead of replacing the old ones.
The old ones would then be marked as deprecated (with a hint to use the new one) and removed after some time. This way everyone would have time to change their trees themselves. It would also be possible to make the old nodes output a message to the console telling the developer to replace them with the a one.
This would not break existing trees... if the developer updates in time.
While this is not a pretty approach to this, it is a simple one.
I don't know that much about Blender's Python API, so I don't know how complex such an auto upgrade procedure would be. However the ability to change nodes automatically upon deprecation would probably come in handy in the future, if, at any point in time, other nodes are being deprecated as well.

zaethan:
tl;dr: I'm confused about gamepad button mappings.

So, I did split it up like this a few days ago
image

But before pushing I wanted to make shure that it was working as expected... which it did not.
So I tried to find out the button/axis mapping of the controller by creating a new trait, which just traces the buttonsDown Array (had to make it public first) and the sticks.
* The higher array slots are never being activated, no matter what I press (XB360 on Linux)
* The left Stick works fine
* Only one axis of the right sticks actually works, the other axis is controlled by one of the analog triggers
* The other trigger does not do anything
* The DPad does not work at all

I thought it might be my setup (although everything is working fine in dolphin/cemu emulator for example), so I tried html5gamepad.com, which you suggested in this thread http://forums.armory3d.org/t/my-gamepad-cannot-work-in-armory/1310/6
And it is working just fine...
Is it only me, or is the gamepad input broken?

luboslenco:
Hm is it behaving the same when you run your armory test in a browser?

zaethan:
... Wow, it's perfect in a browser. Even the analog outputs I added work just fine.

I also tested the Linux C target and get the same weird behaviour I get in Krom. So it seems only html5 is working for gamepads right now...

zaethan:
I also wanted to add "back" and "start" to "share" and "options", but those are specific to the xbox 360 controller... on the xbox one controller those are called "view" and "menu"... or + and - on nintendo controllers... And the PS3 controller calls them select/start I think...
It's a little annoying that these names differ that much... 
Thus, the nodes can only correctly represent one controller type (PS4 right now), or have very long property names... When using Haxe scripts it is the same I guess, since the gamepad.down function expects a string parameter...

Another idea would be names like "Button 1" (and so on) combined with a reference table somewhere in the armory documentation...

zaethan:
Interesting... The analog triggers as well as the dpad is handled as axes on linux, while they appear to be handled as buttons on windows.
html5gamepad shows 4 axes in windows and 8 axes in windows. This makes cross platform input handling a little difficult. I wonder how the html5 target is handling this, because it seemed to work perfectly fine on Linux, and I assume it works on windows as well, which would make it suitable for cross platform uses.

zaethan:
I am currently developing an automatic node replacer. It is currently an operator, which can be invoked with F3 in blender.

It can
* replace nodes (obviously)
* remap inputs
* remap outputs

I told it to replace "OnKeyboard" with a "Time" node and map the first output of the "OnKeyboard" to the second output of the "Time" node.
I also told it to replace the multiplexer with a gravity attractor and remap the inputs.
And to replace bool nodes with float nodes.
This
grafik
turned into 
grafik

(It actually replaces ALL occurrences of these nodes anywhere in the blend file.)
I am planning to clean up the code a little and add some helpful comments before uploading it to the logic pack (since adding operators through it is super easy and does not require modifying armory at all). Or should I put it somewhere else?

zaethan:
I added the replacement code into the file nodes_logic.py.
It seems like that is the place where it belongs.

It defines class which represents a single replacement rule and a replacement map, which maps node id's to replacement rules.
It contains an operator, which checks for every node in every tree, whether it is contained in the replacement map. If it is, the operator then invokes the actual replacement method.
This method adds the new node, rearranges the links and removes the old node.

The idea is that anyone can add replacement rules through the add_replacement() method.
For example,
add_replacement(Replacement("LNOnKeyboardNode", "LNTimeNode", {}, {0: 1}))
add_replacement(Replacement("LNBranchNode", "LNGateNode", {0: 1, 1: 0}, {0: 1, 1: 0}))
adds a replacement for the onKeyboardNode and a replacement for the branch node.

This operator only replaces one node with another node. It cannot replace one node with many nodes.

luboslenco:
Let's take it in! I guess we can now utilize the node replacement code in the auto-update logic at https://github.com/armory3d/armory/blob/master/blender/arm/props.py#L303?

zaethan:
This was merged faster than I expected.
I would recommend either merging my new PR aswell, or undoing the changes in the meantime since this PR makes some nodes "undefined" without replacing them. Replacing only happens with the new PR.

PR #1136, opened 2019-01-31
Added support for script subfolder organisation using haxe package syntax

DiegoBM:
At the moment there is no way of visually splitting your trait files using subfolders directly from Blender, therefore all the haxe trait files will hang from "Sources/arm".

It would be beneficial specially for maintenance if users were able to split their projects into subfolders that map features/scenes or whatever makes more sense for the creator. Therefore this pull request includes functionality that will allow users to put their traits into subfolders by using haxe package syntax (package example.MyClass;)

This way, when the user creates a new haxe trait in blender, if s/he creates a name such as core.MyTrait, a new haxe script will be automatically created in "Sources/arm/core/" with name "MyTrait.hx" containing the correct package syntax "package arm.core;", that will be searchable in blender and will execute properly from the Armory final product.
The same way if the script name is "test.core.Trait", a trait called "Trait.hx" will be created in the folder "Sources/arm/test/core/", and so on. And if no haxe package syntax is used (no dots in the file name), the file will be created in "Sources/arm/" as usual

Trait creation:
new_trait

Searchable:
searchable

Correct source:
correct_source

Correct execution:
correct_execution

Hope it helps.

### DIFF
--- a/blender/arm/props_traits.py
+++ b/blender/arm/props_traits.py
@@ -261,7 +261,8 @@ class ArmEditScriptButton(bpy.types.Operator):
             obj = bpy.context.scene
         item = obj.arm_traitlist[obj.arm_traitlist_index]
         pkg = arm.utils.safestr(bpy.data.worlds['Arm'].arm_project_package)
-        hx_path = arm.utils.get_fp() + '/Sources/' + pkg + '/' + item.class_name_prop + '.hx'
+        # Replace the haxe package syntax with the os-dependent path syntax for opening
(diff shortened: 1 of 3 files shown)


zicklag:
Awesome! This will be nice to have. :+1:

luboslenco:
That's brilliant! Thanks for contributing to armory, another great improvement is in. 🙂

DiegoBM:
Thank you so much @luboslenco, I appreciate it :), I hope it helps everyonw. I have been testing for the last day and it seems to work well, but if you find anything not responding as it should please let me know.

I have also contributed a documentation entry for the feature for your consideration. If you find any typos or any problems with it please also let me know.

PR #1237, opened 2019-03-29
"On Local Axis" option for Translate-, Impulse- and Force- nodes

Simonrazer:
The title says it. The old Translate on Local axis node is now redundant I think, but maybe leave it in if so old project don't need to be changed. I touched it by mistake in this pull request, only added some lines tho so nothing happened.


luboslenco:
Many thanks!

PR #1242, opened 2019-04-01
fix an exception where hitPointWorld gets null.

rbx775:
...due to too short Rays set by the user.
Suggestion for the fix by zicklag, see here: http://forums.armory3d.org/t/bullet-physics-keeps-freezing-the-games/2925/2
Can confirm its working now.


luboslenco:
Yay, thanks!

PR #1243, opened 2019-04-01
Fix issue reading trait properties

greyheron:
Change ArmTraitPropList to ARM_UL_PropList following latest blender API changes to fix error in blender: 
```
search for unknown uilisttype ArmTraitPropList
uiTemplateList: List type ArmTraitPropList not found
```


luboslenco:
Thanks a lot for the fix!

PR #1263, opened 2019-04-30
auto replace properties

zaethan:
(Sorry for spamming the pull requests...)
I added the ability to map properties (which is especially important for the input nodes, since they would be resetted otherwise) and to map unconnected sockets as well.
I also realized that since the bl_idname is different now, I also had to rename the haxe classes.


luboslenco:
Awesome, thanks for all the work on this!

PR #1271, opened 2019-05-09
small fix for rotate around axis node

zaethan:
this seems to fix the issue in http://forums.armory3d.org/t/fly-through-navigation-controller-how-to-adjust-axis-for-forward-and-back-movement/3038/7 for me.
The error which showed up in the web console stated that inputs[2].get().normalize was not a function.
I guess that inputs[2].get().normalize() might have created issues due to a null value...

It is weird though that this only happened when using the html5 exporter... F5 with Krom and Browser worked fine.


luboslenco:
Thanks! Exporter does dead-code-elimination by default to minimize size, may be related.🤔

PR #1290, opened 2019-06-06
Camera Motion Blur Pattern Fix

indokan1945:
When we flip the y coordinate of coord in getVelocity() function, the y of the velocity vector will also get flipped. If we let this on, the result of the motion blur will be weird, creating X pattern blur rather than O pattern blur, noticeable especially when looking down/up and rotating around camera front axis. To fix this, negative the y of velocity vector manually and problem solved.

Before:
armory2

After:
armory1


luboslenco:
Thanks!

Issue #1155, opened 2019-02-08
Get Property returning the property name too


Didiercode:
Just a remark that could reduce number of nodes and help during debuging for example,
that is to add to the Get Property node 
![image](https://user-images.githubusercontent.com/20455592/52478708-848e6200-2ba6-11e9-88a8-88b88ebcd1e0.png)
 an output of the Property name too. 

(this can be done actually by adding a node String with the property name inside, but could reduce number of nodes)
*(drag & drop zipped .blend file here)*

luboslenco:
Closed via https://github.com/armory3d/armory/pull/1300.

PR #1300, opened 2019-06-21
Return property name in Get Property node

MoritzBrueckner:
Implement request from https://github.com/armory3d/armory/issues/1155

screenshot


luboslenco:
Awesome, thanks a lot!

PR #1303, opened 2019-06-23
 Hide is_object property from user during trait script creation

MoritzBrueckner:
This property is used internally only and doesn't need to be exposed to the UI.

Now the UI looks like this:
Screenshot


luboslenco:
👍 👍

PR #1310, opened 2019-06-30
 expand the info raycast returns

N8n5h:
This change will allow to get the position of a hit with raycast, the normal and the body, some extra info that otherwise you couldn't obtain because raycast only returns a body, this may break some compat but I think is better than having multiple functions that do the same logic only for obtaining each piece of information separated (to keep compatibility).


N8n5h:
https://github.com/armory3d/armory_examples/pull/3

luboslenco:
Thanks, added a note about the change at https://github.com/armory3d/armory/commit/d3a3aa022ecc439134065c2719c9116429ee064e.

Issue #1254, opened 2019-04-15
Baked F-Curves don't work


blackno666:
Baked F-Curves for animations don't work.

Create a simple animation: works
Bake F-Curve: won't work

Here's a workflow to reproduce the issue:
![out](https://user-images.githubusercontent.com/47137964/56166711-96d9b180-5fd6-11e9-8fa9-9e299dc4beb6.gif)

This is a pitty since the modifiers and interpolation modes (like elastic our bounce) for F-curves can be used to easily create complex animations.

MoritzBrueckner:
@luboslenco Is there a border case where sampled animations can't be converted to keyframes? If not, why not convert them in `export.py`? For an addon of mine, I did exactly that and it worked.

If you want I can try that out, this would also simplify the code as the distinction between keyframes and sampled data would not be required anymore.

luboslenco:
@MoritzBrueckner if I recall correctly it's from the older days when only sampled animation worked - removing / simplifying code always welcome!

MoritzBrueckner:
So I finally got time for this and it works now, I've even implemented a per-keyframe interpolation like in Blender :) Only the bone animations use the old system as I didn't really understand the way their transformations work. There are a lot of unnamed and uncommented matrices in the code and its very hard to understand what's happening there, but I'm not that into matrix math anyway. But it works and thats important.

The problem is that FCurve modifiers are not included in the animations, as their ouput is neither sampled nor consists of keyframes (modifiers must be baked first). So my idea now is to use the `evaluate(frame)` method of the FCurve class to export each frame (we would then just use this method for everything and export a value per frame, like sampled data). This would require a little bit more storage for the game but saves a very little bit of performance as well. Also, every possible interpolation would be supported that way without implementing them in armory too.

So, before opening a PR: whats your opinion on that, @luboslenco?

luboslenco:
> So my idea now is to use the evaluate(frame) method of the FCurve class to export each frame (we would then just use this method for everything and export a value per frame, like sampled data). This would require a little bit more storage for the game but saves a very little bit of performance as well. Also, every possible interpolation would be supported that way without implementing them in armory too.

Sounds pretty good to me. Thanks for working on that!

PR #1323, opened 2019-07-19
Simplify animation export and support for baked FCurves & modifiers 

MoritzBrueckner:
Linked to https://github.com/armory3d/iron/pull/70

See https://github.com/armory3d/armory/issues/1254

Now, every object transform action is treated as sampled data (no matrices but actual values, saves some storage). This way, we can handle all possible types of interpolation, baked/sampled FCurves and modifiers.

The exported animation now starts at the first keyframe of the action or the first frame of any modifier with restricted frame range and ends following the same principle (but the other way around of course).

There might be room for improvement, but this should make animations export more stable :)
For example, single frames (time) doesn't have to get exported (range only), but this would require a little work on animations in iron.
Also, every FCurve is evaluated for the whole length of the action (even if there are only two frames at the beginning e.g.), this is because iron doesn't like tracks of varying length.


luboslenco:
Wohoo let's do it! Thanks for contributing!

PR #1324, opened 2019-07-21
Improve reroute handling in logic nodes

zaethan:
This PR addresses an issue with reroutes.
In the creating outputs stage of the `build_node` function, it previously checked whether an output was linked and then added the result of `build_node` to the outputs array.
While `build_node` correctly handled reroutes as inputs, this didn't seem to work when collecting the outputs.

This PR introduces a new recursive `collect_outputs_from_link` function, which takes an (outgoing) link and recursively adds all non-reroute nodes to an array, which is then returned.
The create outputs in `build_node` was modified to use these instead of the raw links.

grafik
For testing I hooked up a keyboard node to a print node through two reroutes.
With the original exporter, this was produced:
`_Keyboard.addOutputs([_Keyboard]);`
As can be easily seen, there is no print node, but the keyboard is being added as its own output... I guess this is because of how the reroutes are being handled in the beginning of the `build_node` function:
```
if node.type == 'REROUTE':
        if len(node.inputs) > 0 and len(node.inputs[0].links) > 0:
            return build_node(node.inputs[0].links[0].from_node, f)
        else:
            return None
```
it seems to only work in the "input direction".
Thus, when the create output stage would have called `build_node` on the reroute, it would have just returned the keyboard node...

With this change, this should be produced
`_Keyboard.addOutputs([_Print]);`

Now all of this seems to be working fine:
grafik

There is a Keyboard with an arbitrarily large number of reroutes, some aren't even attached to anything. There is a keyboard with no reroute, a completely unconnected keyboard, and an alternate node with reroutes at one output.
In case of the alternate node, it will run "hello" and "world" at the same time, which is the desired behavior.

One last thing: currently, pressing space will result in "b", "true" and "a" in that order. Thus I assume, that in a chain of reroutes, the nodes attached to the "last" reroute (the one being the furthest away) will be executed first. If we were to invert the list in the for loop (line 184), we could probably guarantee that nodes will be executed in the exact order as they are attached to the reroute chain. While not really necessary, this could be used as an implicit sequence node, to improve readability in some cases.
However, whenever the chain "splits" (into two new chains), there would be no guarantee which of these would be executed first...

### DIFF
--- a/blender/arm/make_logic.py
+++ b/blender/arm/make_logic.py
@@ -181,9 +181,9 @@ def build_node(node, f):
         if out.is_linked:
             out_name = ''
             for l in out.links:
-                n = l.to_node
-                out_name += '[' if len(out_name) == 0 else ', '
-                out_name += build_node(n, f)
+                for output in collect_outputs_from_link(l, f):
+                    out_name += '[' if len(out_name) == 0 else ', '
+                    out_name += output
             out_name += ']'
         # Not linked - create node with default values
         else:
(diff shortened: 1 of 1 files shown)


luboslenco:
Nice, thanks again!
> If we were to invert the list in the for loop (line 184), we could probably guarantee that nodes will be executed in the exact order as they are attached to the reroute chain.

Sounds good to me, would you mind including that in this PR?

zaethan:
I think my initial assumption about guaranteeing a specific execution order was wrong. To be able to guarantee this execution order I thus changed the `collect_outputs_from_link` function to a `collect_nodes_from_output` function. This new function expects an output socket and first appends all non-reroute nodes to a list. After that, it will recursively check the reroute nodes.

I also added a safety check in the input handling to handle unconnected reroutes.

This setup seems to work fine for me (with and without connecting the reroutes in the top):
grafik

PR #1337, opened 2019-08-08
Lens dirt masking + Chromatic Aberration

Naxela:
I've been attempting at making the lens texture a bit more useful by making it usable as a lens dirt filter/mask:

LensDirt1

LensDirt2

Not entirely sure about the UI (text-input reveal), but I tried to make it minimally cluttering, also it might technically fit better together with the bloom shader, but at least it works for now, if you think it fits :)

ui

### DIFF
--- a/Shaders/chromatic_aberration_pass/chromatic_aberration_pass.frag.glsl
+++ b/Shaders/chromatic_aberration_pass/chromatic_aberration_pass.frag.glsl
@@ -1,0 +1,69 @@
+#version 450
+
+#include "compiled.inc"
+
+uniform sampler2D tex;
+
+in vec2 texCoord;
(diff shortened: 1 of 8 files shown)


Naxela:
Added chromatic aberration as well, with two types:
- Simple: Fast and plain
- Spectral: Looks better, but requires a fair bunch of samples to look good

Chromatic1

Chromatic2

Also, it only works with deferred for now, I couldn't get it working with forward for some reason

VicentGJ:
Great!!! It's looking terrific, and it seems to address my concerns about the bloom shader and lens flares not working properly. I hope this gets merged soon.

luboslenco:
@Naxela 😍 super cool, thanks!

PR #1338, opened 2019-08-08
update for navmesh generation

N8n5h:
What has changed:
- added back visualization of navmesh generation
- readded navagents moving in the z axis. I tested for a bit and got good results with that small modification, not really sure why it was commented out but will look at it further if the results are not consistent for now.
- added functions to change navmesh generation settings for recast, used in the interface when generating visualization and when running the project
- moved the navmesh generation button for consistency

image

Depends on:
https://github.com/armory3d/armory_tools/pull/1
https://github.com/armory3d/haxerecast/pull/1

I tried to do some threading to prevent the UI to lock but I couldn't get the importing model part to work with threads so I dropped it for now.
Error catching is missing too, planning to add it in the future with a few other features.


luboslenco:
Wow! 😍

PR #1357, opened 2019-08-27
Add getElements():Array<TElement> function to CanvasScript

greyheron:
Adding getElements() function to CanvasScript, it can be a handy method for this Trait. Sometimes we would need to iterate all elements attached to a Canvas without having to call one by one using getElement(name:String)


luboslenco:
Thanks! 👍

PR #1381, opened 2019-09-11
Add Sublime Text as a supported text editor

MoritzBrueckner:
Linked to https://github.com/armory3d/armsdk/pull/5

**This PR adds support for Sublime Text (and other custom code editors). I have reworked the selection menu in the user preferences and the detection of the standard code editor.**

## What has changed:
- Instead of a folder path, the user must now specify the file path to the executable file in the user preferences, so that user-defined editors are also supported.
- Since Kode Studio is no longer included in armsdk, the methods for automatically searching for this editor inside the SDK folder have been removed
- If the user has defined the environment variables "VISUAL" or "EDITOR", these will be considered first when choosing the default editor.
- If Sublime Text is selected, a simple project file is created.

## What is missing so that the PR is ready to merge:
- Test if Mac ~~and Linux~~(https://github.com/armory3d/armory/pull/1381#issuecomment-534611360) still work after the changes. Unfortunately, I only have Windows, so if someone can test it, it would be great.
- Custom build system for sublime text. So far the project can only be built in Blender with F5 or in Kode Studio. @luboslenco Is there a way to build the project using an external script? Then it could be called from Sublime (and more editors in the future).
- What about the mklink methods in utils.py? Are they still needed or why were they commented out?
- Should the editor-related methods in utils.py be moved to their own script? The file becomes more and more cluttered.
- Should we write a tutorial (in the Wiki?) on how to add new editor configurations? Should the system change to make adding editors even easier?

### DIFF
--- a/blender/arm/props_traits.py
+++ b/blender/arm/props_traits.py
@@ -313,7 +313,7 @@ class ArmoryGenerateNavmeshButton(bpy.types.Operator):
     bl_label = 'Generate Navmesh'
     def execute(self, context):
         obj = context.active_object
-        
+
         if obj.type != 'MESH':
             return{'CANCELLED'}
 
@@ -329,10 +329,10 @@ class ArmoryGenerateNavmeshButton(bpy.types.Operator):
         nav_full_path = arm.utils.get_fp_build() + '/compiled/Assets/navigation'
         if not os.path.exists(nav_full_path):
             os.makedirs(nav_full_path)
-        
+
         nav_mesh_name = 'nav_' + obj.data.name
         mesh_path = nav_full_path + '/' + nav_mesh_name + '.obj'
-        
+
         with open(mesh_path, 'w') as f:
             for v in obj.data.vertices:
                 f.write("v %.4f " % (v.co[0] * obj.scale.x))
@@ -343,10 +343,10 @@ class ArmoryGenerateNavmeshButton(bpy.types.Operator):
                 for i in reversed(p.vertices): # Flipped normals
                     f.write(" %d" % (i + 1))
                 f.write("\n")
-        
+
         fp_build_name = arm.utils.get_fp_build().rsplit('/')[-1]
         nav_mesh_path = '/' + fp_build_name + '/compiled/Assets/navigation/' + nav_mesh_name
-       
+
         buildnavjs_path = arm.utils.get_sdk_path() + '/lib/haxerecast/buildnavjs'
 
         # append config values
@@ -372,7 +372,7 @@ class ArmoryGenerateNavmeshButton(bpy.types.Operator):
         navmesh.rotation_euler = (0,0,0)
         navmesh.location = (obj.location.x,obj.location.y,obj.location.z)
         navmesh.arm_export = False
-        
+
         bpy.context.view_layer.objects.active = navmesh
         bpy.ops.object.editmode_toggle()
         bpy.ops.mesh.select_all(action='SELECT')
--- a/blender/arm/props_traits.py
+++ b/blender/arm/props_traits.py
@@ -252,18 +252,19 @@ class ArmEditScriptButton(bpy.types.Operator):
         arm.utils.check_default_props()
 
         if not os.path.exists(arm.utils.get_fp() + "/khafile.js"):
-            print('Generating Krom project for Kode Studio')
+            print('Generating Krom project for IDE build configuration')
             make.build('krom')
 
         if self.is_object:
             obj = bpy.context.object
         else:
             obj = bpy.context.scene
+
         item = obj.arm_traitlist[obj.arm_traitlist_index]
         pkg = arm.utils.safestr(bpy.data.worlds['Arm'].arm_project_package)
         # Replace the haxe package syntax with the os-dependent path syntax for opening
         hx_path = arm.utils.get_fp() + '/Sources/' + pkg + '/' + item.class_name_prop.replace('.', os.sep) + '.hx'
-        arm.utils.kode_studio(hx_path)
+        arm.utils.open_editor(hx_path)
         return{'FINISHED'}
 
 class ArmEditBundledScriptButton(bpy.types.Operator):
--- a/blender/arm/props_ui.py
+++ b/blender/arm/props_ui.py
@@ -347,7 +347,7 @@ class ARM_PT_ArmoryProjectPanel(bpy.types.Panel):
         layout.use_property_split = True
         layout.use_property_decorate = False
         row = layout.row(align=True)
-        row.operator("arm.kode_studio")
+        row.operator("arm.open_editor")
         row.operator("arm.open_project_folder", icon="FILE_FOLDER")
 
 class ARM_PT_ProjectFlagsPanel(bpy.types.Panel):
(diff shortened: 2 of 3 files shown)


MoritzBrueckner:
Altough this PR is still WIP (not tested on Mac/Linux, no build system for Sublime Text yet), I think it is ready to merge.
If Mac/Linux bugs are reported (I can't test it, I have only windows), it's not a big thing to fix them afterwards. It tried keeping everything the same way it was before, so maybe it's not an issue at all. And the Sublime build system can be added in a later PR, I currently don't now if it is possible without requiring the user to install Krom/Nodejs.

I'm trying to get this merged fast as it is a fix for #1399

What do you think, @luboslenco ?

zeeshan595:
The issue for linux is that `get_vscode_path()` gives the executable `/usr/bin/code` so if you just return p it should resolve the issue where it opens in a browser window.

Line: 571
```
def get_kode_bin():
    p = get_kode_path()
    if get_os() == 'win':
        return p + '/Kode Studio.exe'
    elif get_os() == 'mac':
        return p + '/Contents/MacOS/Electron'
    else:
        return p

def get_vscode_bin():
    p = get_kode_path()
    if get_os() == 'win':
        return p + '/Code.exe'
    elif get_os() == 'mac':
        return p + '/Contents/MacOS/Electron'
    else:
        return p
```

zeeshan595:
Can confirm your pull request works for me using vs code

MoritzBrueckner:
Great, thanks for testing :)

Additionally, I have now added a way to select a custom editor from the list in the Preferences (refer to the PR in armsdk). That option itself does (for now) nothing other than the "Kode Studio" selection (the executable is called together with the file path), but I think it's clearer for the user and easier when something needs to be adjusted for Kode Studio/VS code later on.

zeeshan595:
Indeed, I just found out that `System Default` option opens in terminal in Linux. Not sure if you want to change that but to be honest I think it should be fine. 
Also, I added the `project folder` button support in a pull request to your fork. Thought it would be better since this pull request was changing a bunch of code on armory paths anyway

https://github.com/MoritzBrueckner/armory/pull/1

MoritzBrueckner:
Thank you, it's merged.

>  I just found out that System Default option opens in terminal in Linux.

I think this has got to do something with this line: https://github.com/armory3d/armory/pull/1381/commits/22dd37cc67430e18387137fff3ce866a8d26414e#diff-5e90d27a49a6aa824980b6148e49ff31R573

If `Default` is selected, armory tries to find the editor by looking at the `VISUAL` and `EDITOR` environment variables, in that order. `EDITOR` actually refers to a command line editor and `VISUAL` to a GUI-based editor. The problem here is that many people use `EDITOR` also for visual editors, so it is probably better to include both variables in the search. (https://unix.stackexchange.com/q/4859) What do you think?

zeeshan595:
> 
> I think this has got to do something with this line: [22dd37c#diff-5e90d27a49a6aa824980b6148e49ff31R573](https://github.com/armory3d/armory/commit/22dd37cc67430e18387137fff3ce866a8d26414e#diff-5e90d27a49a6aa824980b6148e49ff31R573)

So checked my os variables and they are both wrong :D VISUAL is blank and EDITOR is set to a terminal editor. But i changed both to vs code and works now. Makes me think if linux distros actually use this.

MoritzBrueckner:
> Yea, but i think `VISUAL` should be used always unless it can't be found and `EDITOR` exists.

Luckily, that's the case here :)

> I don't know how it works in other platforms

I can only talk about windows and windows doesn't even use these variables by default. But many people use them. And if I'm not mistaken, git (for example) uses them too for getting the user's editor of choice. I think we're safe using them. The user will always have the possibility to change it or to switch to another custom editor via armory's preferences.

luboslenco:
>Custom build system for sublime text. So far the project can only be built in Blender with F5 or in Kode Studio. @luboslenco Is there a way to build the project using an external script? Then it could be called from Sublime (and more editors in the future).

If sublime can get path to armorysdk then it should be possible to build like a regular kha project (ie. what kode studio does) using `node Kha/make krom`. However scene changes made in Blender need to be rebuild from Blender as of now.

>What about the mklink methods in utils.py? Are they still needed or why were they commented out?

No longer needed, removing now.

>Should the editor-related methods in utils.py be moved to their own script? The file becomes more and more cluttered.

Sounds good!

>Should we write a tutorial (in the Wiki?) on how to add new editor configurations? Should the system change to make adding editors even easier?

Wiki additions on any topic always welcome! 🙂

bjarque:
Tested working on Mac, OS 10.14.4. You need to select it inside the .app package. Like this: "/Applications/Sublime Text.app/Contents/MacOS/Sublime Text"

MoritzBrueckner:
> You need to select it inside the .app package. Like this: "/Applications/Sublime Text.app/Contents/MacOS/Sublime Text"

I'm not familiar with MacOS, is that bad? Will users know that they have to select an executable inside a package? If not, is there a way of getting the executable in the script without hardcoding its filepath + filename relative to the .app package?

bjarque:
> I'm not familiar with MacOS, is that something bad? Will users know that they have to select an executable inside a package? If not, is there a way of getting the executable in the script without hardcoding its filepath + filename relative to the .app package?

Its the same with Kode, you need to select a binary inside the Contents/MacOS directory.  
Theres 3 options as i see it.

First, and most work is;
When you select the package, it doesn't browse into it, but sees it as a file. 

Then Armory could be set to run whatever executable file defined by **CFBundleExecutable** in **.app/Contents/info.plist, which is an XML describing the package. The executable file is always in the Contents/MacOS directory. Then you could skip the 2 extra options in dropdown, and just have "Custom" and "System Default" in the dropdown. Project files will then be generated from whastever the **CFBundleDisplayName** says.

Heres how the plist file looks for Kode:

> 

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>BuildMachineOSBuild</key>
	<string>16G1510</string>
	<key>CFBundleDisplayName</key>
	<string>Kode Studio</string>
	<key>CFBundleExecutable</key>
	<string>Electron</string>
...
```

Another option could be just just have predefined strings filled out in the config when you select Kode, VS Studio, or Sublime Text on mac. And then hide the input field, like it does when System Default is selected. The application path is the same for all users, and the path is AFAIK always in English language since its Unix. If one wants a custom path, they can choose Custom. Right now it doens't make sense to have all 3 menu options(Custom, "VS Code ı Kode Studio", Sublime Text), since they do exactly the same - run the executable defined in the text field. 

String for Kode: 
/Applications/Kode Studio.app/Contents/MacOS/Electron

String for VS Code
/Applications/Visual Studio Code.app/Contents/MacOS/Electron

String for ST3:
/Applications/Sublime Text.app/Contents/MacOS/Sublime Text

_**Third option:**_
Leave it as it is, and write some good documentation.

--

First option is most work, but leaves it very easy for the user, just select the package in "Custom" and then it works. The problem with that, is that it isn't transparent for the user what editor generates project files.

Second option gives intuitive selection of predefined editors, and makes it easier to generate project files, since you dont have to identify the package.

Third option is not really working well, since documentation isnt read upfront. As a fairly experienced mac user, it took me a few tries to get it to work.

I vote for the second option. Maybe there's better solutions out there?

MoritzBrueckner:
Are programs on mac always installed into the same folder? If yes, I agree that the second option is probably the best.
Unfortunately I'm not at home this week and I have not much time, so I will try it on the next weekend. Or, if you want to create a PR for this, feel free :)

PR #1382, opened 2019-09-14
Logic node that loads url

Sandy10000:
This node is dedicated to browser (HTML5) compilation.
Python : native_loadUrl.py
Haxe : LoadUrlNode.hx
Example : LoadUrl.blend

A logic node created with reference to the Haxe code on this page.
http://forums.armory3d.org/t/solved-open-web-browser-showing-url/3499


Sandy10000:
Since there is an error, withdraw it.

PR #1384, opened 2019-09-16
add a triangulation modifier on export

zaethan:
This should fix issues related to differences in triangulation by adding a temporary triangulation modifier on export to let blender handle mesh triangulation.
Sets the triangulation method to fixed and activates the keep normals option in an attempt to not mess up shading.
I am a little unsure about the `self.depsgraph.update()`, since the documentation of this method says that this `invalidates all references to evaluated data-blocks`, but this didn't seem to be an issue.

Initial idea based on the accepted answer of this stackexchange thread 

the relevant part of the blender python api https://docs.blender.org/api/current/bpy.types.ObjectModifiers.html#bpy.types.ObjectModifiers

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -1191,6 +1191,13 @@ class ArmoryExporter:
         table = objectRef[1]["objectTable"]
         bobject = table[0]
         oid = arm.utils.safestr(objectRef[1]["structName"])
+        
+        # add a standard triangulate modifier
+        tri_mod = bobject.modifiers.new(name="exportTriangulation", type='TRIANGULATE')
+        # try not to mess with the shading
+        tri_mod.quad_method = "FIXED"
+        tri_mod.keep_custom_normals = True
+        self.depsgraph.update()
 
         wrd = bpy.data.worlds['Arm']
         if wrd.arm_single_data_file:
@@ -1316,6 +1323,8 @@ class ArmoryExporter:
 
         if hasattr(bobject, 'evaluated_get'):
             bobject_eval.to_mesh_clear()
+            
+        bobject.modifiers.remove(tri_mod)
 
     def export_light(self, objectRef):
         # This function exports a single light object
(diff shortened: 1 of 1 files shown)


luboslenco:
Any idea on how much this impacts geometry export times?

zaethan:
As for how much this impacts the export, I don't have any numbers to base this off but I think it should not take longer than adding a triangulation modifier by hand (time between clicking on the modifier and seeing it in the viewport), which is usually just a fraction of a second on any given object, but could take longer on bigger ones (obviously). Adding a triangulate modifier on a human mesh created by mb-lab with about 70k vertices and about 68k quads takes less than a second for me, and I guess this is the time that would be added to the export...

I tested the ngon method and it should be set to clip. While beauty looks better, clip is what seems to be used in blenders viewport.
Also, triangulation messes with the normals of ngons pretty bad, unless the auto smooth option (Object Data properties -> Normals) is set. However the actual angle of the option is irrelevant. (To be honest, I don't know exactly why this is the case.)
Thus, on export, I save the original values of this option. If this option is not enabled I enable it and set the angle to 180 degrees. This way it should not mess up anyones normals, but it allows us to preserve the normals of the ngon. If the option is set, I just leave it as it is. After export I restore the original values.

To illustrate what I am talking about:
This is a simple sphere with a decimate modifier set to planar and some random angle (and blender wireframe overlay):
grafik


There are lots of ngons here to test.
Now use the normals as base color.



Adding a triangulation modifier, this definitely does not look like our object:


Setting the auto smooth option to any random value makes it look good again:


Armory on this PR (without wireframe):

(the slight shadows seem to be caused by some amient occlusion effect. At least turning on AO in eevee produces the same shadows)

Armory without this PR:

(notice the differences in the red and yellow parts at the bottom)



To be honest, this PR actually seems somewhat "cheap" to me as, instead of fixing the exporter, I just set some blender options the end user would have to set manually: apply blenders triangulation and then apply some "hack" to not mess up normals... to triangulate the mesh before it is even exported.
This makes the assumption that neither the triangulate modifier, nor the behavior of the auto smooth option changes in the near future, thus this is probably more of a workaround than a proper fix.
As can be seen in the last image, the normals are being exported correctly (by the current exporter), it is only the geometry that is a little "messed up".

zaethan:
I think I was able to fix the exporter (instead of hacking around it), I will make a new PR for that.

Issue #1374, opened 2019-09-06
Weird artefacts showing up in the geometry when everything is fine in Blender.


onelsonic:
I’ve been tweaking a model in blender, then when I compile the project I have weird artifacts showing in the windows of the model.
Seems like a problem with the geometry in Armory?
(I deleted some vertices, scaled the model, apply some transform, etc...)
See the files : 

Blender File : http://forums.armory3d.org/uploads/short-url/r69kKIMpgHjKJCcGDx44hXJnDnA.blend
Texture File : http://forums.armory3d.org/uploads/default/original/2X/0/07b5f3cbb5612eb1c323680c65a37bda80c8f829.jpeg

Blender Viewport : http://forums.armory3d.org/uploads/default/original/2X/e/e46a5f988a98dded4db570d93e3e4a6921d2bf84.jpeg

Armory Output :
http://forums.armory3d.org/uploads/default/optimized/2X/f/fade041e9712f5c6cc8fa49cad2b2f92aa33296b_2_1035x586.jpeg

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**
Remove the weird artifacts

**System**
Blender: 2.80 official

N8n5h:
<img width="518" alt="fade041e9712f5c6cc8fa49cad2b2f92aa33296b_2_1035x58" src="https://user-images.githubusercontent.com/42382648/64639165-a28ea000-d3dd-11e9-9afa-6ca743b1e18d.png">

By artifact you mean those? seems like the wall is overlapping with the window, moving outward the window from the wall fixes the issue: 

![](https://media.giphy.com/media/h5vE3OFJbikbOtDC1i/giphy.gif)

zaethan:
Something else that should help: triangulate the object (before exporting).
After downloading your example I got the same artifacts you are getting. Adding a triangulate modifier (leaving it at its default setting "beauty") to the object and applying it solved the issue.

Something else I noticed: if you try out the different options (other than beauty) of the triangulation modifier, you actually get the same artifacts in blender. I guess what is happening here is that armory triangulates your object on export, but the algorithm used by armory produces a "different" triangulation than blender does in its viewport. Since your windows are very close to the wall as @N8n5h pointed out, this difference (although small) makes your windows and wall overlap and create those artifacts.

onelsonic:
thank you guys.
the triangulate modifier fixed it. I know now what to do if the issue occur again.
When possible I will try to adjust the modeling directly as well.
Thank you again.

[...]

PR #1401, opened 2019-09-23
change the exporter to use blenders triangle loops

zaethan:
Previously, the exporter exported triangles directly (l. 1086) and triangulated polygons (l. 1091). To triangulate polygons, it used the last loop of the polygon and "connected" it to all the other loops in the polygon (first + j, first + j +1). This is probably not what blender does internally.
However blender provides access to the triangles it creates through the loop triangles. By mapping the polygon indices to these triangles, it is possible to just export all loop triangles belonging to a polygon.

As for performance, I guess the biggest impact will come from `exportMesh.calc_loop_triangles()` or the iteration over the loop_triangles.
Testing with a landscape of 256*256 vertices (ant landscape addon), calc loop_triangles uses 0.9 ms of a total of 0.85 s export time for this mesh. Mapping the polygons to the triangles takes about 0.15 s of the total time, so if you know of a faster way to build the `polygon_index -> loop_triangles` map, using it would be a good idea.
I don't know (and I don't think) whether the impact of removing the previous triangulation makes up for this.

Testing with a sphere like [this](https://github.com/armory3d/armory/pull/1384#issuecomment-533889802), the model from [this issue](https://github.com/armory3d/armory/issues/1374#issue-490369480), a landscape from the ant addon and some of the armory examples it seems to be working.

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -951,7 +951,7 @@ class ArmoryExporter:
 
     def export_mesh_data(self, exportMesh, bobject, o, has_armature=False):
         exportMesh.calc_normals_split()
-        # exportMesh.calc_loop_triangles()
+        exportMesh.calc_loop_triangles()
 
         loops = exportMesh.loops
         num_verts = len(loops)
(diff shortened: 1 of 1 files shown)


luboslenco:
Awesome!! 🎉

PR #1403, opened 2019-09-23
Open folder button implimentation

zeeshan595:
Open folder button always opens in a web browser. This pull request will implement it so the default file manager is used to open the folder when open folder button is clicked.

NOTE: I am not sure about the windows implementation, could someone have a look to see if it works


zeeshan595:
Moved to https://github.com/MoritzBrueckner/armory/pull/1

Issue #1391, opened 2019-09-18
Change rotation logic nodes to use quaternion only


N8n5h:
This is to start a discussion/feature thread to gather some feedback on this, but I think the logic nodes that have rotation in their code should receive and return quaternion only and not use euler angles at all. 

This may cause some breakage with projects that depend on rotation nodes using euler angles and the loss of the convenience of modifying every angle but it will solve every problem with rotation that euler angles has like [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock) and [this infamous issue](https://github.com/armory3d/armory/issues/512) and probably any other issues the rotation nodes have right now.

The [double conversion problem](https://stackoverflow.com/questions/15047471/incorrect-conversion-from-quaternions-to-euler-angles-and-back) that exists when converting could be solved for one axis, but when you try to use more than one the issue shows up again, something reproduceable in other engines aswell.

So I think rotations should be done using axis, angles and vectors instead.

Let me know what you think.

PR #1406, opened 2019-09-26
Added support for angle axies rotation nodes

zeeshan595:
See https://github.com/armory3d/armory/issues/1391 for details

image


MoritzBrueckner:
I don't know if this works, but it could make sense (for keeping compatibility to the current `SetRotation` node) to combine both nodes (axis angle and euler) into one node which has a dropdown menu in which you can change the mode. Do you know if it is possible to change the input sockets dynamically so that they depend on the selected mode?

This would also preventing the add node menu of getting even more bloated.

Btw, I think you misspelled "Euler" in your node's title :)

zeeshan595:
Indeed I did miss spell it which means it is miss spelled everywhere. I can have a look to see if i can merge the 2 nodes into one and fix the spelling mistake :D

zeeshan595:
Now can use 
Euler Angles: 3d vector
Axies Angles (radians): 3d vector + float
Axies Angles (degrees): 3d vector + float
Quaternion: 3d vector + float
image

JamieB3rt:
Oooh, this is great! Would love to have this in master. I get so many problems with rotations, i.e copying the rotation of one object to another is a nightmare! This solves so many issues i have been having!

luboslenco:
Thanks!

PR #1409, opened 2019-09-26
Fix open_folder() method

MoritzBrueckner:
- Project folder was opened twice on windows and mac
- On windows, the parent folder was opened instead with the project folder selected


luboslenco:
Thanks!

Issue #1438, opened 2019-10-16
Read json node malfunctioning or misbehaving


QuartzKnight:
**Description**
Read Json node always output the same result even if you overwrite the json via Write Json nodes.
(Note that the same seems to happen with Write File and Read File nodes)

**To Reproduce**

Create a nodetree that allows you to store different values on a json file via Write Json nodes, and to read and visualize said values via Read Json nodes.
You'll realize that everytime you read the json it will output the same value, even if you try to overwrite it.

**Expected behavior**
You can write and read the json repeatedly and it will read the last written value.

**System**
Blender: 2.80
Armory: 2019.10.0
OS: Windows 7 64bit
Graphics card: Nvidia GeForce GTX 750 Ti

**Test File**
[jsonTest.zip](https://github.com/armory3d/armory/files/3735000/jsonTest.zip)

PR #1439, opened 2019-10-17
fix https://github.com/armory3d/armory/issues/1438

Sandy10000:
Added a checkbox for setting whether to use cache for "Read JSON" and "Read File" logic nodes.


luboslenco:
Yo, can you move the `Use cache` socket below the `File` socket? That way we can prevent the breaking change.

Sandy10000:
Fixed. There is a part where the check was canceled. Is this a problem? I think there is no problem with the corrections.

luboslenco:
Looks like the checks have passed, many thanks!

PR #1489, opened 2019-11-22
add initial support to Mapping dynamic inputs

N8n5h:
This is to help transition to Blender 2.81.
I tried to tackle obtaining the input value if something was connected but it turned more complicated than I thought, so for starters I added support to the dynamic inputs default values so it compiles with no errors when a mapping node is present.

Please point out any ugly code.


luboslenco:
Awesome, thanks!

Issue #1466, opened 2019-11-06
Modifiers are not applied when loading scenes.


blenderaiser:
**Description**
When I load a scene through 'set' or 'spawn', the modifiers on its objects are not applied. I tested through nodes as well as scripts, and the same result happened.
Interestingly, it works sometimes, however I need to runs separately each scene, and then run the main scene.
*This modifiers problem always occurs in the exported game.

**To Reproduce**
- Create a blend file.
- Create a new scene.
- Create a cube with the bevel modifier in the new scene.
- Return to the main scene.
- Create an empty object and attach a node tree like this:
![Screenshot_2019-11-05_18-49-58](https://user-images.githubusercontent.com/13911941/68256431-e03f2c00-000e-11ea-8574-ae20b42b79ec.png)
- Run F5 and press space.

**Expected behavior**
Modifiers should be applied to the loaded scene even when the scene is in another blend file (via link).

**System**
Blender: 2.8
Armory: 2019-11
OS: Windows 7 and Linux (Manjaro)
Graphics card: geforce gtx 750 ti

**Test File**
[test.zip](https://github.com/armory3d/armory/files/3811853/test.zip)

N8n5h:
I could reproduce it on my end thanks to your file :+1:

I was digging a bit on the Scene class to see if I could spot where the issue may be, but AFAIK Armory "apply" modifiers (by obtaining an evaluated copy from the dependency graph) when exporting meshes here: https://github.com/armory3d/armory/blob/3e57614e1254cd25f207be9c6d4bf8e5f211b0bc/blender/arm/exporter.py#L1269
So meshes should be at their "exported state" when loading scenes, regardless of being on the main scene or not.

My bet is that the dependency graph is not giving a correct copy of the object with "applied" modifiers for different scenes given the current trend: https://github.com/armory3d/armory/issues/1378#issuecomment-531398321

blenderaiser:
Thanks for your attention @N8n5h!

I made a humble example of a possible solution :P 
Open the main.blend and just run the script and look at the terminal to see the count of vertices with the modifiers applied.
Do you think this example will be helpful?

*If I'm not mistaken, the Unity3D also open each scene in exportation. But as there are few objects in the example, this is not even perceptible.
[Blender - Eval All Scenes.zip](https://github.com/armory3d/armory/files/3822470/Blender.-.Eval.All.Scenes.zip)

N8n5h:
[...]

PR #1498, opened 2019-12-02
fix export with a 'zoo collection'

N8n5h:
Add a temporary collection to the current scene that links all the foreign objects meshes and collections with meshes from other scenes so the depsgraph can "see" it and then the collection is destroyed. Later the "charged" depsgraph is passed to the exporter.

This is done because the dependency graph requires a scene change in order to be updated with the view layer. So instead of changing scenes to update the dependency graph, the objects are brought to the dependency graph. This behavior and other solutions are discussed with more detail in the issue linked below.

"Zoo" collection only gathers MESH and EMPTY (linked collections) to keep it minimal, but it could be expanded to include other types of objects if required.

Fixes this https://github.com/armory3d/armory/issues/1466

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -1763,7 +1763,7 @@ class ArmoryExporter:
         for o in self.meshArray.items():
             self.export_mesh(o, scene)
 
-    def execute(self, context, filepath, scene=None):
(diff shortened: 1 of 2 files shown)


luboslenco:
Thanks!

Issue #1378, opened 2019-09-10
Textures in materials are not visible in rendering with Armory "IS MISSING UV MAP"


kone9:
Thank you for contributing to Armory!

**Description**
Hi, I have a scene with several materials.The materials are 1 texture and the main shader connected.The theme is that the textures are not seen, it is rare because it is a simple material.
<img width="798" alt="Screenshot_3" src="https://user-images.githubusercontent.com/12847027/64580452-c067f080-d35c-11e9-957f-78343fb8d04d.png">
In armory render path it is configured as MOVILE and as a rendering engine this CYCLES, although it would not have to influence the selected render engine, since armory uses a separate engine.
<img width="800" alt="Screenshot_4" src="https://user-images.githubusercontent.com/12847027/64580611-4be18180-d35d-11e9-8b0b-ac226885ed0b.png">

**To Reproduce**
Create one as if it were a house, add materials to the objects using the main BSD and a texture as color ... Then render and you will notice that the textures are not visible ... There are also rare transparencies that should not be.

**Expected behavior**
When rendering with armory engine, simple materials with textures would have to work correctly ... In any Game engine where I export this scene in GLTF format the materials work well, including godot engine or Unity.

**System**
Blender:
<img width="268" alt="Screenshot_5" src="https://user-images.githubusercontent.com/12847027/64580755-e641c500-d35d-11e9-9031-c03d90497a8c.png">

Armory:
<img width="702" alt="Screenshot_6" src="https://user-images.githubusercontent.com/12847027/64580782-f8bbfe80-d35d-11e9-90b0-eb8ed0c1ac65.png">

OS:
<img width="626" alt="Screenshot_7" src="https://user-images.githubusercontent.com/12847027/64580794-0b363800-d35e-11e9-8841-4c2317eb392d.png">

Graphics card:
<img width="309" alt="Screenshot_8" src="https://user-images.githubusercontent.com/12847027/64580826-2739d980-d35e-11e9-8d27-6e76140d7ff1.png">

**Test File**
*(drag & drop the zipped .blend file here)*
[terremoto_juego.zip](https://github.com/armory3d/armory/files/3593485/terremoto_juego.zip)

N8n5h:
Can you toggle the system console by Window > Toggle System Console and check if there is any missing texture/image error message when you press Play? 

In the zip file you provided there is no external assets included like images, so I assume you link them from an external folder, right?

kone9:
[...]

PR #1503, opened 2019-12-05
Fix for faulty "missing uv maps"

N8n5h:
Ask for `export_uvs` to the original object instead of the evaluated copy. 

Fixes this annoying issue: https://github.com/armory3d/armory/issues/1378.

Reason is, it seems the dependency graph fails to update whenever custom properties are set. 
So for that I think asking the original object in this case could be sufficient to fix it seeing it doesn't seem to be necessary to use the evaluated copy info I think.

> The same could be applied to the other functions that are there in that chunk of code if they don't really require asking the evaluated copy.

Before:

Notice how in the end I change to different workspaces to trigger the dependency graph to update.

After:


luboslenco:
Sweet, thanks once again!

PR #1525, opened 2019-12-21
Add support for generated image textures

MoritzBrueckner:
No more pink meshes when using generated textures :)
The generated images are now converted and stored in the same place as converted textures were stored before (`[buildpath]/compiled/Assets/unpacked/`).
generated_textures


luboslenco:
Wow!

PR #1548, opened 2020-01-20
lights wouldn't render at all if use_shadow / cast_shadow was unchecked

ForestCSharp:
When testing lights without shadows, the shadow maps were in an invalid state, preventing the lights from rendering at all.


luboslenco:
Thanks!

PR #1563, opened 2020-02-07
Fix bool type trait properties

MoritzBrueckner:
Trait properties of boolean type were incorrectly interpreted as blank/undefined when they defaulted to `true`.


luboslenco:
Thanks!

PR #1578, opened 2020-02-19
Some smaller fixes

MoritzBrueckner:
- `Helper` and `Rotator` classes were missing imports and relied on old `Math` class
- `AddGroupNode` didn't initialize group with an instance offset


luboslenco:
👍👍

PR #1584, opened 2020-02-22
Fix Navmesh generation + few code style improvements (pep8)

MoritzBrueckner:
The navmesh generation operator was still using the old trait property system.

Thanks to nubleet on Discord for reporting :)


luboslenco:
Yay, thanks!

PR #1589, opened 2020-03-01
Deactivate "Debug Console" checkbox when Zui is not enabled

MoritzBrueckner:
Just a small improvement to prevent an error during compilation.
I've also updated the property description:
debug_console_description


luboslenco:
Nice!

PR #1590, opened 2020-03-02
Fix Vertex Color and Tangent not exporting sometimes

N8n5h:
Applied the same fix that is used for uvs, since it seems to solve the issue of not updating correctly for exporting of vertex colors and tangents too, until it's fixed on Blender's part.
Also added support for the Vertex Color node.


luboslenco:
🚀🚀

PR #1591, opened 2020-03-02
Fix `armory.logicnode.SetRotationNode` vector type for input

Sanva:
Fixes for example the `virtual_gamepad` demo crashing on startup
in Android.


luboslenco:
Awesome, thanks!

PR #1592, opened 2020-03-02
Action autobaking

philipmduarte:
Hey Lubos, can you check this out?

I've always had trouble animating characters for armory because I like to work with bone constraints. Usually what I did was export to fbx and then import back into blender. That workflow is a mess, so I made this little modification to bake constraints automatically.

It introduces a new property arm_autobake for the Armature type, and it displays it in the object data properties panel. If enabled, it doesn't export any bone constraint at all, but instead bakes all actions.

Now I dont need to do blender -> fbx -> blender in order to tweak animations. Hopefully this will help someone else too :)

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -802,12 +802,25 @@ class ArmoryExporter:
                     o['bone_actions'].append('action_' + armatureid + '_' + aname + ext)
 
                 orig_action = bobject.animation_data.action
+                baked_actions = []
                 for action in export_actions:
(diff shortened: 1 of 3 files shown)


philipmduarte:
This is a demo file for the feature
[autobake.blend.zip](https://github.com/armory3d/armory/files/4275902/autobake.blend.zip)

philipmduarte:
demo file
[autobake.blend.zip](https://github.com/armory3d/armory/files/4276007/autobake.blend.zip)

luboslenco:
Thanks for contributing!

Issue #1523, opened 2019-12-21
Infinite loop if an object has the same name as its collection


MoritzBrueckner:
### Description
A collection having the same name as an object belonging to that collection leads to an infinite loop in iron/Scene.hx causing a black screen and an recursion depth error.
_I've opened this issue in this repository because it could probably be fixed in `exporter.py`. I don't understand yet how collections are handled by armory, so maybe this is related to iron instead._

A not so good fix I came up with can be found here: https://github.com/MoritzBrueckner/armory/commit/8c08a68fd663fb6736ad04a5894aba1ee72ec179. As I don't fully understand how collections work, this might break something. I noticed that some name references to the object in the collection are replaced by its filename in `Scene.json` after exporting with the "fix" applied, so it doesn't solve the problem as it still occurs if you rename the collection to the "file name" of the exported object (see the outliner in `collectionNameBugWithFix.blend`).

If the issue is iron related, it can probably be fixed in `Scene.hx` (`spawnObject()`, `spawnObjectTree()` or `createObject()`).

I hope this explanation isn't too confusing, first I wanted to open a PR, then an issue and during the writing of this issue I realized that things are a little bit more complicated than I thought :)

Edit: A possible fix for this could be to reference objects not only by name but also by an ID

### System
Blender: 2.81
Armory: db4f0cc107cf987eb9cee6b23ac960c8da2433b2

### Test File
[collectionNameBug.zip](https://github.com/armory3d/armory/files/3991453/collectionNameBug.zip)
[collectionNameBugWithFix.zip](https://github.com/armory3d/armory/files/4104940/collectionNameBugWithFix.zip)
`collectionNameBug.blend` leads to the bug in current master, `collectionNameBugWithFix.blend` unfortunately leads to the same bug with my fix attempt applied but interestingly in current master too. Maybe this helps someone :)

philipmduarte:
I just had the issue. For the matter, my solution ( exporting all objects from linked collections ) fails with your _collectionNameBugWithFix.blend_ file as well. I think I'm gonna check iron next.

MoritzBrueckner:
I think the issue is that iron differentiates objects purely on their names whilst Blender has no problem with objects having the same name.

Actually, I found a exporter workaround I'm happy with but I totally forgot about it, will create a PR as soon as I'm sure it does not break anything :)

MoritzBrueckner:
https://github.com/armory3d/armory/pull/1595

PR #1595, opened 2020-03-03
Exporter: better handling of duplicate object names

MoritzBrueckner:
Fix/Workaround for https://github.com/armory3d/armory/issues/1523.

Iron does not allow duplicate names for objects but the now replaced code was still adding exporting objects to collections if their names were already "taken".

Also, objects in collections are now always exported with their full asset name, taking the collection name into account which decreases the rate of name conflicts massively.

In the very rare case that an object from a collection still has the same name as another object in the scene (that object then must be named `[collection_name]_[object_name].blend` for this problem to occur, which is very unusual), a warning is printed to the console and the collection object with the duplicate name is not exported.

---
*Just an idea for the future (off-topic):* Warnings sometimes get buried amongst all other kinds of messages in the console and tend to get overseen. I think a message at the start of the game that says that n warnings occurred during export could help here. Or a hint in the UI if it is not too annoying). Maybe I will look into that and open another PR.

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -837,7 +837,7 @@ class ArmoryExporter:
                         action_obj['name'] = aname
                         action_obj['objects'] = bones
                         arm.utils.write_arm(fp, action_obj)
-                
+
                 #clear baked actions
                 for a in baked_actions:
                         bpy.data.actions.remove( a, do_unlink=True)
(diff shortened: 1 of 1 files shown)


luboslenco:
Thanks!

> Just an idea for the future (off-topic): Warnings sometimes get buried amongst all other kinds of messages in the console and tend to get overseen. I think a message at the start of the game that says that n warnings occurred during export could help here. Or a hint in the UI if it is not too annoying). Maybe I will look into that and open another PR.

Sounds good. We could also de-clutter console output a bit by default, I think khamake has some silent mode switch.

PR #1600, opened 2020-03-05
Support for collection instance with autobaking armatures

philipmduarte:
I made some changes so the action autobaking feature wouldn't crash on collection instances. Here's a demo file, _main.blend_ should crash on master
[colinst-autobake.zip](https://github.com/armory3d/armory/files/4294058/colinst-autobake.zip)


luboslenco:
Thanks!

PR #1603, opened 2020-03-06
De-clutter console output and show warnings

MoritzBrueckner:
Requires https://github.com/Kode/khamake/pull/227 (already merged in the upstream repo).

This PR adds a new `Verbose Output` option to the UI that, when turned off, shows only some basic compile information, warnings and error messages. Most of the time you don't need messages about each asset that get's exported, it just distracts from warnings. When turned on, it shows the same content as before (so `true` is the old default and `false` the new one).

Verbose output option

In the best case, the new console output looks like this now:
Export screenshot

Also, warnings are now counted during export and a message is displayed both in the console and the UI to inform about warnings.

Console:
Warnings in the console

UI:
Warnings in the UI

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -182,7 +182,9 @@ class ArmoryExporter:
             self.export_pose_markers(oanim, action)
 
             if True: #action.arm_cached == False or not os.path.exists(fp):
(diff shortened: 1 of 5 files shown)


luboslenco:
Horay!

PR #1609, opened 2020-03-12
Update canvas dimensions if fullscreen is enabled

Simonrazer:


luboslenco:
Any description? Maybe canvas should always behave like this when running in game, and the resolution property should only be used as a reference in armory2d editor?

link https://github.com/armory3d/zui/pull/117 https://github.com/armory3d/armory2d/pull/51

Simonrazer:
> Maybe canvas should always behave like this when running in game , and the resolution property should only be used as a reference in armory2d editor?

I think that this might not be intuitive, and it takes options away from the game-creator, maybe someone wants to have the UI be fixed at a certain resolution. Admittedly unlikely, but possible. More options is always better, and here I think it is easy for the creator to see what turning on fullscreen does. Maybe it is best to have this option turned on by default. If fullscreen is turned off the set resolution in armory2d is basically only a reference.

luboslenco:
> Admittedly unlikely, but possible.

I have broken this many times myself but if we believe the feature is unlikely to be put to use and does not address a direct problem then we should skip it. Godot has a [great document](https://docs.godotengine.org/en/stable/community/contributing/best_practices_for_engine_contributors.html) on this. Perhaps we can adopt this for Armory as well and link to it from the manual? Any feedback welcome!

> easy for the creator to see what turning on fullscreen does

If I understand correctly the canvas size only affects anchors? I thought those are fullscreen already, so this fixes a bug! If a container of fixed size is needed you can still drop in an Empty element. 🙂

Simonrazer:
This document looks like a great guide, a guidelines place in the [Contribution-wiki ](https://github.com/armory3d/armory/wiki/contribute) with a link to this would likely be good. 
For me when playing the game without fullscreen enabled the anchors are placed at the position they'd be in if the game were at the resolution set in armory2d, so often not going all the way to the right/bottom. 
If you think that wanting a set resolution for the UI is not a problem if the UI were to always behave like it does with fullscreen enabled I will change it to behave this way. I personally don't know how one would use an empty element to achieve this but I'll figure it out I guess.
Maybe then there should be an indication that this set resolution is only a reference?

luboslenco:
> This document looks like a great guide, a guidelines place in the Contribution-wiki with a link to this would likely be good.

Great, link added.

> if the UI were to always behave like it does with fullscreen enabled I will change it to behave this way

Sounds good.

> I personally don't know how one would use an empty element to achieve this

If elements are parented to empty element, then the anchors will respect the empty instead of canvas. Empty should act like canvas-in-canvas with predefined size.

PR #1610, opened 2020-03-12
Fix geometry not working in published projects

N8n5h:
Updated the opt_exporter to fit new vertex format.

Fixes geometry not loading in published linux project and crash in published windows projects https://forums.armory3d.org/t/unable-to-build-and-run-default-scene-for-windows-krom/3948.


luboslenco:
👍👍

PR #1611, opened 2020-03-13
Fix noise texture

### LINKED ISSUE #2: Cyclesgame website links to GitHub homepage instead of repository
Hello,

It's not exactly an issue, but the GitHub link on the [cyclesgame](http://cyclesgame.org/) website links to http://github.com instead of the expected repo URL (https://github.com/luboslenco/cyclesgame). Just thought I'd let you know in case this isn't intended behaviour.

Great work, by the way! Big fan ever since zBlend. Keep it up :)

Simonrazer:
The way this fix is done might be against the way the shader was intended to function. Fix # 1 will still apply though, because for some reason the scale input of the noise texture is at id "2" instead of being at id "1". Fix # 2 is not letting the shader use the Z coordinate of its input because that crashes it. I have no idea why. Also changed the shifting parameters of the G and B values because I think that it looks closer to what it looks like in Blender with that.


luboslenco:
Thanks!

Issue #1292, opened 2019-06-09
GLSL compiler error box projection with mapping 


N8n5h:
#### Short Description:
GLSL Compiler crashes with this particular setup for a material:

![](https://user-images.githubusercontent.com/42382648/59153907-30a06680-8a3c-11e9-9bec-dd3f25cc3608.png)

#### Crash log:
```
Armory v2019.5 ($Id: 1669fe81d52283ecb6ecd4c6c8f0c8ca6770fadb $)
OS: linux, Target: krom, GAPI: opengl, Blender: 2.80 (sub 44)
Exporting Scene
Scene exported in 0.005975484848022461
Exported modules: ['audio']
Shader flags: ['_Irr', '_EnvCol', '_Deferred', '_ShadowMap', '_CSM', '_SinglePoint', '_SMAA', '_SSAO']
Running:  ['/home/user/Downloads/Armory_06_linux64/Armory_latest_commit/armsdk//nodejs/node-linux64', '/home/user/Downloads/Armory_06_linux64/Armory_latest_commit/armsdk//Kha/make', 'krom', '-g', 'opengl', '--shaderversion', '330', '--parallelAssetConversion', '4', '--to', 'build_test1/debug', '--nohaxe', '--noproject']
Using project from /home/user/Documents/Armory projects/self/latest_commit/test
Using Kha from /home/user/Downloads/Armory_06_linux64/Armory_latest_commit/armsdk/Kha
Creating Kha project.
Exporting asset 1 of 3 (Scene.arm).
Exporting asset 2 of 3 (World_irradiance.arm).
Exporting asset 3 of 3 (mesh_Cube.arm).
Exporting asset 1 of 2 (Material_data.arm).
Exporting asset 2 of 2 (shader_datas.arm).
Exporting asset 1 of 1 (grid_b.png).
Exporting asset 1 of 1 (brdf.png).
Exporting asset 1 of 1 (smaa_area.png).
Exporting asset 1 of 1 (smaa_search.png).
Compiling shader 1 of 19 (Material_mesh.frag.glsl).
Compiling shader 2 of 19 (Material_mesh.vert.glsl).
Compiling shader 3 of 19 (Material_shadowmap.frag.glsl).
Compiling shader 4 of 19 (Material_shadowmap.vert.glsl).
[...]

PR #1613, opened 2020-03-14
Fix support for box/triplanar mapping

N8n5h:
I modified the shader parser so that triplanar mapping of textures is supported.
Currently tested and working with Armory PBR and Principled BSDF.
Normals are supported.

Fixes this https://github.com/armory3d/armory/issues/1292

image

image


luboslenco:
Wohoo

PR #1614, opened 2020-03-14
Added node to start animation from a particular frame index

QuantumCoderQC:
New node to start the animation from a set frame in the animation.

Works best with parameter `Blend` not zero.


luboslenco:
Thanks!

PR #1616, opened 2020-03-16
Keep Canvas dimensions the same as the game dimensions

Simonrazer:
I had to make a new PR because I couldn't figure out how to add  a new commit since I didn't push the first PR from a my fork. 
Armory would now set the canvas dimensions every frame to the game's dimensions, so the anchors are where they should be even if the windows gets resized.


luboslenco:
Thanks a lot!

PR #1617, opened 2020-03-16
Upgrade Wave Texture node + Fix noise texture Value output

Simonrazer:
The Wave texture now takes all the values set in Blender into account. Also fixed the Noise texture value output where I forgot to set the scale input id to 2 when I initially tried to fix it.

### DIFF
--- a/blender/arm/material/cycles.py
+++ b/blender/arm/material/cycles.py
@@ -575,7 +575,18 @@ def parse_vector(node, socket):
         else:
             co = 'bposition'
         scale = parse_value_input(node.inputs[1])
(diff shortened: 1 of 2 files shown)


luboslenco:
🎉🎉

QuantumCoderQC:
Hello, 

Is the wave texture fixed only for Deferred clustered rendering? I am unable to compile it for Forward clustered rendering. I ask because I'm not sure if it is a bug or a missing feature.

Best regards,
QC

Simonrazer:
@QuantumCoderQC I missed to check if it works for Forward rendering, sorry. Fix will be up shortly.

PR #1621, opened 2020-03-19
Make procd textures 3d again, fix broken Z coord in bposition

Simonrazer:
I finally found out why the Z coordinate of 'bposition' was broken, a div by 0 occurred in the vertex shader if the mesh is flat. Fixed that and made the textures 3d again so they don't look stretched on not-flat objects.


luboslenco:
Thanks!

PR #1644, opened 2020-04-08
Fix debug console

MoritzBrueckner:
- Previously there was a null pointer exception when a scene was selected that had no world
- The outliner look was wrong since https://github.com/armory3d/zui/commit/038c6d4e40716a416583b464bd7461fc756b0a28


luboslenco:
Thanks!

Issue #1529, opened 2019-12-26
Logic node "Trait" does not work depending on conditions


Sandy10000:
**Description**
Click here for related forum page.(https://forums.armory3d.org/t/remove-object-if-it-has-contact-with-object-from-collection/3788)

If try to apply a trait that has not been applied to any other object using "Add Trait", the "Trait" node will not get the Trait. Specifically, "cname" is Null both times.

In the test file, press the space bar to fire a bullet. By setting traits in advance for Suzanne (unnecessary objects), can acquire traits and the first bullet firing works without any problem. However, after the second bullet, there is no error but no trait has been applied. (The bullet does not disappear)

By commenting out the following line, the trait is applied to the second and subsequent bullets.
```
if (value != null) return value;
```

**To Reproduce**
Run the test file.

**Expected behavior**
Allows to get traits that have not been applied to any objects.

**System**
Blender:2.81
Armory:2019.12

**Test File**
[Trait.blend.zip](https://github.com/armory3d/armory/files/4002587/Trait.blend.zip)

MoritzBrueckner:
The issue is that Traits were not exported when unused. Now there is a "Fake User" option added to address this in https://github.com/armory3d/armory/pull/1646.

PR #1646, opened 2020-04-10
Add "Fake User" setting for traits + reorder trait UI

MoritzBrueckner:
Traits now have a `Fake User` setting. If set to `true`, the trait will be exported even if it is deactivated. This fixes https://github.com/armory3d/armory/issues/1529.

Screenshot of the "Fake User" option

I've also slightly changed the trait UI:
**Before**:
traits_before

**After**:
traits_after

### DIFF
--- a/Sources/armory/logicnode/TraitNode.hx
+++ b/Sources/armory/logicnode/TraitNode.hx
@@ -14,6 +14,7 @@ class TraitNode extends LogicNode {
 
 		var cname = Type.resolveClass(Main.projectPackage + "." + property0);
 		if (cname == null) cname = Type.resolveClass(Main.projectPackage + ".node." + property0);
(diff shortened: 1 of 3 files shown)


Sandy10000:
Thank you for your support. I downloaded the file and checked it using the test file "Trait.blend".

Check the fake user checkbox of the trait "Bullet" set in the object "projectile". 



Press the space bar to shoot one shot and hit the floor to disappear. No problems occur while repeating this.

Shooting the second shot before it disappears after shooting the first one will freeze without error. It seems that there are still some problems.

MoritzBrueckner:
Oh ok I had this issue as well (only if two bullets are invisible at the same time) but after changing the code to the current one I could not reproduce that issue anymore. I will try to find the issue, thanks for testing :)

PR #1648, opened 2020-04-10
Proxy: Add option to keep local trait properties when syncing

MoritzBrueckner:
This PR adds an option to keep the values of trait properties when syncing with a proxy object and makes it the new default behaviour. This is helpful when you change and sync a proxy but you have many objects with different property values that reference that proxy.


luboslenco:
Thanks!

Issue #1515, opened 2019-12-15
combine vector connected to mapping node


Redfill:
**Description**
when connecting a vector to the scale input of the mapping node, it doesn't have any influence

**To Reproduce**
add a mapping node between your uv and your texture, add a "combine vector" and connect it to the scale input of the mapping node, set random numbers and run.

**Expected behavior**
that the scaling is done properly according to the combine vector

**System**
Blender: 2.81
Armory: latest
OS: ubuntu
Graphics card: gtx 950

[VectorConnectedToMapping.zip](https://github.com/armory3d/armory/files/3965339/VectorConnectedToMapping.zip)

N8n5h:
Not a bug, support for taking the input of the mapping node has yet to be implemented, currently it uses the default value.

Redfill:
then is there another way i can change the values during gameplay?

N8n5h:
By using material parameters? Not sure if right now it's possible. 

Taking the values for the input sockets has to be done here 
https://github.com/armory3d/armory/blob/1f8aed23bf29e56047081416c0942a0cd4bc00fe/blender/arm/material/cycles.py#L898

MoritzBrueckner:
Should be fixed in https://github.com/armory3d/armory/pull/1649, thanks for reporting :)

PR #1649, opened 2020-04-11
Refactor and fix mapping node

MoritzBrueckner:
- The mapping node now supports input values from connected nodes, fixes https://github.com/armory3d/armory/issues/1515
- Rotations are now supported for all three axes
- Now `Point`, `Texture` and `Vector` mapping types are supported (only `Normal` is missing, I don't know exactly what it does)

mapping_node

The only issue that's still left is that the z location seems to be a tiny bit off sometimes (cannot reproduce it reliably) and for planes the rotation can be off too but that's probably because of the location issue. Anyway, the mapping node is now much more supported than before.

### DIFF
--- a/blender/arm/material/cycles.py
+++ b/blender/arm/material/cycles.py
@@ -14,16 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import math
-import bpy
 import os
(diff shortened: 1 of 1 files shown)


luboslenco:
Wow!

Issue #1641, opened 2020-04-08
Armory need some more Canvas elements related nodes


knowledgenude:
Actually we have a good amount of Canvas nodes, but it not seems complete because:

We can't change the Checkbox value trought nodes, but slider, texts, etc. we can. If have a way to make a node like "Set Canvas Value" that works in all Canvas elements, this would be great.

Also we need a way to choose between Released, Down and Started in the button element. Already have a way to make this using events, but it act like a Started only.

luboslenco:
See https://github.com/armory3d/armory/pull/1651 👍

PR #1651, opened 2020-04-11
Added set checkbox and OnCanvasElement Node

QuantumCoderQC:
Two new canvas nodes as requested at #1641 

1. Canvas node to set CheckBox value

CanvasCheckBox

2. OnCanvasElement Started, Released, and Down node.

OnCanvasElement

Best Regards,
QC

### DIFF
--- a/Sources/armory/logicnode/CanvasSetCheckBoxNode.hx
+++ b/Sources/armory/logicnode/CanvasSetCheckBoxNode.hx
@@ -1,0 +1,42 @@
+package armory.logicnode;
+
+import iron.Scene;
+import armory.trait.internal.CanvasScript;
+
+class CanvasSetCheckBoxNode extends LogicNode {
+
+	var canvas: CanvasScript;
+	var element: String;
+	var value: Bool;
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+#if arm_ui
+	function update() {
+		if (!canvas.ready) return;
+
+		// This Try/Catch hacks around an issue where the handles are
+		// not created yet, even though canvas.ready is true.
+		try {
+			canvas.getHandle(element).selected = value;
+			tree.removeUpdate(update);
+		}
+		catch (e: Dynamic) {}
+
+		runOutput(0);
+	}
+
+	override function run(from: Int) {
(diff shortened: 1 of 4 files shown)


luboslenco:
Thanks!

knowledgenude:
Hi, @QuantumCoderQC, i am very grateful for the nodes! Canvas Set Check Box is Working perfectly, but On Canvas Element node is not updating the element position. It is working fine, but when i change the position of the element, is like the button keeps in the same place "invisible". Can you please test by yourself when you have time? Thanks!

QuantumCoderQC:
Hi, @knowledgenude, Thank you very much for the report. I just did some tests and there seems to be another "hidden" issue the reason why the current code isn't working.

In fact, you can check it with the "canvas get location" node, it also does not update the location if the element changes its location. 

Will try to find a fix soon.

QuantumCoderQC:
Hello again, @knowledgenude So I did a test again and I could not reproduce the issue that you just stated. That is the button position updates when movedAAAaAAAA. If it is not the case with you, maybe a sample `.blend `file will help me.

However, the other node "canvas get location" problem still persists.
Best Regards, 
QC

knowledgenude:
I meant that when the element's position is changed in Canvas, its position is only changed visually. Sorry for the bad information.

How to reproduce:

1. create an element in the Canvas editor

2. change the position of the element and save the canvas

3. when starting the player, the position will be visually changed, but to click it must be in the element's origin location (upper left corner of the screen)

Test file: 
[test_file.zip](https://github.com/armory3d/armory/files/4476519/test_file.zip)

QuantumCoderQC:
@knowledgenude, thank you for the test file.

The issue occurs due to a different anchoring position:
By default, the anchor for an element is `Top-Left`. But in your case, the anchor is at the `Centre`.

GithubCanvasTest

I will resolve the issue, but as a temporary workaround try setting the anchor of all elements to `Top-Left` please. Then you can drag/move your button to any position on the screen and click should work.

Best Regards,
QC

knowledgenude:
Okay i will do that. Thank you, @QuantumCoderQC and have a great day!

PR #1654, opened 2020-04-12
A bunch of really small improvements

MoritzBrueckner:
- Fixed the look of the debug console outliner
- Now using `arm.utils.safestr()` for sublime text project names
- Added an icon for the "Code Editor" operator so that it is more distinguishable


luboslenco:
Cool!

PR #1656, opened 2020-04-13
Not perfect but improved

Sandy10000:
Unfortunately, the scale remains strange, but the location and rotation work correctly.


luboslenco:
Thanks. Isn't the matrix left uninitialized now (ie. retains the value from last run) with `m.setFrom(w);` commented out?

Sandy10000:
Thank you for your advice. I will reconsider.

PR #1657, opened 2020-04-14
OnCanvasElement node now takes into account anchoring point

QuantumCoderQC:
OnCanvasElement node now takes into account anchoring points.


MoritzBrueckner:
Very nice! Just a small remark: switch-case statements in Haxe don't need brackets for the cases and you can use the enum names itself for comparison (instead of the values they represent).
So instead of
```haxe
switch(anchor)
    {
        case 1:
        {
            mouseX -= cx/2 - canvas.getElement(element).width/2;
        }
        case 2:
        [...]
    }
```

you can write
```haxe
switch(anchor) {
   // You don't even need to write Anchor.Top
    case Top:
        mouseX -= cx/2 - canvas.getElement(element).width/2;
    case TopRight:
    [...]
}
```
Thats a little bit shorter :)

PR #1665, opened 2020-04-16
Refactor exporter + huge cleanup

MoritzBrueckner:
- There is now one `ArmoryExporter` instance per scene (way less error-prone due to forced reset per scene)
- A lot of variables got renamed, for example the mysterious `o` is now `out_[datatype]` in many cases (not in all, yet)
- Added type annotations for many methods and variables. It was often very hard to find out which type a variable had. This makes the code much more readable and maintainable and improves the work with IDEs (code completion with [this package](https://github.com/nutti/fake-bpy-module)!) and mypy.
- A lot of code was made more comforming to the PEP8 standards
- Made a lot of methods static that don't require `self` as a parameter
- Better, os independent paths with `os.path.join()` in many cases. This also makes the exporter less error-prone
- String interpolation instead of `+` -> better performance
- Better performance through direct list and array initialization where possible.
- Added docstrings to some methods
- Also some methods in other modules got some smaller improvements like type annotations and docstrings

Overall, in my test files, the scene export is about 0.01s - 0.1s faster. Thats too less to be statistically significant and might be background noise, but it is a little bit and the only goal of this PR is better code that is *not* slower as before :) Type annotations actually introduce a tiny little bit of overhead during the loading of the module but it is almost not measurable - it is encouraged to use them. There are many more performance optimizations possible (optimizing loops and `''.join()` for string concatenation for example) that I might do eventually in the future.

Also, the exporter might profit from splitting it into multiple files, but I was not sure which method is the best for that. A good overview about the possibilities can be found here: https://stackoverflow.com/a/9638653/9985959

However, there is one flaw in this PR: There are so many lines changed that `git blame` will not be as helpful as before.

The cleanup is not fully complete for all methods (it takes a lot of time) but I think its way better than before and ready to merge if you agree. I tested it as much as possible :)

I would encourage all contributors to add docstrings and some comments to their code (if the code is not self-evident) in the future to keep the code quality stable :)


luboslenco:
Amazing!

PR #1667, opened 2020-04-17
WalkNavigation: set flying speed via mouse wheel

MoritzBrueckner:
Helpful for flying around with the default camera in larger scenes


luboslenco:
Thanks!

Issue #1404, opened 2019-09-25
ColorRamp Combination


DoMal27:
**Description**
Combining node 'ColorRamp' results cutouts. Both mixing sequences of colorramps by connect them by them self or by mixing with MixRGB node.
Similar problem:  #905 

**To Reproduce**
My setups:
Set 1:
or gradient texture or some socket from 'Texture Coordinate' in ColorRamp in another ColorRamp

Set 2:
a gradient texture in ColorRamp, another ColorRamp receiving same grandient texture, both ColorRamp mixing in a MixRGB node by a factor of a third ColorRamp

**Expected behavior**
No cutouts or missing portions of the mix or combination

**System**
Blender: 2.80- (sub75) 2019 07 29
Armory: 2019.9.0
OS: Windows 10.0.17763
Graphics card: Intel UHD Graphics 630 & Radeon RX580

**Pictures**
Left-Blender Result; Center-Node Setup example; Right-Armory Result

![Bug_MixColorRamp](https://user-images.githubusercontent.com/20363585/65560224-2a16fb80-df14-11e9-9f15-dfeb7312f3d7.png)

![Bug_DoubleColorRamp](https://user-images.githubusercontent.com/20363585/65559866-cb04b700-df12-11e9-9580-1e9925f62237.png)

j2l:
In 2019.11 B2.80
I also noticed a weird behavior with color ramps:
The range is re-interpreted, 
- below 0.5 has less "weight" than in Blender
- above 0.8 has more weight

See, top is Krom, bottom is Blender, same object (at 0,0,0), shader and color ramp:
![image](https://user-images.githubusercontent.com/65325/68992155-162eac80-0868-11ea-90dd-5bd29900b657.png)

Issue #905, opened 2018-09-29
colorramp problem b0.5 2.8


Simonrazer:
These two colorramps should be the same, but one of them is glitching out.
![b1](https://user-images.githubusercontent.com/30187223/46242046-b6efad80-c3c3-11e8-9961-12b0e9c7112a.PNG)
![b2](https://user-images.githubusercontent.com/30187223/46242049-bc4cf800-c3c3-11e8-8dcf-03e480a68d21.PNG)

[colorramp.zip](https://github.com/armory3d/armory/files/2430352/colorramp.zip)

AFWSI:
Might be the second one causing it.

@luboslenco 
I also discovered a issue with the color ramp. The node setup in this thread on my reply works in 0.4 ,but not 0.5b.

http://forums.armory3d.org/t/black-texture-error/1779/7

In version 0.4 it gives color ,but version 0.5 it just shows up as white and black.

Simonrazer:
@AFWSI no it is not, it is still broken with the second one removed

PR #1671, opened 2020-04-18
Fix and refactor the colorramp node (+ more)

MoritzBrueckner:
- Colorramps now use the "factor" value if the input socket is not connected
- Fixed an issue when the factor value was larger than the largest element's value, this fixes https://github.com/armory3d/armory/issues/905 (the output from the test file in that issue still looks off but it is not related to the colorramp anymore)
- Fixed an issue when a color output socket was connected to a float input socket. Now, there is proper automatic rgb to bw conversion copied over from the `RGB to BW` node (this fixes https://github.com/armory3d/armory/issues/1404)
- There were `#todo` comments in the code mentioning `const` for the colorramp element vectors, so they are constant now :)
- Value node outputs are also constant now as value nodes don't have any inputs
- New `Shader.add_const()` method to add global constants
- Added some type annotations (+ a little bit of code cleanup again)

**Before:**
colorramp_pre

**After:**
colorramp_post

### DIFF
--- a/blender/arm/material/cycles.py
+++ b/blender/arm/material/cycles.py
@@ -1060,18 +1060,20 @@ def parse_normal_map_color_input(inp, strength_input=None):
         con.add_elem('tang', 'short4norm')
     frag.write_normal -= 1
 
-def parse_value_input(inp):
+def parse_value_input(inp) -> str:
     if inp.is_linked:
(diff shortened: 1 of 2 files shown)


luboslenco:
😍😍

Issue #1674, opened 2020-04-23
ColorRamp Node error when using Parameter as Factor Input


Simonrazer:
I get this error when using a Node that is tagged as a parameter as a Factor Input for the colorRamp Node: 
```
ERROR: C:\Users\SK\Desktop\Armory\blends\FPS\build_FPS\compiled\Shaders\BumpMine_mesh.frag.glsl:7: '=' : global const initializers must be constant ' const float'
ERROR: 1 compilation errors.  No code generated.


ERROR: Linking fragment stage: Missing entry point: Each stage requires one entry point

SPIR-V is not generated for failed compile or link

Compiling shader 14 of 29 (Cyl_mesh.frag.glsl).
Compiling shader 11 of 29 (BumpMine_mesh.frag.glsl) failed:
Shader compiler error.
(node:11564) UnhandledPromiseRejectionWarning: Shader compiler error.
(node:11564) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:11564) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Finished in 3.6821541786193848
2 warnings occurred during compilation!
Build failed, check console
```

**To Reproduce**
Use a ColorRamp node in a Material, plug a Value node into its Factor input and check the Parameter option of this Value node. Occurs on Forward & Deffered, Krom & Browser.

**Expected behavior**
No error, Parameter syncing with the logic.

**System**
Blender: 2.82
Armory: $Id: 0cdb249d80bd88c9205526171b9cbd1310fe48e7 $
OS: Win 10
Graphics card: GTX 1060

PR #1675, opened 2020-04-23
Fix value node when used with material params

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1674

----

@luboslenco It looks like the ID in `arm_commit` does not use the commit hash but a hash value of the file's content(https://stackoverflow.com/a/1796675/9985959). Do you have any idea on how to improve this (git hooks maybe?) I stumpled upon this because I tried to find out which version @Simonrazer had in his issue above.
https://github.com/armory3d/armory/blob/df30e9b68158265e1fa181d3c2b43c91ece01a80/blender/arm/props.py#L13-L17

### DIFF
--- a/blender/arm/material/cycles.py
+++ b/blender/arm/material/cycles.py
@@ -1559,7 +1559,7 @@ def write_result(link: bpy.types.NodeLink) -> Optional[str]:
             res = parse_value(link.from_node, link.from_socket)
             if res is None:
                 return None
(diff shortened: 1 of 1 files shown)


luboslenco:
Hm, maybe `arm_commit` can be removed and python can handle that instead. If it's not a git repo and no commit sha is available then it's a standard release indicated by `arm_version`.

PR #1678, opened 2020-04-27
Fix compiling for multiple mapping nodes

MoritzBrueckner:
Previously there were variables in the fragment shader that were redeclared when using multiple mapping nodes.


luboslenco:
Thanks!

PR #1679, opened 2020-04-27
Added `CanvasScript.setUiScale` method.

Sanva:
This exposes the `Zui.setScale` method trough `CanvasScript` to allow changing the scale factor for the UI elements.

I'm not really sure if this is the right place to put this nor if there would be another way of accessing this setting from Armory at runtime :thinking:


luboslenco:
Looks good, thanks!

PR #1682, opened 2020-04-28
Fix irradiance export

MoritzBrueckner:
When `arm_minimize` was false, exported irradiance files got the ending `.json.json` and were not found.

---

@luboslenco Is there a reason that `arm.utils.write_arm()` does not add the `.arm` extension to minimized files but the `.json` extension to not minimized ones? Currently, there are a lot of lines like
```python
ext = '.arm' if bpy.data.worlds['Arm'].arm_minimize else ''
```
all over the codebase to compensate for this.

### DIFF
--- a/blender/arm/write_probes.py
+++ b/blender/arm/write_probes.py
@@ -228,8 +228,8 @@ def write_probes(image_filepath, disable_hdr, cached_num_mips, arm_radiance=True
 
     return mip_count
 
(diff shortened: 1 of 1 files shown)


luboslenco:
Thanks! Hm I do not think there is a good reason for `.arm` extension not being added. Maybe `.json` was added later on and that's where the mess started, and there is also `.lz4` for compressed arm files.

PR #1685, opened 2020-05-02
Corrected missing 'tex' and 'tex1' for custom materials/shaders

kennylerma:


luboslenco:
Many thanks for contributing!

Issue #1701, opened 2020-05-14
Mouseover in "On Canvas Element" node


RomStef:
Adding the Mouseover event inside the "On Canvas Element" node would give more GUI management possibilities. from the change of cursor to the passage of the mouse to the change of color and state of the interface are just some examples.

Thanks.

MoritzBrueckner:
Implemented in https://github.com/armory3d/armory/pull/1702, thanks :)

PR #1702, opened 2020-05-14
Refactor OnCanvasElementNode and add hover listener

MoritzBrueckner:
- OnCanvasElementNodes can now listen to hover events (click events are still default), closes https://github.com/armory3d/armory/issues/1701
- Fixed an error when Zui was disabled
- Better variable names
- Performance improvement (previously there were many iterations over all canvas elements on each frame in `canvas.getElement(element)`)


luboslenco:
Wohoo, thanks!

Issue #1704, opened 2020-05-15
Sounds don't stop when the scene is changed


knowledgenude:
[Sound issue.zip](https://github.com/armory3d/armory/files/4634072/Sound.issue.zip)

Issue #1703, opened 2020-05-15
Improve Play Sound Node


knowledgenude:
The Play Sound node should be like:

![playsound](https://user-images.githubusercontent.com/63247726/82042695-b7b08280-9680-11ea-8b5a-b34eb41db1e8.jpg)

Play - Plays the sound

Pause - Pauses the sound

Stop - Stops the sound

Repeat - If checked will Repeat the sound, but i wonder if is not better to make a Loop input instead of repeat, like:

![playsound](https://user-images.githubusercontent.com/63247726/82044062-07904900-9683-11ea-9bfb-8441b4a23690.jpg)

Loop work in this case would be: if the input is activated and the audio already done its loop, then the audio will be started again and again... I don't know about nodes internal work.

Out - ...

Done - Useful for play a sequence of different sounds separately

Running - Useful to avoid conflicting sounds

Also i think that can have practice way to sync audio and animations but i don't know if it is good in the same node or with another node like "Get Animation Frame".

The use would be: "When reloading animation is on frame 1 or more, then Play Sound..."

MoritzBrueckner:
HI, had some spare time, so here it is: https://github.com/armory3d/armory/pull/1705

knowledgenude:
Thank you so much, @MoritzBrueckner!

PR #1705, opened 2020-05-15
Big PlaySoundRawNode update

MoritzBrueckner:
This PR implements https://github.com/armory3d/armory/issues/1703 and adds even further functionality to the `PlaySoundRawNode`. Thanks to @knowledgenude for the proposal :)

**Screenshot**:
Screenshot

- Fixed boolean type node properties
- New inputs `Pause` and `Stop`
- New outputs `Running` and `Done`
- Setting for enabling loop playback (default `false`)
- `Retrigger` option that will play a sound from the beginning if `Play` is true but the sound is still playing (default `false`)
- Optional sample rate override to compensate for wrong playback pitch and speed (possibly a Kha issue)
- Changed the Node icon (does this even show up in the UI somewhere?)

**Example file**: (adapted from issue https://github.com/armory3d/armory/issues/1704)
[sound_node.zip](https://github.com/armory3d/armory/files/4637372/sound_node.zip)

### DIFF
--- a/Sources/armory/logicnode/PlaySoundRawNode.hx
+++ b/Sources/armory/logicnode/PlaySoundRawNode.hx
@@ -50,6 +50,7 @@ class PlaySoundRawNode extends LogicNode {
 			case Stop:
 				if (channel != null) channel.stop();
 				tree.removeUpdate(this.onUpdate);
+				runOutput(2);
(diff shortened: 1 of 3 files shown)


luboslenco:
👍👍

Issue #1710, opened 2020-05-21
Running blender in background mode fails


tong:
**Description**
Running blender in background mode fails with armory addon.

**To Reproduce**
Start blender in background mode:
`blender -b test.blend`

```sh
Blender 2.82 (sub 7) (hash 375c7dc4caf4 built 2020-05-13 23:06:57)
Read prefs: /home/tong/.config/blender/2.82/config/userpref.blend
psutil available
Exception in module register(): /usr/share/blender/2.82/scripts/addons/object_collection_manager/__init__.py
Traceback (most recent call last):
  File "/usr/share/blender/2.82/scripts/modules/addon_utils.py", line 382, in enable
    mod.register()
  File "/usr/share/blender/2.82/scripts/addons/object_collection_manager/__init__.py", line 98, in register
    km = wm.keyconfigs.addon.keymaps.new(name='Object Mode')
AttributeError: 'NoneType' object has no attribute 'keymaps'
Error: Traceback (most recent call last):
  File "/home/tong/.config/blender/2.82/scripts/addons/armory.py", line 260, in execute
    start.register(local_sdk=local_sdk)
  File "/home/tong/sdk/ArmorySDK//armory/blender/start.py", line 34, in register
    arm.keymap.register()
  File "/home/tong/sdk/ArmorySDK//armory/blender/arm/keymap.py", line 9, in register
    km = wm.keyconfigs.addon.keymaps.new(name='Window', space_type='EMPTY', region_type="WINDOW")
AttributeError: 'NoneType' object has no attribute 'keymaps'

location: /usr/share/blender/2.82/scripts/modules/bpy/ops.py:201

Traceback (most recent call last):
  File "/home/tong/.config/blender/2.82/scripts/addons/armory.py", line 260, in execute
    start.register(local_sdk=local_sdk)
  File "/home/tong/sdk/ArmorySDK//armory/blender/start.py", line 34, in register
    arm.keymap.register()
  File "/home/tong/sdk/ArmorySDK//armory/blender/arm/keymap.py", line 9, in register
    km = wm.keyconfigs.addon.keymaps.new(name='Window', space_type='EMPTY', region_type="WINDOW")
AttributeError: 'NoneType' object has no attribute 'keymaps'

location: /usr/share/blender/2.82/scripts/modules/bpy/ops.py:201

location: /usr/share/blender/2.82/scripts/modules/bpy/ops.py:201
Error: Traceback (most recent call last):
[...]

PR #1711, opened 2020-05-21
Fix for running blender in background mode

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1710. Thanks to @tong for reporting this.


tong:
Works!

luboslenco:
Yay!

PR #1712, opened 2020-05-23
Allow to set custom play scene

tong:
This allows to (optionally) specify a scene to launch in player menu.  

I think it's a common workflow to load objects from other scenes into your main 'Game' scene, and pretty annoying to have to switch between scenes just for launching.


luboslenco:
Thanks!

PR #1715, opened 2020-05-24
Proper debug console log output formatting

tong:
Print the position and customParams passed to trace in the DebugConsole log.   

```haxe
trace("A","B","C");
```

Prints:
`my.Class.hx:23: A, B, C`

Instead of just:
`A`


luboslenco:
Thanks!

PR #1724, opened 2020-06-05
Replace __js__ with Syntax.code

tong:
Replace `untyped __js__` with `Syntax.code` and get rid of deprecated compiler warnings.


luboslenco:
Thanks!

PR #1725, opened 2020-06-05
Add option for exclusive Fullscreen (WindowMode 2)

Simonrazer:
I am having issues with using Vsync in WindowMode 1, while it works fine in WindowMode 2. This is why I added this option to the UI. The original Fullscreen option still does the same thing (WindowMode 1).

### DIFF
--- a/blender/arm/props.py
+++ b/blender/arm/props.py
@@ -87,7 +87,8 @@ def init_properties():
     bpy.types.World.arm_compiler_inline = BoolProperty(name="Compiler Inline", description="Favor speed over size", default=True, update=assets.invalidate_compiler_cache)
     bpy.types.World.arm_winmode = EnumProperty(
         items = [('Window', 'Window', 'Window'),
-                 ('Fullscreen', 'Fullscreen', 'Fullscreen')],
+                 ('Fullscreen', 'Fullscreen', 'Borderless Fullscreen Window'),
+                 ('Exclusive Fullscreen', 'Exclusive Fullscreen', 'Exclusive Fullscreen')],
         name="Mode", default='Window', description='Window mode to start in', update=assets.invalidate_compiler_cache)
     bpy.types.World.arm_winorient = EnumProperty(
         items = [('Multi', 'Multi', 'Multi'),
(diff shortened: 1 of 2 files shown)


MoritzBrueckner:
Hi, I'm not sure whether those two lines need to be updated as well?

https://github.com/armory3d/armory/blob/e30a8c7f4679a415029282080af9e80d46e63941/blender/arm/write_data.py#L415
and
https://github.com/armory3d/armory/blob/e30a8c7f4679a415029282080af9e80d46e63941/blender/arm/write_data.py#L429

Simonrazer:
Jup, changes there are necessary for HTML5 target to work as well. Good catch, thanks!

Simonrazer:
Sorry for the late reply. The problem I had was that Vsync would set the framerate in fullscreen mode to 60, even though I had a 144Hz monitor. It seemed to work when using exclusive Fullscreen mode, even though at the time of testing it wasn't even really using exclusive Fullscreen mode but still just Fullscreen mode. Now exclusive Fullscreen actually locks it at 60, while Fullscreen does what it is supposed to do and lock it at 144. Great!
Since you say exclusive Fullscreen is in a way deprecated and Fullscreen now seems to work just fine I think this PR can be closed and forgotten.
A personal problem that may effect many is that I can't increase the logicrate in iron/system/time anymore, there seems to be no effect on the amount of frames put out, just on the logic itself. This murders my brain because that was working a fine all the time, now to get "real" high framerates in my games I have to execute the update function in iron/App on every render(), otherwise everything looks like at 60FPS or lower. The difference is very noticeable. This brings the problem of the logic running at different speeds on different setups.
To recap: 
Default behavior: Games run at 500+FPS, the logic is being executed at 60FPS -> the game looks jittery.
If the logicrate is changed to 1/200: Game runs at 500+FPS, the logic is being executed at 200FPS, the game still looks jittery (I think because the same frames are put out multiple times but actually I am clueless to why that is).
If the logicrate is tied to the Framerate: Game runs at 500+FPS, logic runs at 500+FPS -> the game looks buttery smooth with the issue of it running at different speeds on different hardware/when different stuff is on screen.
I wish there is a way this can be preferably two options in the UI: one for the game to run at a logic rate set by the user and one for the logicrate to be tied to the framerate, then allowing the user to fix the speed issues by using time.realDelta wherever they need (this would need to be changed for nodes as well).

PR #1726, opened 2020-06-07
Node socket improvements

MoritzBrueckner:
- Strings with quotes are now properly escaped (previously there was an error)
- Added a new socket type `ArmCustomSocket` that can be inherited by logic node pack custom sockets
  Previously, custom sockets were interpreted as floats because they have the type `VALUE`. There is a `CUSTOM` socket type in Blender but it is forbidden to set it directly, so this custom socket is a workaround for that. Note that it isn't registered so it cannot be used directly as a socket.


luboslenco:
👍👍

PR #1727, opened 2020-06-08
Colored terminal output

tong:
Adds colored terminal output using ansi escape codes.  
There might be some more places to use it, currently added to warnings and errors.   
I've only tested on linux but afaik this should be fine with macos, windows.

Screenshot_2020-06-08_12-29-42

### DIFF
--- a/blender/arm/log.py
+++ b/blender/arm/log.py
@@ -1,3 +1,9 @@
+DEBUG = 36
+INFO = 37
+WARN = 35
+ERROR = 31
+
+no_colors = False
 info_text = ''
 num_warnings = 0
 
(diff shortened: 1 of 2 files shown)


MoritzBrueckner:
Very nice!

I think those escape codes work only on Windows 10, I'm not sure if they are omitted or echoed directly (maybe someone can test this?)

If this doesn't work below Windows 10, we could add a global variable to the log module:
```python
import platform

if platform.system() == "Windows":
    # This needs to be reworked if Windows gets a new release in the future
    HAS_COLOR_SUPPORT = platform.release() == "10"
else:
    HAS_COLOR_SUPPORT = True
```

tong:
A windows guy just told me that cmd.exe (windows<10) doesn't support ansi codes but has it's own commands for colors (https://www.howtogeek.com/howto/16815/how-to-personalize-the-windows-command-prompt/).

MoritzBrueckner:
The problem with this command is that it changes the colors of the entire command line and not just individual output parts. There are Python packages like colorama that can print colors also on Windows 8 and below but this would mean another dependency for Armory3D, so its probably best to keep it how it is now.

luboslenco:
Thanks!

PR #1728, opened 2020-06-10
Implement custom default values for custom node sockets

MoritzBrueckner:
Not connected custom input sockets now can have arbitrary default values (as long as they can be represented in Haxe syntax), previously they would just have the value `null`.


luboslenco:
Thanks! 🙂

PR #1735, opened 2020-06-18
Add reroutes and frames to add node menu

MoritzBrueckner:
Reroutes and frames were always supported by the logic node editor but not available from the add node menu. This PR adds them in the new category "Layout".

The layout category is appended to the end of the list by purpose to mimic the other node editors and to not interfere with the actual logic nodes.

LayoutNodesScreenshot

I've also added the possibility to create more layout nodes in the future (maybe Blender gets text areas in the future so we could add a proper comment node) by simply assigning them to the layout category.


luboslenco:
Yay!

Issue #1734, opened 2020-06-17
World Background wrong when using HDRI and second Scene is present


Simonrazer:
**Description**
The World Background of a starting Scene uses a standart grey world if more than one scenes are present and the starting Scene uses an HDRI, the effect on the environment of the correct HDRI is still present.

**To Reproduce**
Create a scene and give it a HDRI World background, create a second scene with a flat color world, play.

**Expected behavior**
The correct world background to appear.

**System**
Blender: 2.82
Armory: 2020.6
OS: Win 10
Graphics card: GTX 1060

**Test File**
[worldbug.zip](https://github.com/armory3d/armory/files/4794396/worldbug.zip)

knowledgenude:
I did not understant what you mean. Do you want the same background for both scenes? If yes, you have to set the same World in the scene properties for both scenes, or you can create a new World in the same place for different backgrounds.

Simonrazer:
I want the correct world background to appear for the first scene (the HDRI), at the moment it shows a standart Blender grey world background (if the second scene with its own world exists in the .blend as well).

MoritzBrueckner:
Found the issue, now I have to find a way to fix it :)

There is a big underlying problem in the way environment textures and other world attributes are handled by the exporter. There is a "list" (actually a string) of definitions in the `"Arm"` world called `world_defs`. The problem is that this list is generated once for all scenes and it does not take multiple scenes into account, so the export of the second scene deletes an already set flag called `_EnvTex` that is required by the first scene.

So even if there was a fast fix for your issue, I think it is better to try fixing the whole system behind it. Hopefully I'll have some more time in the next days.

MoritzBrueckner:
Hi @luboslenco, I need your opinion:

It looks like this problem is only solvable by a quite big refactor of the world exporter including the world shader. Before I proceed I want to make sure that you're ok with my proposal on how to solve this.

Currently, there is one single shader used for all worlds (`world_pass`) with certain parts turned on/off via defines. The problem is that different worlds need different defines. My proposal is to rewrite the world exporter so that it exports a unique shader per world. Needless context switches are no issue here as obviously only one world is rendered at the same time.

[...]

PR #1744, opened 2020-06-26
Multi-world support

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1734.

Requires https://github.com/armory3d/iron/pull/96.

**~This PR is WIP but opened to enable discussion, I will remove the draft status when~ it's done.**

~There is still a lot to do and the currently committed code is definitely not optimal and very messy. What makes everything more complicated is that the shader export has to be done manually for worlds as the available helper classes require materials in order to fully work. But this is out of the scope for this PR, I don't understand enough about the current system to also refactor this.~

~I currently have very limited time unfortunately, so please don't expect big updates everyday! But I will work on it :)~

---

**To do:**
- [x] Make some `Arm` world attributes individual per world
  - [x] Remove old `Arm` world attributes
- [x] Update UI
- [x] Export individual shaders per world
- [x] Replace the old world pass functionality with the new individual world shaders
  - [x] Static colors
  - [x] Hosek Wilkie skies
  - [x] Clouds
  - [x] Environment maps and static background images#
- [x] Remove world_pass
- [x] Global render path switch for clouds
- [x] Update deferred/forward render paths
- [x] Find a good solution for scene changes (Iron)
- [x] Code clean up


MoritzBrueckner:
The entire world pass functionality is now reimplemented for individual words, only the Iron side of things and the render path change is missing now (at least here on Github, I already did a few things locally) :)

I decided to remove the water and fog functionality from this PR todo list as this is not related to the world shader and needs further discussion (should this be on a per-scene or per-world basis?)

MoritzBrueckner:
I found a way to unload shaders now, the Iron commits will follow tomorrow and then it's almost ready to merge.

@luboslenco Sorry to bother you again, but in case you missed the notification: there are two open questions in the above post that we probably should resolve before merging (at least the irradiance thing). Thank you :)

MoritzBrueckner:
I think I have found good solutions now for the questions above and this PR is ready to merge if you agree :)

luboslenco:
Insane, thanks for pulling this off!

Regarding the make_world callback, it was a way for libraries to inject defines but I do not think it's needed anymore.

Issue #1706, opened 2020-05-18
Physics Ray Cast / Pick Object don't works in C targets


knowledgenude:
Tested in release and debug mode

The test file contains an exported release apk

[tests.zip](https://github.com/armory3d/armory/files/4641529/tests.zip)

QuantumCoderQC:
Thanks for the report!

Unfortunately, it is also the case with Windows(C) targets. Both Release and Debug variants do not react to ray cast and Pick Object.

PR #1747, opened 2020-06-29
Fixed `armory.trait.physics.bullet.PhysicsWorld.rayCast` for Hashlink.

Sanva:
Just enabling the same code used for C++.

This fix for example https://github.com/armory3d/armory_examples/tree/master/physics_raycast not working in C targets — and I think it should fix (or at least improve the situation with) #1706.


knowledgenude:
Physics Ray Cast and Pick Object are working perfectly now. Thanks!

PR #1759, opened 2020-07-05
Fix initialize var to null

tong:
Static targets and doc-generation are complaining:
```sh
../armory/Sources/armory/system/Starter.hx:17: characters 8-13 : Warning : Local variable tasks might be used before being initialized
```


luboslenco:
Yo, is the `var tasks : Null<Int> = null;` box not needed anymore on static platforms?

tong:
Indeed (#).
set to `tasks = 0`

luboslenco:
Thanks!

PR #1762, opened 2020-07-07
Print compositor flags on export

MoritzBrueckner:
@luboslenco Off-topic, but also regarding the verbose mode: there was a bug in Khamake that didn't show shader errors when verbose mode was disabled. It was fixed a few months ago, but is not yet included in Armory's version of Kha/Khamake. Could you please update that? Thanks :)


MoritzBrueckner:
Oops, I forgot to update my local submodules (checking out the right commit was not enough)^^ Sorry about that

luboslenco:
No prob at all! Thanks!

PR #1763, opened 2020-07-07
Fix SetSceneNode when using json export

MoritzBrueckner:
@luboslenco Should Iron print more informative error messages than the usual cannot read hxbytes thing or is it bad because of the overhead of try/catch statements? Also, should the game stop with such error messages (does not work on every target)? Currently it just prints out errors every frame if such bugs as the now fixed one occur.

### DIFF
--- a/Sources/armory/logicnode/SetSceneNode.hx
+++ b/Sources/armory/logicnode/SetSceneNode.hx
@@ -13,6 +13,10 @@ class SetSceneNode extends LogicNode {
 	override function run(from: Int) {
 		var sceneName: String = inputs[1].get();
 
+		#if arm_json
+		sceneName += ".json";
+		#end
+
 		iron.Scene.setActive(sceneName, function(o: iron.object.Object) {
 			root = o;
 			runOutput(0);


luboslenco:
We could do try/catch in logic nodes, for script it's easy to do manually when there is a risk of file missing and quit if necessary.

MoritzBrueckner:
Hmm, wouldn't that mean that we have a lot of duplicate code everywhere and that we cannot really differ between different errors? Another way would be to use Kha's `failed` callbacks for the asset functions (`getBlob()` etc.) but last time I used them I think they didn't work on many targets or situations.

But your right, there might be situations where the error is so critical that you wan't to quit and there are others where it's ok to continue. Maybe a combination of a central error message and a callback would be good. I don't know.

luboslenco:
> Hmm, wouldn't that mean that we have a lot of duplicate code everywhere and that we cannot really differ between different errors?

Do you mean duplicate code in regards to logic nodes or scripts? For logic nodes there should be very few nodes accessing the file system.

MoritzBrueckner:
> Do you mean duplicate code in regards to logic nodes or scripts?

For both actually but I had nodes in mind when I wrote it. You're right - it's probably not that much redundant code and scripts probably can handle that ("everywhere" definitely wasn't the right word from me^^).

The only problem I still see is that it can be difficult to differ between types of errors (a try/catch block for `setActiveScene` for example catches all kinds of errors, not only those that occur because of missing files). But let's see, maybe Kha will update the exception handling when updating to a newer Haxe version (4.1 introduced a `haxe.Exception` type).

PR #1767, opened 2020-07-11
Fix OnContactNode

MoritzBrueckner:
Fixed a nullpointer error that occured when the first given object has no rigid body trait.


luboslenco:
Many thanks!

Issue #1737, opened 2020-06-18
Receive shadows don't work as disabled


knowledgenude:
In material settings there is an option to disable the material shadows receiving, but the option don't work, the material always receive shadows.

MoritzBrueckner:
Hi, it works for me, do you have a minimal example file? Please note that this option only works with forward rendering, otherwise it should be disabled in the UI.

knowledgenude:
Here is an example: [untitled.zip](https://github.com/armory3d/armory/files/4907736/untitled.zip)
I tested it again and it's not working. What option i have to enable in UI?

MoritzBrueckner:
It has something to do with materials for mobile render path, the required code is missing in `make_mesh.py` in `make_forward_mobile()` but it's included in `make_forward_base()`.

I don't know yet if I'm able to fix it (I can't test it on mobile devices, my android builds always fail with the most obscure errors or the built app just crashes instantly) but if I do I'll let you know and open a PR :)

knowledgenude:
Okay. Thank you! If you need, feel free to send me any exported test build and i try to generate an APK from it. ^^  Also if you have some error in Android Studio you can send it too, then maybe we can find a solution for it. But as the mentioned problem is happening inside Armory Player, an apk shouldn't be necessary.

PR #1770, opened 2020-07-13
Implement "receive shadow" setting for mobile path + all light types

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1737.

The required checks were only implemented for sun lamps on full render path, now each lamp type and the mobile render path supports the `Receive Shadow` material setting (forward rendering only).

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -346,7 +346,7 @@ void main() {
 	fragColor.rgb += sampleLight(
 		p, n, v, dotNV, pointPos, pointCol, albedo, roughness, occspec.y, f0
 		#ifdef _ShadowMap
-			, 0, pointBias
+			, 0, pointBias, true
 		#endif
 		#ifdef _Spot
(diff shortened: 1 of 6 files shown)


luboslenco:
Awesome!

onelsonic:
Hi there,
looking at the custom celshade renderpath from the example :
[https://github.com/armory3d/driver_celshade/blob/master/blender.py](https://github.com/armory3d/driver_celshade/blob/master/blender.py)
it has the same issue with "receive shadow" not being correctly implemented.

adding the following : 
`frag.write('float emission;') ` 
at ligne 65 is fixing the emision error but the "receive shadow" is still troublesome.
Also it seems that only the sun light is implemented.
Any ideas how to update it? It will be nice to have some NPR rendering as well.

MoritzBrueckner:
Thanks for mentioning this, it should work again with https://github.com/armory3d/driver_celshade/pull/8. However, I didn't fix the receive shadow setting but I couldn't reproduce that issue. Can you please test if it works now with that PR and let me know at https://github.com/armory3d/driver_celshade/issues/7?

PR #1775, opened 2020-07-22
Fixes `PhysicsConstraint.init` `Generic6DofConstraint` usage for Hashlink

Sanva:
Currently https://github.com/armory3d/armory_examples/tree/master/physics_ragdoll doesn't build for C targets — builds and works fine with this fix, tested in Linux and Android.


luboslenco:
Thanks!

PR #1777, opened 2020-07-25
UI for quick Node variable access

Simonrazer:
Quickly add references to node-variables, and setters. Here is a  showcase:
https://youtu.be/Sx4aoxVHiPo
It is definetly more handy in more complex Node setups, just wanted an quick test to demonstrate this.

Any suggestions? Having the "Set" button spawn two nodes is a bit janky, but I don't know how to make a single node that modifies a node-variable based on its ID.

### DIFF
--- a/blender/arm/nodes_logic.py
+++ b/blender/arm/nodes_logic.py
@@ -91,6 +91,117 @@ class ArmOpenNodeSource(bpy.types.Operator):
             name = context.active_node.bl_idname[2:]
             webbrowser.open('https://github.com/armory3d/armory/tree/master/Sources/armory/logicnode/' + name + '.hx')
         return{'FINISHED'}
+    
+#Node Variables Panel
+class ARM_PT_Variables(bpy.types.Panel):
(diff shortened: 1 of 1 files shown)


knowledgenude:
The biggest problem in nodes for me, is when i want to connect some node too far from the other one, it gives me tendonitis and force too much my vision (that zoom out) :/ I can't figure out a way to avoid this.

For example when i have a float for "Speed" and i want to consider the value from "Speed" in other part of the tree, but this part is too far.

I don't know about a way to do this, but is possible to connect two sockets without "drag and drop" that end line? Just like "click output" then "click input". This can allow to zoom in/out without problems.

Also increasing that magnet range that pulls the cursor to the nearest socket should be good, like 2 or 3 squares.

Simonrazer:
To connect two nodes you can also press "F". I made this PR also to find a way around long connections being used everywhere, since now you can easily get the value of a node-variable, so you don't need to create a new reference to it via the ID by hand, and so will use more instances of node-variables instead of long connections to the original node.

PR #1788, opened 2020-08-04
Remove some operators from op search menu + code cleanup

MoritzBrueckner:
There are a lot of trait related operators that are heavily dependent on the currently selected trait indices for example, those were removed from the operator search menu because they were throwing errors in some cases. I've also given some operators more descriptive names and added tooltips to a few.

operator_search


luboslenco:
Thanks!

PR #1792, opened 2020-08-06
Removed logic nodes icons

knowledgenude:
I've just set all nodes icons to 'NONE' and fixed 'Set Checkbox' label name.

There are 3 commits to stay inside Github 100 files limit. I had 2 fails in the process, this is why there have "duplicated" commits.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,0 +1,5 @@
+# Import all nodes
+from os.path import dirname, basename, isfile
+import glob
+modules = glob.glob(dirname(__file__)+"/*.py")
+__all__ = [basename(f)[:-3] for f in modules if isfile(f)]
--- a/action_add_group.py
+++ b/action_add_group.py
@@ -1,0 +1,17 @@
+import bpy
+from bpy.props import *
+from bpy.types import Node, NodeSocket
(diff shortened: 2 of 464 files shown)


MoritzBrueckner:
Hi, I think there were some problems with Git, there are copied node files in the Armory root folder (see [this](https://github.com/armory3d/armory/tree/1fe6f7d84a79c595a8d0e9804da70172520b18b6))?

knowledgenude:
Thanks to notice that! I tough that they only were duplicated

I should cancel these: **'Removed logic nodes icons'** and **' " (part 2)'**

Is there some place in the interface to do that? Or only trough terminal? Because i don't did it trough terminal

MoritzBrueckner:
There is probably some way to do this directly on Github but I don't know that. If you're using a graphical Git client there should be a button "Revert commit" somewhere, otherwise use `git revert <commit hash>` in the terminal.

knowledgenude:
I found a way to delete each file from the commit without terminal, but it took too much time. I think is faster to close this and make other PR later, but i will try to do using terminal first. Thanks for the help!

PR #1793, opened 2020-08-07
Set logic nodes icons to none

knowledgenude:
Now i think everything is fine :+1: 

I also removed the "()" from all node labels to speed up searching for them / except for Load URL node


luboslenco:
Thanks!

Issue #1781, opened 2020-08-03
Copy from active don't works for Armory physics properties


knowledgenude:
![copyfromactive](https://user-images.githubusercontent.com/63247726/89202303-1fdc7980-d589-11ea-8d2f-fa69a0d7cd58.jpg)

[copyfromactive.zip](https://github.com/armory3d/armory/files/5017006/copyfromactive.zip)

Only Blender physics properties are copied.

PR #1795, opened 2020-08-09
Copy Armory rigid body settings on Copy from Active

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1781.

It's a bit of a hacky solution because Blender doesn't allow extending already existing operators. There is now a `on_operator_post()` function in handlers.py that allows updating Armory's data after operator executions if the context is correct (that's a Blender limitation). It also gets executed when the operator raises an error which cannot be prevented due to it's hacky nature.


luboslenco:
Thanks!

PR #1797, opened 2020-08-10
Add CopyTraitsFromActive operator

MoritzBrueckner:
Implements https://github.com/armory3d/armory/issues/1783.

This PR adds a new operator to `3DView > Object > Copy Traits from Active Object` and to the search menu that copies all traits and their properties from the last selected object to all other selected objects.

menu

If there are traits on one of the target objects (= a conflict exists), a dialog is displayed that asks if the traits should be overwritten or if they should be kept and the new traits get appended to the end of the trait list:

dialog

Additional changes:
- Fixed a bug in `update_trait_group()` that would happen if an object gets reassigned the same trait again
- Fixed a bug/missing feature in `proxy.sync_collection()` that would not take multiple same traits into account
- Cleaned up some imports

### DIFF
--- a/blender/arm/props_traits.py
+++ b/blender/arm/props_traits.py
@@ -48,7 +48,13 @@ def update_trait_group(self, context):
                     col = bpy.data.collections.new('Trait|' + t.name)
                 else:
                     col = bpy.data.collections['Trait|' + t.name]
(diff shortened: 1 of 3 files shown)


luboslenco:
Awesome!

PR #1808, opened 2020-08-18
Add shader uniform support for the attribute node

MoritzBrueckner:
This PR adds the possiblity to retrieve uniform values via the attribute node. The attribute node now interprets strings of the form
`uniformType uniformName` as a shader uniform. uniformType can be `vec2`, `vec3`, `vec4`, `int` and `float`, although `vec4` and `int` are not likely to be used a lot inside the node editor.

**Example**:
uniform_nodes

This node setup now generates the following output in combination with an animated point lamp:
uniforms


luboslenco:
Awesome!

PR #1810, opened 2020-08-19
Add ShaderData node for easier uniform and input access

MoritzBrueckner:
This PR extends the node support for retrieving uniform values that was merged yesterday by a much more general approach that allows to also retrieve shader inputs from both the vertex shader and fragment shader stages (vertex inputs are passed through to the fragment shader). Thanks to @QuantumCoderQC for the idea of retrieving shader inputs as well!

To accomplish this, I created a new node called `Shader Data` (there was no viewport preview for the attribute node approach so we can use custom nodes instead and if there is a better name for the node, please make a suggestion!) and removed the attribute node support for uniforms again. This new approach also allows to add further sockets and nodes in the future that can handle matrices etc.

I'm not happy with the naming of the UI fields, so if you have better ideas please tell me so that I can change it :) Also the node is currently in a new category called "Armory" to make it easier to find and easier to distinguish from Blender nodes. Is it better to put it into the "Input" category?

Node screenshot

**Example with instance position (`ipos`) as color**:
Example screenshot

### DIFF
--- a/blender/arm/material/arm_nodes/__init__.py
+++ b/blender/arm/material/arm_nodes/__init__.py
@@ -1,0 +1,6 @@
+"""Import all nodes"""
+import glob
+from os.path import dirname, basename, isfile
+
+modules = glob.glob(dirname(__file__) + "/*.py")
+__all__ = [basename(f)[:-3] for f in modules if isfile(f)]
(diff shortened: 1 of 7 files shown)


luboslenco:
Awesome, looks good to me!

PR #1817, opened 2020-08-27
Add sampler2D access to ShaderData node

MoritzBrueckner:
This allows access to envmaps etc., mapping is done using screen coordinates.

@luboslenco Is it possible to somehow access buffers (e.g. the depth buffer) as uniforms in a custom shader without modifying the render path? Me and others would really like to have it available in node materials.

### DIFF
--- a/blender/arm/material/arm_nodes/shader_data_node.py
+++ b/blender/arm/material/arm_nodes/shader_data_node.py
@@ -30,7 +30,8 @@ class ShaderDataNode(Node):
                  ('float', 'float', 'float'),
                  ('vec2', 'vec2', 'vec2'),
                  ('vec3', 'vec3', 'vec3'),
-                 ('vec4', 'vec4', 'vec4')],
+                 ('vec4', 'vec4', 'vec4'),
+                 ('sampler2D', 'sampler2D', 'sampler2D')],
         name='Variable Type',
         default='vec3',
         description="The type of the variable")
@@ -65,6 +66,11 @@ class ShaderDataNode(Node):
     def parse(self, frag: Shader, vert: Shader) -> str:
         if self.input_type == "uniform":
             frag.add_uniform(f'{self.variable_type} {self.variable_name}', link=self.variable_name)
+
+            if self.variable_type == "sampler2D":
+                frag.add_uniform('vec2 screenSize', link='_screenSize')
+                return f'texture({self.variable_name}, gl_FragCoord.xy / screenSize).rgb'
+
             return self.variable_name
 
         else:


luboslenco:
Thanks! Pushed some commits, it should be possible to do `link='gbuffer0'` (or any other render target) to read the texture and `link='gbuffer0_depth'` to read its depth buffer now.

MoritzBrueckner:
That's awesome, thank you!

QuantumCoderQC:
> Thanks! Pushed some commits, it should be possible to do `link='gbuffer0'` (or any other render target) to read the texture and `link='gbuffer0_depth'` to read its depth buffer now.

@luboslenco I am still unable to access the texture or the depth buffer with `link = 'gbuffer0_depth'` using this node. If I am wrong somewhere, could you please clarify how it must be used with this node.

luboslenco:
What happens when you access `link = 'gbuffer0_depth`? Might be something wrong in the code, you can check if the

https://github.com/armory3d/armory/blob/master/Sources/armory/object/Uniforms.hx#L29

and 

https://github.com/armory3d/iron/blob/master/Sources/iron/object/Uniforms.hx#L144

get triggered, and whether the [g.setTextureDepth()](https://github.com/armory3d/iron/blob/master/Sources/iron/object/Uniforms.hx#L145) is called.

MoritzBrueckner:
For me the output is just black, also when trying different buffers. The locations you mentioned are executed, but when printing the target object returned by `textureLink()` in `armory/object/Uniforms.hx`, the `image.pixels` value is null (for `gbuffer0_depth`).

QuantumCoderQC:
@luboslenco I have the same situation. Any ideas on how this could be solved. 

Thanks,
QC

luboslenco:
> the image.pixels value is null

That should be ok, pixels will be non-null when texture pixels have been pulled from gpu via `Image.getPixels()`.

But there are a few problems:

- Right now material nodes are used to create shaders for models, not full-screen passes.
- To map textures to models, texture coords are usually used (or whatever is connected to the `Vector` input in case of `Image Texture` node) instead of screen coords (`gl_FragCoord.xy / screenSize`).
- In deferred renderer, these shaders generated from material nodes are the ones writing to the gbuffer, so there is no way to read it at the same time.

Do you want to read gbuffer depth to create a post-processing pass via nodes? Or do you want to map gbuffer onto the model for some reason?

QuantumCoderQC:
@luboslenco 

Although I am not sure if this is possible, I would like to use the depth texture to create shader effects such as water foam or water opacity based on depth as in this short video [here](https://www.youtube.com/watch?v=Jq3he9Lbj7M). 

Regards,
QC

MoritzBrueckner:
@luboslenco Would it be a solution to copy the depth buffer to another buffer with `copy_pass` if materials declare that they need to read from the depth buffer (once per frame or before every material is drawn, don't know whats better)? Do you know how other engines handle this?

luboslenco:
It might work in a forward renderer with depth prepass enabled that way. You could copy the depth and then read it in geometry pass afterwards (geometry pass = material nodes).

In deferred renderer the material accessing depth buffer in a geometry pass would need to be rendered last, and would either not write itself to the depth buffer or the depth buffer would need to be copied. The common way is to read the depth + gbuffer in a full-screen pass (or things like decal pass), after the geometry pass.

If you just need the depth of the current vertex / fragment in a geometry pass, then there is no need to read the depth buffer and the distance to the camera can be taken.

MoritzBrueckner:
What about sorting the meshes that read from the depth buffer back in the drawing order (I think Godot does this?), like that:

```haxe
public static function sortMeshesDistance(meshes: Array<MeshObject>) {
	meshes.sort(function(a, b): Int {
		return a.cameraDistance >= b.cameraDistance ? 1 : -1;
	});
	meshes.sort(function(a, b): Int {
		// depthRead is calculated each frame based on a new attribute of the material context
		return a.depthRead && !b.depthRead ? 1 : (!a.depthRead && b.depthRead ? -1 : 0);
	});
}
```

With that and a bit of modification of the render loop (only tested with Deferred so far):

```haxe
var readingDepthBuffer = false;
for (m in meshes) {
	// First mesh to access the depth buffer
	if (!readingDepthBuffer && m.depthRead) {
		// Copy the depth buffer so that we can read from it while writing
		// "depth" is a new render target, I couldn't get it working with other targets (?)
		setTarget("depth");
		bindTarget("_main", "tex");
		drawShader("shader_datas/copy_pass/copy_pass");

		// Also, gBuffer0 is probably wrong but it works
		setTarget("gbuffer0");
		bindTarget("depth", "gbufferD");

		readingDepthBuffer = true;
	}

	m.render(currentG, context, bindParams);
}
```

... i'm able to get the following results:

2020-10-21 15_00_01-Window

Below the plane are objects, it displays the depth buffer. Unfortunately I'm not able to clear the depth buffer and I don't know why, so moving the camera is not good currently...

Issue #1813, opened 2020-08-23
On Local Axis checkbox for Location input in Apply Force/Impulse At Location


knowledgenude:
I really don't know why this nodes uses the location relative to the object origin. I think this need to be applied in world location, but even trying to use the relative location, the node don't take in count the object rotation too.

Here is the example file: 
[CarPhysics.zip](https://github.com/armory3d/armory/files/5114912/CarPhysics.zip)

**Press 'Space' to accelerate the car**

It works good if the car is not rotated, but if you rotate the car with the PhysicsDrag, you will able to see what i am saying. I think that the **On Local Axis** checkbox is not working.

**Inside the yellow frame**
The properties X, Y, Z are just coordinates that i put to simulate the ray cast start point relative to car origin. You can remove it and put the vector from ray cast start point and disable 'On Local Axis', but it will work even worst.

QuantumCoderQC:
Hi @knowledgenude 

I devised a simple test to check this issue. The apply force node seems to be working as expected.

About the test: I apply force to a cube on the positive y-axis when 'space'  is held. Then I rotate the object and check which direction it moves depending on whether the local axis flag is set. 

Here is the blend file thas has a small "how-to" in it. Try and let me know if this solves the issue.

[ApplyForceOnLocalAxis.zip](https://github.com/armory3d/armory/files/5115274/ApplyForceOnLocalAxis.zip)

knowledgenude:
Hi @QuantumCoderQC , thanks for the reply

Apply Force works properly, the one i said is **Apply Force At Location**:

[ApplyForceOnLocalAxis.zip](https://github.com/armory3d/armory/files/5118415/ApplyForceOnLocalAxis.zip)

Just use space to apply the force and rotate the object. The red arrow is the part to go up, no matter the object rotation and location. It works only before you change the cube location in the world, but i really can't tell what is happening here. Maybe there is some issue with the object transformation  in both nodes (force/ impulse at location behaviours the same).

This two nodes can be used to make hovering objects, see the example in unreal: https://www.youtube.com/watch?v=ko6lDSSNhV8

QuantumCoderQC:
HI @knowledgenude, I believe the physics or the nodes are not the issues here. The application/understanding is the issue. 

The `location` attribute is the location with respect to the `object `to which force/impulse is applied. **Not** with respect to the `world location`.  The `location` attribute defines the point in space where the force **originates** from, and is described with respect to `object` co-ordinates. 

[...]

PR #1818, opened 2020-08-28
Upgrade Apply Force at location and Apply Impulse at location to include object rotation

QuantumCoderQC:
This PR intends to solve issue #1813 

With this PR,

- Location of Force/ Impulse is always defined with respect to the object location. Object rotation is not accounted for. 
- If `Location On Local Axis` is enabled, object rotation is also accounted for. 
- Force/ Impulse is defined with respect to the global axis. 
- If `Force/ Impluse On Local Axis` is enabled, Force/ Impulse will be applied with respect to object local axis. 

Capture


luboslenco:
Thanks!

Issue #1815, opened 2020-08-27
Android - Setting window orientation not working


Dok-Sergey:
Thank you for contributing to Armory!

**Description**
It doesn't matter which option you select in the _Armory Project_ - _Window_ - _Orientation_ settings, as a result of Android (C) publishing in the _AndroidManifest.xml_ file, the following will be set: 
android: screenOrientation = "sensor"

**To Reproduce**
Build a project using Android (C) as the target.

**Expected behavior**
Waiting for the selected option (portrait, landscape, sensor) to be installed when publishing.

**System**
Blender: 2.83.4
Armory: git, master (Aug 19, 2020)
OS: Windows 10 64-bit
Graphics card: NVIDIA GeForce 930M

MoritzBrueckner:
Hi,

can you change `project.targetOptions.android` to `project.targetOptions.android_native` in `blender/arm/write_data.py`, line 290:

https://github.com/armory3d/armory/blob/0d6422f77a485caba5f47c10bd3d54efbaa1347f/blender/arm/write_data.py#L286-L290

Then restart Blender. If it works, feel free to open a pull request (Armory doesn't support the Android java target afaik so it should be ok to just replace those lines) or I can do that if you want.

Dok-Sergey:
@MoritzBrueckner, thanks It works.
Better to post the revision to you.

PR #1819, opened 2020-08-28
Fix android target options

MoritzBrueckner:
This fixes https://github.com/armory3d/armory/issues/1815.

As Armory only supports the native android target, we can change `android` to `android_native` in the khafile (see https://github.com/Kode/kincmake/pull/11).

I've also changed `f` to `khafile` in the code for better readability.


luboslenco:
👍👍

PR #1821, opened 2020-08-30
Improved quaternion and angle handling in logic nodes (+2bugfixes)

niacdoial:
- Made so all nodes outputting a quaternion object also output it as a XYZ (vector) + W (float) combination
- Modernized the interface of the "Action/Rotate Object" node, to align on the newer "Action/Set Rotation" node interface  ("Action/Rotate Object Along Axis" is now depreciated, but still usable)
- Fixed a blender-side-only bug with the "Logic/Switch" node (...which technically could have lead to a compile-time problem if exploited the right way)
- Fixed a bug on the "Action/Set Rotation" node: now, quaternion input is automatically normalized in order to avoid accidental scaling
- Added a "Value/Separate Quaternion" node
- Made so the names of some sockets change in the "Set Rotation" and "Rotate Object" nodes, so they adapt to those nodes' input types.
  (Same thing with "Value/Vector From Transform"'s output type)

### DIFF
--- a/Sources/armory/logicnode/QuaternionNode.hx
+++ b/Sources/armory/logicnode/QuaternionNode.hx
@@ -1,6 +1,7 @@
 package armory.logicnode;
 
 import iron.math.Quat;
+import iron.math.Vec4;
 
 class QuaternionNode extends LogicNode {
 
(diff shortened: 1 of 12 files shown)


luboslenco:
Thanks!

PR #1826, opened 2020-08-30
Lightmapper

Naxela:
Hi,

I think lightmapping is reaching a point where it's sufficiently stable to be available for Armory3D as an included option. 

## The Lightmapper

billede

Currently, I've just implemented it as a new tab in the **Armory Bake** menu, with the individual object settings available under the **Armory Props** menu, so to keep the UI clutter to a minimum:

billede

billede

I tried to keep the implementation as simple as possible, with the files being in their own "lightmapper" folder, and the only affected arm files are props_ui.py and props_bake.py - That way it should be very easy to keep feature parity with the standalone version:

billede

I plan to add some documentation and tutorials, although I hope the setup is somewhat easy to figure out.


luboslenco:
Wohoo 🚀

PR #1828, opened 2020-09-01
Fix for OpenCV installation

Naxela:
A minor fix for getting OpenCV to work, plus removed unnecessary preference files


luboslenco:
👍👍

Issue #1830, opened 2020-09-02
Rigid bodies parenting issue


knowledgenude:
The object comes to its original position/rotation after clear parent
[parent_test.zip](https://github.com/armory3d/armory/files/5163342/parent_test.zip)

Left object is with active physics, right object is without physics

Use **WASD** to rotate the parent, **SPACE** to toggle both objects parent mode

There is also an issue when you uncheck **Keep Transform**. For me the object should return to its original pos/rot after clear parent (same as active body is doing in the issue), anyway i am not familiar to transforms work. Is not better to remove "Keep  Transform"? Maybe is better to store the object transform trough another way (in variable for example) and after remove its parent, set the desired transform (loc or rot)

knowledgenude:
There is also an issue happening when the parent is set:

![screenshot](https://user-images.githubusercontent.com/63247726/92268476-d5685880-eeb8-11ea-98c9-86081b1a02f2.jpg)

[parent_test.zip](https://github.com/armory3d/armory/files/5176117/parent_test.zip)

PR #1833, opened 2020-09-03
Apply tansform for RigidBody if Keep Transform

QuantumCoderQC:
This PR partially solves #1830 . `RigidBody` now obeys `keep Transform`. 

However, when `keep transform` is disabled, the object transform does not return to the original transforms. This must be fixed in [Iron](https://github.com/armory3d/iron.git).


luboslenco:
Thanks!

PR #1836, opened 2020-09-03
Rename camera nodes (no longer duplicate names)

MoritzBrueckner:
Previously there were two nodes equally named `Set Camera` which are now renamed to `Set Active Camera` and `Set Camera Postprocessing` (their get-parts were also renamed to reflect that change). I didn't rename the classes, idnames and files to keep compatibility but in the future they should be renamed too.


luboslenco:
👍👍

Issue #1780, opened 2020-08-03
Ray cast ignores collision mask


knowledgenude:
[collisionmask.zip](https://github.com/armory3d/armory/files/5016964/collisionmask.zip)

-- Red objects are from group 1 -- detected by physics ray
-- Green objects are from group 2 -- not detected by physics ray

Issues: 
1. **Group 0** don't exists, it shouldn't be in Cast Physics Ray node

2. **Cast Physics Ray node** only works if searching for **group 1**. If  you put **group 2** or whatever, it will not work. Same for **PhysicsDrag** and probably for every trait that use **ray cast**.

knowledgenude:
Fixed https://github.com/armory3d/armory/pull/1841

PR #1841, opened 2020-09-05
+3 physics nodes

knowledgenude:
* Set Friction Node

* Set Activation State Node

* Remove Physics node

* Fixed Ray Cast mask: https://github.com/armory3d/armory/issues/1780


QuantumCoderQC:
Hello,

Thank you for the fixes. I would like to suggest a few changes for this PR

1. Remove Rigid Body node: I believe you need to remove the Rigidbody trait from an object instead of calling `removeFromWorld()`. Otherwise the trait might still be applied on the object but rigid body does not. And this is not appropriate.

you might need to call `object. getTrait(RigidBody).remove`. This removes both trait and RB.
[implemented remove trait](https://github.com/armory3d/armory/blob/5c1b55d22163d48ad4cf4120de8d3afbc22c5b21/Sources/armory/trait/physics/bullet/RigidBody.hx#L287)

knowledgenude:
@QuantumCoderQC , I don't know what is a trait? Is the rigid body settings like mass, etc.? Do Remove From World keep some settings into the object? Maybe it is good when you need to remove the collision of the object and reapply it again with the same settings after sometime, but my intention is to remove the physics of the object. Maybe we need both kind of nodes?

There is a small example of use of the nodes:

Set Friction: a cube that starts with the material "earth" and the material is changed to "ice", then the rb friction must be reduced.

Set Activation State: it is basically a smaller version of "set linear/angular factor 0.000". It also can be used to stop the object in air but reactive to physics, like in this example, where all boxes have activation state 0 on init: 
[setactivationstate.zip](https://github.com/armory3d/armory/files/5179685/setactivationstate.zip)

I also hope for a Get Activation State, it is the better way to know if an object is moving, because with Get Velocity, if the object hit a wall, the velocity will be 0 but it was not stoped properly.

screenshot1

The only one i can't think in a use is Sync Transform, i expected this behavior: when you scale a cube to 0.5, the cube keeps flying in the air until receive some force. I tough that sync transform force some update on the cube scale, but not.

Anyway i am ok if this nodes don't are useful, i think that there is not too much nodes related to physics, but i am just trying to understand it more. I am grateful for tips.

One more question here is what is this numbers in in the end of this lines in Ray Cast node? I removed  them and no effect

>>v.set(hitPointWorld.x, hitPointWorld.y, hitPointWorld.z, 1);

>>v.set(hitNormalWorld.x, hitNormalWorld.y, hitNormalWorld.z, 0);

QuantumCoderQC:
@knowledgenude 
>> I don't know what is a trait? Is the rigid body settings like mass, etc.? Do Remove From World keep some settings into the object?

When a RigidBody is added to an object, the RigidBody trait will be applied to it. This contains all the information and parameters about the RigidBody. 

>>Maybe it is good when you need to remove the collision of the object and reapply it again with the same settings after some time, but my intention is to remove the physics of the object.

If the RigidBody will be removed and the trait kept as is, then when adding the RigidBody again later will retain the same information, position, rotation, etc. even if the object in the scene was moved. This will cause strange behaviour of objects later. 

>>Maybe we need both kind of nodes? 

I guess that will only add up to the confusion, having two different nodes do the same thing. I highly recommend only implementing `object.getTrait(RigidBody).remove()` as a node. This will remove both physics and the trait. Then later when re-adding the rigid-body, both the RigidBody and the trait will be added with the updated position, rotation and velocities. 

>>One more question here is what is this numbers in in the end of this lines in Ray Cast node? I removed them and no effect.

The position of the hit-point is the position in world-coordinates where the ray first contacts the object.
The normal of the hit-point is a vector pointing in the direction of normal of the surface the ray hit.

knowledgenude:
@QuantumCoderQC , ok thanks for the clarification. I already replaced `Remove From World` by `Remove Physics`and i think now`Set Activation State` performance was improved a bit, as the state number will be checked together with the rigid body.

* I have only a few questions about what i do recently:

1. Should i remove `Sync Transform` or it can have some use (now or in the future)?

2. Should i keep `Set Activation State `after the example i send?

3. From performance view, should i check in the `Set Parent Node` if the rigid body is `static` and if it is, i will not set its state? Sounds redundant to remove the motion of a object that don't have it.

If someone can clarify anything, i would like to leave this nodes in "perfect" state and apply this practices in future nodes. Anyway i will follow how the already existent nodes do things.

QuantumCoderQC:
@knowledgenude 

1. Right now I don't know if we have any use for the syncTransform node. For now, to keep things clean, I suggest removing it. We can add it later if such a requirement arises. 

2. The example you stated for SetActivationNode is indeed convincing. I am okay with keeping it. It might also help improve performance when we have a large world with many physics objects in the scene.

On the side note, it would be better to have a drop-down selection in this node(in the node-python side) instead of an integer input. Of course, this will make it less dynamic(cannot provide inputs), but also improve performance(since if statements will be reduced) and will be more intuitive(having activation state names instead of integers). What do you think about it?

3. The performance will degrade with the number of branching statements, such as`if( )` and not the actual functions. Hence I suggest not adding a check for static RigidBodies in the SetParentNode.

knowledgenude:
Done. Thanks, @QuantumCoderQC !

screenshot1

knowledgenude:
> 3. The performance will degrade with the number of branching statements, such as`if( )` and not the actual functions. Hence I suggest not adding a check for static RigidBodies in the SetParentNode.

Considering this an unexpected situation, and that nodes like `Set Velocity` can be called in every frame, is not better to remove `if (rigidBody == null)` from all of them and introduce a node called `Is Rigid Body` for such specific situations?

It also will be useful to remove objects that have rigid body or not from the scene for example.

QuantumCoderQC:
@knowledgenude

>>is not better to remove if (rigidBody == null) from all of them

Yes, that might actually be better. In fact, that will also help to debug the game. Say if force /velocity was applied and you are not sure if the object made any changes, there is no indication/warning that there is no rigid body.

However, removing the if statement gives an runtime error, if no rigid body is present.

luboslenco:
Thanks!

Issue #1835, opened 2020-09-03
Logic node improvements


MoritzBrueckner:
Hi,

this issue is a continuation of the discussion started in https://github.com/armory3d/armory/issues/1786.
In my opinion, there a quite a few things about logic nodes that need discussion and very likely some improvements. I'm collecting them in this issue:

- ### There are multiple similar nodes where the differences might not be very clear
  Do you know, without looking, what the difference between "Array Remove" and "Array Remove Value" is? Or what is the difference between "Collection", "Set Collection", "Add Collection" and "Spawn Collection"?

  ### So, some nodes need a renaming
  For example, there are two `Set Camera` nodes with different functions. It would be better to name them "Set Camera" and "Set Camera Settings" or "Set Camera Data".

- ### Some of the larger categories like `Action` and `Value` urgently need to be splitted in smaller ones
  Scrolling in Blender's menus is a pain^^

- ### Set and Get nodes for the same thing are in different categories
  I think it would be better to sort the nodes not according to their kind of function (get, set, etc.) but to their subject like it is done for some categories already (Canvas, Physics, Postprocess etc.). Also, the "Variables" and "Values" categories are hard to distinguish. We could use something like this as an inspiration ([Source](https://animation-nodes-manual.readthedocs.io/en/latest/user_guide/getting_started/interface.html#node-menu)):

  ![Screenshot of a node menu with many small categories with icons](https://animation-nodes-manual.readthedocs.io/en/latest/_images/node_menu.png)

  If implemented exactly like in the picture above (with icons and layout separators) that would need a custom drawn menu but it is possible ([source code](https://github.com/JacquesLucke/animation_nodes/blob/master/animation_nodes/ui/node_menu.py)) and it probably would also allow us to use the node's `bl_description` values as tooltips when hovering. Currently, Blender only shows `Add a node to the active tree` as a tooltip. If we decide to implement such a menu with icons, we can also update the nodes to show the correct category icons in their header. We maybe even could add sub-categories that are separated by those separator lines like in the screenshot.

  Changing the categories and/or the node names should not destroy already existing node setups so I don't think that there are big backwards compatibility issues.

- ### The documentation page in the wiki should be sorted regarding to the categories.
[...]

PR #1851, opened 2020-09-09
Logic nodes overhaul 

MoritzBrueckner:
This PR introduces a redesigned logic node menu and many changes to the logic node system like discussed in https://github.com/armory3d/armory/issues/1835.

**Notable changes**:
- New node menu with sections and icons:

  New logic nodemenu
- Nodes automatically use the icon of their category in their header
- A bunch of new categories to clean-up the obverstuffed categories
- Group nodes together by their subject, not by their technical purpose
- Categories now are in their own subpackage
- New API for logic nodes and extension packs to sort nodes into categories/sections, including automatic category-selection by the module's name
- Some smaller node renamings, one bigger renaming: the `Mix` node is now called `Interpolate`
- New custom node search operator to allow search improvements in the future
- Node socket definitions are now in their own module
- Removed an obsolete eye dropper operator for selecting objects that is now a Blender feature
- Fixed docstring format for logic nodes (`"""` instead of `'''`, the latter isn't used for docstrings according to PEP8)
- `ArmLogicTreeNode` is now a subclass of `bpy.types.Node`
- Use a custom socket shape for sockets which can be accessed as variables by the `Set Variable` node to compensate for the now splitted `Variable` category:

  Example screenshot

  @luboslenco Should variable nodes implement a variable type interface to ensure they have `set`and `get` methods or is this over-the-top and just makes noise?

- More streamlined way of adding sockets to a node
- Tooltips for nodes (most nodes don't have a descriptive tooltip yet):

  nodes_tooltips

- Fixed a warning described [in the forum](https://forums.armory3d.org/t/system-console-error/4204).

**This PR does not break compatibility with older projects.**


luboslenco:
It's awesome!

PR #1852, opened 2020-09-09
New nodes

knowledgenude:
* `Get Mesh Culled` can be used to make flashbangs, jumpscares, doors, etc.

There is a reason to get if the shadow is culled. In The Witcher 3 for example, there is an optimization that stops the characters animations when they are offscreen. But if you see the character shadows, you can see that the shadows are also stopped. In some cases it can not be good.

Small test file: [visible_test.zip](https://github.com/armory3d/armory/files/5197443/visible_test.zip)

* Other nodes related to visibility

`Is Rigid Body` is to avoid the crash when you try to apply physics in objects that don't have it. But of course can be used to filter if an object is or not rigid body.

`Get Rigid Body Properties` can be used to clone physics atributes from one object to another, or simple get this kind of info. There is a version of it separated [single_rigid_body_attributes.zip](https://github.com/armory3d/armory/files/5197449/single_rigid_body_attributes.zip)

Unfortunately, i am not able to make this work properly: [not_functional.zip](https://github.com/armory3d/armory/files/5197450/not_functional.zip). These are nodes to set the rigid body attributes, but they don't affect bullet engine. If you print the value you set, it is working, but not for bullet.

* Fix for https://github.com/armory3d/armory/issues/1830. Now is possible to set active rigid bodies parents. Unfortunately, if the rigid body is animated, this property will be lost in the process. I don't know how much this can be a problem.


Sources/armory/logicnode/GetMeshCulledNode.hx
```diff
@@ -0,0 +1,18 @@
...
+class GetMeshCulledNode extends LogicNode {
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function get(from: Int): Dynamic {
+		var object: Object = inputs[0].get();
+
+		if (object == null) false;
```
MoritzBrueckner, line 14: This line doesn't make a lot of sense, did you mean `return` instead of `false`? The same goes for the other nodes.

Sources/armory/logicnode/GetRigidBodyAttributesNode.hx
```diff
@@ -0,0 +1,32 @@
...
+
+	override function get(from: Int): Dynamic {
+		var object: Object = inputs[0].get();
+
+		if (object == null) false;
+
+		#if arm_physics
+		var rigidBody = object.getTrait(RigidBody);
+
+		if (from == 0) return rigidBody.group;
```
MoritzBrueckner, line 20: I would use a switch-case statement here to speed things a bit up, like this:
```haxe
return switch (from) {
	case 1: rigidBody.mask;
	case 2: rigidBody.animated;
  ...
  default: null;
}

MoritzBrueckner (changes requested):

MoritzBrueckner:
Awesome!

Sources/armory/logicnode/GetMeshCulledNode.hx
knowledgenude: It is because i made this node based in `Get Object Visible`, is not the same method?

Sources/armory/logicnode/GetRigidBodyAttributesNode.hx
knowledgenude: Okay i will fix this too. It is also because i based it in `Get Linear Velocity`. Thanks for the tips!

Sources/armory/logicnode/GetMeshCulledNode.hx
MoritzBrueckner: I think it's a mistake there, too. This statement just doesn't have any effect whatsoever and is probably even optimized away by the compiler.

Sources/armory/logicnode/GetRigidBodyAttributesNode.hx
MoritzBrueckner: Both ways work, but for larger "sets" of values switch statements are often the better choice as they usually have a O(1) (constant-time) lookup due to a lookup table instead of if-statements where each statement has to be evaluated in order. For only two values it's not a big difference

Sources/armory/logicnode/GetMeshCulledNode.hx
knowledgenude: I also have fixed Get Visible. Thanks to notice it.

knowledgenude: When i change the line `if (object == null) return false;` to `if (object == null) return;`, i got `Void should be Dynamic` error. What is wrong?

knowledgenude: The same happens to `Get Visible`. I think there is a reason why it is with `false` instead of `return`

MoritzBrueckner: Oh, my bad. Then it is `return false` instead of just `false`. `return` would just return nothing (`Void`) but the function signature requires that something is returned.

knowledgenude: Alright, i will fix it. Thanks :+1:

Issue #1558, opened 2020-01-31
FR: Set Visible Node: add "children" checkbox


blackno666:
I'd be more than helpful if the "Set Visible" node had a checkbox "children" (maybe even the option for recursion).

knowledgenude:
This workaround unexpectedly works: 
[getchildren.zip](https://github.com/armory3d/armory/files/5032040/getchildren.zip)


![getchildren](https://user-images.githubusercontent.com/63247726/89480449-bcf70800-d76b-11ea-9a21-73cff34b8f85.jpg)

When children prefix is blank, even the parent will be set to invisible, i don't know why.

knowledgenude:
@MoritzBrueckner , what is better here? A checkbox or an array? With array it can be used to hide an collection

MoritzBrueckner:
I would go for a checkbox. Most of the time this node will probably be used with single object, so people don't want to put that in an array first. If you need to make an entire collection visible, you can either loop through that array or you can make the collection's root object visible (+ children).

knowledgenude:
Closed via https://github.com/armory3d/armory/pull/1864

PR #1856, opened 2020-09-10
Improved nodes and +3 new nodes

knowledgenude:
Now i think this PR makes more sense

* Improved `Get Visible` and `Set Visible` nodes.
* Added `Is Rigid Body`, `Get Object Offscreen` and `Get Rigid Body Data` nodes.
* Temporary fix for: https://github.com/armory3d/armory/issues/1830
* Implemented Children checkbox in `Set Visible` node thanks to @QuantumCoderQC [here](https://forums.armory3d.org/t/help-with-arrays/4207/5) / Closes: https://github.com/armory3d/armory/issues/1558

Preview: 
screenshot1

Test file: [visible_test.zip](https://github.com/armory3d/armory/files/5202475/visible_test.zip)

### DIFF
--- a/Sources/armory/logicnode/ClearParentNode.hx
+++ b/Sources/armory/logicnode/ClearParentNode.hx
@@ -19,9 +19,13 @@ class ClearParentNode extends LogicNode {
 
 		#if arm_physics
 		var rigidBody = object.getTrait(RigidBody);
-		if (rigidBody != null) rigidBody.syncTransform();
+
+		if (rigidBody != null) {
+		rigidBody.animated = false; // Workaround
+		rigidBody.syncTransform();
+		}
 		#end
-		
+
 		iron.Scene.active.root.addChild(object, false);
 
 		runOutput(0);
--- a/Sources/armory/logicnode/GetObjectOffscreenNode.hx
+++ b/Sources/armory/logicnode/GetObjectOffscreenNode.hx
@@ -2,7 +2,7 @@ package armory.logicnode;
 
 import iron.object.Object;
 
-class GetVisibleNode extends LogicNode {
+class GetObjectOffscreenNode extends LogicNode {
 
 	public function new(tree: LogicTree) {
 		super(tree);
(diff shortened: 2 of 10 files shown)


Sources/armory/logicnode/ClearParentNode.hx
```diff
@@ -19,9 +19,13 @@ class ClearParentNode extends LogicNode {
 
 		#if arm_physics
 		var rigidBody = object.getTrait(RigidBody);
-		if (rigidBody != null) rigidBody.syncTransform();
+
+		if (rigidBody != null) {
+		rigidBody.animated = false; // Workaround
```
QuantumCoderQC, line 24: @knowledgenude , Thank you for the fixes,
I do not fully understand the use of this line here. What is its purpose? 
Setting `rigidBody.animated = false` only affects the trait and not the actual bullet rigid body.

knowledgenude:
@QuantumCoderQC , this is not a correct fix, but it is the only way i found to get `Set Parent` working with rigid bodies. Note that i Set the animated setting trough `Set Parent` and in `Clear Parent` i have to disable animated or the rigid body will be not "active".

Sources/armory/logicnode/ClearParentNode.hx
knowledgenude: @QuantumCoderQC , this is not a correct fix, but it is the only way i found to get Set Parent working with rigid bodies. Note that i Set the animated setting trough Set Parent and in Clear Parent i have to disable animated or the rigid body will be not "active".

QuantumCoderQC:
@knowledgenude Okay. Now I understand the purpose.

But if I may, in general, the RigidBodies should not, and cannot be a child object. If such a situation arises where parenting is necessary, the RigidBody trait must be removed, and then after clearing the parent, the trait is added back. This is my opinion.

knowledgenude:
@QuantumCoderQC I also have this opinion, but how is the best way to do that? Because if i remove the rigid body with Set Parent, i need to store its settings in same way that can be applied bacl in Clear Parent. Is possible to simple disable its collision and activation? Activation yes, but even collision mask is not possible currently

QuantumCoderQC:
@knowledgenude  

> Edit: for me i would simply remove the rigid body reference from both nodes and let the users to remove the rigid body themselves and re-add in the moment they want.

Yes. This is better. I am already working on nodes to dynamically add and remove rigid bodies. If I get some time, I will make it happen and create a pull request. Then we can have a discussion about the workflow for parenting rigid bodies. 

Let me know if you agree with this.

knowledgenude:
@QuantumCoderQC , i totally agree :+1: 

I will keep this PR as it is for and i will fix it when you comes with the solution.

PR #1857, opened 2020-09-11
Blender UI – Armory Trait – Logic Node

e1e5en-gd:
Add buttons:
- New Node Tree – creates and sets a new Node Tree in the list;
- Edit Node Tree – loads the Node Tree selected from the list into all Logic Node Editor;
- From Node Tree – loops through all the working Logic Node Editor and from the last in the list sets the value to the list.

done_v3


luboslenco:
Many thanks for contributing!

PR #1858, opened 2020-09-11
Add "armory" flag on export to differ between Armory and Kha

MoritzBrueckner:
Armory's Kha is usually a bit behind the Kode/Kha version which can make it a bit complicated in some cases for libraries to support both versions. For example, Kha replaced `PipelineCache` with `SimplePipelineCache` (the first is now an interface) but Armory still uses the old version.

This PR adds an "armory" flag so it's easy to just check if Armory is used or not. This also makes it possible to check whether Iron is available.

### DIFF
--- a/blender/arm/write_data.py
+++ b/blender/arm/write_data.py
@@ -51,6 +51,9 @@ let project = new Project('""" + arm.utils.safestr(wrd.arm_project_name) + """')
 project.addSources('Sources');
 """)
 
+        # Let libraries differentiate between Armory and pure Kha
+        assets.add_khafile_def('armory')
+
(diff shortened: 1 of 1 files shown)


luboslenco:
Thanks! This might be also worth a khamake bug report. It is supposed to auto-create this define at:
https://github.com/Kode/khamake/blob/master/src/Project.ts#L190

But the problem is we use absolute path to add armory as a library in khafile:

> project.addLibrary("D:/dev/armsdk/armory");

And the hxml define then ends up look as:

> -D D:/dev/armsdk/armory

Instead of a plain:

> -D armory

Issue #1842, opened 2020-09-06
 When the “init” node connects to the “canvas” node and then runs the project, the screen is black 


haizhou37:
**os:deepin 20
cpu:ryzen3 2200g
blender:2.83.5
armorysSDK:2020.9
Even if you create a default new project run, this problem still occur**

![截图录屏_Blender_20200906182055](https://user-images.githubusercontent.com/33272474/92323795-2d948b80-f06e-11ea-8fad-0f809150b0c7.gif)

MoritzBrueckner:
Is there an error when you start Blender from a terminal?

This might be related to https://github.com/armory3d/armory/issues/1181

haizhou37:
> Is there an error message when you start Blender from a terminal and run the game?
> 
> This might be related to #1181

![截图录屏_选择区域_20200909184648](https://user-images.githubusercontent.com/33272474/92589538-45b61600-f2cd-11ea-8238-06b47646e510.gif)

MoritzBrueckner:
Do you have the canvas trait on an object other than the active camera? If yes, thats the problem and moving it to the active camera or changing the trait to a scene trait will solve the issue.

Nonetheless this should not happen. We should probably restrict Canvas traits to scenes to prevent this. Even if restricted to scenes *and* cameras this could cause problems when the active camera is changed.

haizhou37:
> Do you have the canvas trait on an object other than the active camera? If yes, thats the problem and moving it to the active camera or changing the trait to a scene trait will solve the issue. Thanks for reporting this!
> 
> Nonetheless this should not happen. We should probably restrict Canvas traits to scenes to prevent this. Even if restricted to scenes _and_ cameras this could cause problems when the active camera is changed.

Do you mean as shown in the picture? But it's still going to be black
![截图录屏_Navigator_20200910203125](https://user-images.githubusercontent.com/33272474/92729252-a5332500-f3a4-11ea-9738-000c7c03b445.png)

MoritzBrueckner:
Did you run a clean build? I can only reproduce this when the canvas trait is on another object other than the active camera or a scene.

If a clean build doesn't help, can you send a test project (including the canvas file) please?

haizhou37:
> Did you run a clean build? I can only reproduce this error when the canvas trait is on another object other than the active camera or a scene.
> 
> If a clean build doesn't help, can you send a test project (including the canvas file) please?
[...]

PR #1859, opened 2020-09-11
Fix canvas nodes for invalid element names

MoritzBrueckner:
This fixes https://github.com/armory3d/armory/issues/1842.

If the element socket of some canvas nodes had an empty name or the element doesn't exist, the game would show a black screen. Now, if an element doesn't exist for a `Set` node, the output is still run but nothing happens. For `Get` nodes, the output is not run at all if the element doesn't exist to prevent the usage of wrong values left over by old elements.


luboslenco:
Thanks!

PR #1868, opened 2020-09-14
Add poll() function to AddNodeOverride operator

MoritzBrueckner:
This prevents an error when trying to add a node without a logic tree.


luboslenco:
Thanks!

PR #1869, opened 2020-09-15
Logic Node - On Swipe

e1e5en-gd:
The node for the swipe event (section "_Input_").
on_swipe_v3

Input parameters:
- _Time_ - duration of the swipe (default value 0.15 s);
- _Min Length (px)_ - the minimum length that the swipe must “pass” (default value 100 px).

Output parameters:
- _Direction_ - normalized vector of the swipe direction;
- _Length (px)_ - the length of the swipe (to assess the “strength”);
The Length output parameter is calculated by the vector value (without normalization) to relate to the input value.
- _Angle (0-360_) - swipe angle in the range from 0 to 360 degrees (0 is on the right, i.e. Vector (1, 0) = 0 degrees).
- _States_ are states with direction output (Up, Down …). By default, they are not displayed, but if you press “Add 4 States”, then 4 states will appear (Up, Down, Left, Right) and if again, then 8 states will already be displayed.
Determining direction for a state:
for 4 directions - "circle" is divided into 4 segments of 90 degrees;
for 8 directions - "circle" is divided into 8 segments of 45 degrees.
on_swipe_4_8

### DIFF
--- a/Sources/armory/logicnode/OnSwipeNode.hx
+++ b/Sources/armory/logicnode/OnSwipeNode.hx
@@ -1,0 +1,201 @@
+package armory.logicnode;
+import iron.math.Vec3;
+import iron.math.Vec2;
+
+class OnSwipeNode extends LogicNode {
+
+	var point_start = new Vec2();
+	var point_end = new Vec2();
+	var direction = new Vec2();
+	var length = 0;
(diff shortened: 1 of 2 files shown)


luboslenco:
Thanks for contributing!

PR #1874, opened 2020-09-17
Support for LZ4 compression

### LINKED ISSUE #1779: Error publishing with Asset Compression option
**Description**
If you enable this option, then when publish, for example, in HTML5, the scene will be unloaded as "Scene.arm", but the engine is looking for the "Scene.lz4" file.
I assume there is a similar problem for other platforms.

kha.js:
Error_asset_compression_part

**System**
System
Blender: 2.82
Armory: git (from 28 July)
OS: Win 10

MoritzBrueckner:
Fixes #1779.

Link to https://github.com/armory3d/iron/pull/102

This ports the LZ4 encoding from Iron to Python. It is a bit slow (maybe due to the extensive use of numpy, I will run it with a profiler in the next days) so a small warning is printed to the console when compression is used.

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -2003,6 +2003,8 @@ class ArmoryExporter:
         """Exports the scene."""
         profile_time = time.time()
         print('Exporting ' + arm.utils.asset_name(self.scene))
+        if self.compress_enabled:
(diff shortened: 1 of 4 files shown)


luboslenco:
Many thanks!

PR #1876, opened 2020-09-18
On Tap Screen - node for handling touches (double, triple…)

e1e5en-gd:
on_tab_screen_v4

Input parameters:
- _Duration_ - time during which the number of touches is counted (default value is 0.3 seconds);
- _Interval_ - the time that must elapse between touches (default value is 0);
- _Repeat_ - the number of touches that need to be done to execute the event (default value is 2).

Output parameters:
- _Done_ - event for a successful completed action;
- _Fail_ - an event for a failed action;
- _Tap_ - event on each touch;
- _Tap Number_ - how many times the user has already touched the screen;
- _Coords_ - coordinates of the touch.

### DIFF
--- a/Sources/armory/logicnode/OnTapScreen.hx
+++ b/Sources/armory/logicnode/OnTapScreen.hx
@@ -1,0 +1,108 @@
+package armory.logicnode;
+
+import iron.math.Vec4;
+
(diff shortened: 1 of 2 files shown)


luboslenco:
Thanks!

PR #1880, opened 2020-09-20
Logic nodes improvements

knowledgenude:
* Nodes sockets now have more specific names.

* Values from nodes like `Vector Math` has been reset, because when we need to add a value only for one axis, we have to reset the two other axis every time.

* Separated `Mouse Coords` into `Get Mouse Movement` (with multiplier) and `Get Cursor Location`. The reason for this is that the mouse coords are frequently used into objects that have 3 axis, and this requires separate XYZ everytime and a multiplier node to adjust the "sensibility". The same for `surface` nodes.

* Merged `mouse lock` and `mouse visible` nodes. Most of the times this nodes are used together (only in very specific time we need to just hide or just lock the mouse).

Preview:

test

Screenshot from 2020-09-20 14-14-33

### DIFF
--- a/Sources/armory/logicnode/GetCursorLocationNode.hx
+++ b/Sources/armory/logicnode/GetCursorLocationNode.hx
@@ -1,0 +1,24 @@
+package armory.logicnode;
+
+import iron.math.Vec4;
+
+class GetCursorLocationNode extends LogicNode {
+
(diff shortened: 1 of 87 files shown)


luboslenco:
Thanks!

PR #1887, opened 2020-09-24
Rename logic nodes labels

knowledgenude:
* This PR renames the envolved .py files. I don't have sure if this can break the compatibility.

*  Every label changed has the purpose of improving some aspect. Generally, my intention is to know what each simple node do just by reading its name and maintain a consistency between all nodes (because they are named by different authors).

* Now all nodes names follows the same principle. To take advantage of the new way of organizing nodes, alphabetical order is no longer taken into account for nodes such as canvas, render path, etc. This allows the user to maintain the same way of searching (in this case get or set / from where / which value. This will avoid further embarrassment.

* The nomenclature of nodes is not fully defined and it is open to discussion on this topic: https://forums.armory3d.org/t/do-you-aprove-these-logic-node-changes/4216/15

* I returned the name from Interpolate to Mix node, just to be equal to Vector Mix. I think if it is to change a name, it is both to keep the default and "Interpolate Vectors" don't sounds friendly (but somehow I find it interesting for the user to know what interpolation means).

* Minor typo fixes.

Note: as soon as I am given permission or there is a way to do this without breaking compatibility, I can rename the .hx and text files.


MoritzBrueckner:
As long as you don't change the `bl_idname` there should be backwards compatibility :)

"Delete" and "remove" suggest some different purposes in my opinion, "delete" sounds a lot more like the action is irreversible. For example, when you delete a scene it sounds like the scene is just wiped out of existence from the game, "remove" sounds a bit softer and somewhat suggests that you can add it again later (delete=destroy, remove=take it away). But that's just my opinion, I'm not a native english speaker as well.

Renaming the .hx files alone does not break compatibility in theory but then you would have to change the `bl_idname` attribute (the .hx name is derived from the idname) which then breaks compatibilty... @niacdoial Does your node replacement system allow for replacing nodes with different idnames? That's what the old unused system was all about and it would also enable us to automatically replace the now deprecated input nodes by their successors.

niacdoial:
I know this merge has been closed, but @MoritzBrueckner, yes it allow that. However, you need to declare a "dummy" node class with the old bl_idname (and a high enough version number), which indicates how it is supposed to be replaced.
(I would have loved to not have to declare that, but it turns out you can't read the bl_idname in a node without a registered class)

PR #1891, opened 2020-09-26
Further handling of node depreciation

niacdoial:
Implemented node replacement mechanisms for some of the depreciated nodes
(plus a few fixes elsewhere)

Do you think it is appropriate to move all depreciated nodes in a separate subdirectory, like I did?
[that question is the only thing that makes this pull request a drafted one]


niacdoial:
....Oh. I have to be the one marking it ready for review. I guess I misunderstood how that thing worked.

MoritzBrueckner:
Looks good to me, there are only a typos in the node labels (depreciated instead of deprecated). What do you think, how long should we keep deprecated nodes around?

niacdoial:
> depreciated instead of deprecated
oh. well, that's what I get for being french I guess.

> how long should we keep deprecated nodes around?
well, the limit here is that projects that are saved less regularly than that end up broken.
So, maybe a 6 months to a year would be good? My guess is that if a project is left alone for that long, then picked back up, then the author would expect some breakage anyway.

knowledgenude:
My opinion n is that we should take advantage of Armory instabilities to remove, merge or separate the necessary nodes to get a "stable" version of all of them as soon as possible.

Nodes seems to be the more sensible feature in armory in terms of compatibility and they are a good reason for use Armory.

They need a general review and unfortunately at the same time, i don't like the idea to possible break other people project, but in some cases this is necessary to avoid future work (that is keeping deprecated nodes for a long time inside the engine).

There is not too much nodes to be changed in this way.  Replacing them manually should not take too much work.

blender/arm/logicnode/depreciated/LN_get_mouse_lock.py
```diff
@@ -0,0 +1,26 @@
+import bpy
+from bpy.props import *
+from bpy.types import Node, NodeSocket
+from arm.logicnode.arm_nodes import *
+
+class GetMouseLockNode(ArmLogicTreeNode):
+    bl_idname = 'LNGetMouseLockNode'
+    bl_label = 'Get Mouse Lock (Depreciated)'
```
MoritzBrueckner, line 8: Just a small reminder that those typos still need to be fixed :)

niacdoial: oh, sorry. It's fixed now.

luboslenco:
One more folder to go, `blender/arm/logicnode/depreciated`. 👼

niacdoial:
https://64.media.tumblr.com/30dd3c802c8bdfbe04cd28c7112b42c1/tumblr_nzbi2lywbs1rn8gxao3_1280.jpg
https://64.media.tumblr.com/cadf6b46bc7c26e6422ff73ea07a4e81/tumblr_nzbi2lywbs1rn8gxao5_1280.jpg
(original post: `https://sirartwork.tumblr.com/post/135142238128/because-ive-lost-control-of-my-life`)

luboslenco:
Thanks a lot!

Issue #1889, opened 2020-09-26
Get Object node not working


biustudio:
this is all i see
![image](https://user-images.githubusercontent.com/46801032/94340732-ce39f380-fffb-11ea-868c-c73d70816c44.png)
code
![image](https://user-images.githubusercontent.com/46801032/94340734-d2661100-fffb-11ea-8b22-149b220ad19e.png)

PR #1892, opened 2020-09-26
Get Object node: allow getting objects from other scenes

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1889.

The Python part of this PR was accidentally commited already when moving the node definitions into subpackages which led to the above bug. Now it is possible to choose the scene from which to take the object in the `Get Object` node. When the scene selection field is empty, the active scene is used.

I've also fixed a bug that resulted in an invalid cast when any other node other than the `Object` node that returned `null` was connected to the `Spawn Object` node.

@luboslenco Is it much work for you to update Krom? It seems to me that it is very unstable since the Kha update, sometimes the game crashes instantly, sometimes after a while or after spawning an object (happened alot during testing this node but also with other nodes, without any error message). Also compiling to HL/C throws a lot of linker errors suddenly (related to some kore compute shader functions), I don't know where that comes from. All submodules are checked out to the commit they are supposed to in armsdk.

### DIFF
--- a/Sources/armory/logicnode/GetObjectNode.hx
+++ b/Sources/armory/logicnode/GetObjectNode.hx
@@ -1,6 +1,12 @@
 package armory.logicnode;
 
+import iron.data.SceneFormat;
+import iron.object.Object;
+
 class GetObjectNode extends LogicNode {
+
+	/** Scene from which to take the object **/
+	public var property0: Null<String>;
 
 	public function new(tree: LogicTree) {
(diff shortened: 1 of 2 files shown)


luboslenco:
Krom binaries updated now.

MoritzBrueckner:
Thanks a lot :) Unfortunately Krom still crashes a lot, I'll try to get more information and eventually open an issue.

Issue #1820, opened 2020-08-28
Open .hx/.py files trough the context menu


knowledgenude:
This opens .hx/.py files of the selected node with the user default text editor.

![nodecontextmenu](https://user-images.githubusercontent.com/63247726/91592305-b86ddb80-e934-11ea-982a-9d3baed2b44c.jpg)

Currently i don't know a way to open those files without searching for them  trough directories, then this can speed up node testing and be more user friendly :)

I hope Blender allow this

N8n5h:
This should do the trick: https://github.com/N8n5h/armory/tree/nodes-context-menu
I don't have a lot of time to properly test it, so feel free to do so and tell me if it works to open a PR.

I used an existing operator to implement it, but I changed it because it opened the git version of the file, which may not be the same you have in your local sdk if you don't update it frequently. Also I purposely choose to keep it as a browser to avoid the user accidentally modifying the code and breaking their SDK (and creating sub-sequent issues about X not working), but I feel that maybe it could be changed to something else, maybe like a mini-web service to display the file as doc? (like old software used to do with their html documentation), or maybe use the version of the sdk to load the correct version from github.

knowledgenude:
No issues running this, just a small difference between opening .py and .hx files, maybe related to my OS:

![screenshot](https://user-images.githubusercontent.com/63247726/93934308-574de380-fcf9-11ea-8597-3768384806f0.jpg)

![screenshot2](https://user-images.githubusercontent.com/63247726/93934318-5a48d400-fcf9-11ea-9e2b-892c070d401d.jpg)

And i confim that this don't work with custom nodes:

![screenshot3](https://user-images.githubusercontent.com/63247726/93934352-67fe5980-fcf9-11ea-8f93-ab5c01e46f07.png)

I can't affirm that the best is to open the files trough Browser, maybe in some cases it is better. The purpose of this is to provide a easy way to fix nodes (but this is not supposed to be frequently) and also to leave users getting examples of haxe script, or creating their custom nodes and testing it, maybe to contribute in the future too. 

I think you should send the PR, this will already be of great help :+1:

N8n5h:
Yeah I noticed that too, with .py files firefox doesn't like to open them and instead prompts you to download them.

[...]

PR #1894, opened 2020-09-27
Automatically register new nodes and categories from libraries

MoritzBrueckner:
See https://github.com/armory3d/armory/issues/1820#issuecomment-699528410 for details. This change makes creating node libraries a bit more easy.


luboslenco:
Thanks!

PR #1895, opened 2020-09-27
Get System Name - to get the name of the system while the application is running

e1e5en-gd:
v1

Section "_Native_".
Output parameters:
- _System Name_ - displays the system name as a string;
- _List systems_ for quick comparison.


luboslenco:
Thanks!

PR #1897, opened 2020-09-28
Update node docstrings

MoritzBrueckner:
According to https://github.com/armory3d/armory_tools/pull/4.

- Copied the old node descriptions from the wiki to Python docstrings and improved them further
- Copied category descriptions from the wiki
- Swapped input socket position for ArrayAddNode (made a lot more sense, also documentation-wise)
- Fixed a category subsection

@ all: If you have some spare time and some in-depth knowledge about some nodes, feel free to improve the docstrings and add more category descriptions (see `arm/logicnode/__init__.py`)

### DIFF
--- a/Sources/armory/logicnode/ArrayAddNode.hx
+++ b/Sources/armory/logicnode/ArrayAddNode.hx
@@ -13,7 +13,7 @@ class ArrayAddNode extends LogicNode {
 		if (ar == null) return;
 
 		// "Modify Original" == `false` -> Copy the input array
-		if (!inputs[3].get()) {
+		if (!inputs[2].get()) {
 			ar = ar.copy();
 		}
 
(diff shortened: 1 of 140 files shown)


knowledgenude:
I also don't like the way i put these descriptions. I am wondering to put more vague descriptions to reduce the words variation and simplify the understanding. For example:

`"Apply inverse kinematics in an object bone."` -> `"Apply inverse kinematics in bones."`
`"Pause an action" `-> `"Pause actions."`

You think this way is better? I will follow the same style for all nodes, except for the ones that requires more details, like arrays or switch.

Other thing i want to know is if i need to  follow Blender shader editor style, because the editor always looks outdated. I don't think it is a good idea to follow it. There is a typo in "MixRGB" node that have years and all nodes don't have descriptions.

I would prefer a design like this: `Separate XYZ` -> `Split Vector` and a description like: `Separate vectors into XYZ values.`

MoritzBrueckner:
@knowledgenude I have no opinion on that. Using plural sounds better to me but this also might make the descriptions a bit more imprecise.

> I would prefer a design like this: Separate XYZ -> Split Vector and a description like: Separate vectors into XYZ values.

The description sounds good but `Separate XYZ` is much more verbose than `Split Vector` (split which vector (2D/3D/4D?) into what?). Also when following Blender conventions (even if they are ugly) it's easier for most people to adapt to Armory's nodes.

PR #1905, opened 2020-10-01
Logic nodes cleanups

knowledgenude:
* Optimize some nodes
* Increase readability a bit

### DIFF
--- a/Sources/armory/logicnode/BloomGetNode.hx
+++ b/Sources/armory/logicnode/BloomGetNode.hx
@@ -7,14 +7,11 @@ class BloomGetNode extends LogicNode {
 	}
 
 	override function get(from:Int):Dynamic {
-        if(from == 0) {
-            return armory.renderpath.Postprocess.bloom_uniforms[0];
-        } else if (from == 1) {
-            return armory.renderpath.Postprocess.bloom_uniforms[1];
-        }  else if (from == 2) {
-            return armory.renderpath.Postprocess.bloom_uniforms[2];
-        } else {
-            return 0.0;
-        }
-    }
+		return switch (from) {
+			case 0: armory.renderpath.Postprocess.bloom_uniforms[0];
+			case 1: armory.renderpath.Postprocess.bloom_uniforms[1];
+			case 2: armory.renderpath.Postprocess.bloom_uniforms[2];
+			default: null;
+		}
+	}
 }
--- a/Sources/armory/logicnode/BloomGetNode.hx
+++ b/Sources/armory/logicnode/BloomGetNode.hx
@@ -11,7 +11,7 @@ class BloomGetNode extends LogicNode {
 			case 0: armory.renderpath.Postprocess.bloom_uniforms[0];
 			case 1: armory.renderpath.Postprocess.bloom_uniforms[1];
 			case 2: armory.renderpath.Postprocess.bloom_uniforms[2];
-			default: null;
+			default: 0.0;
 		}
 	}
 }
--- a/Sources/armory/logicnode/CameraGetNode.hx
+++ b/Sources/armory/logicnode/CameraGetNode.hx
@@ -17,7 +17,7 @@ class CameraGetNode extends LogicNode {
 			case 6: armory.renderpath.Postprocess.camera_uniforms[6];
 			case 7: armory.renderpath.Postprocess.camera_uniforms[7];
 			case 8: armory.renderpath.Postprocess.camera_uniforms[8];
-			default: null;
-        	}
+			default: 0.0;
+		}
 	}
 }
--- a/Sources/armory/logicnode/CameraGetNode.hx
+++ b/Sources/armory/logicnode/CameraGetNode.hx
@@ -7,26 +7,17 @@ class CameraGetNode extends LogicNode {
 	}
 
 	override function get(from:Int):Dynamic {
-        if(from == 0) {
-            return armory.renderpath.Postprocess.camera_uniforms[0];
-        } else if (from == 1) {
-            return armory.renderpath.Postprocess.camera_uniforms[1];
-        }  else if (from == 2) {
-            return armory.renderpath.Postprocess.camera_uniforms[2];
-        }  else if (from == 3) {
-            return armory.renderpath.Postprocess.camera_uniforms[3];
-        }  else if (from == 4) {
-            return armory.renderpath.Postprocess.camera_uniforms[4];
-        }  else if (from == 5) {
-            return armory.renderpath.Postprocess.camera_uniforms[5];
-        }  else if (from == 6) {
-            return armory.renderpath.Postprocess.camera_uniforms[6];
-        }  else if (from == 7) {
-            return armory.renderpath.Postprocess.camera_uniforms[7];
-        }  else if (from == 8) {
-            return armory.renderpath.Postprocess.camera_uniforms[8];
-        } else {
-            return 0.0;
-        }
-    }
+		return switch (from) {
+			case 0: armory.renderpath.Postprocess.camera_uniforms[0];
+			case 1: armory.renderpath.Postprocess.camera_uniforms[1];
+			case 2: armory.renderpath.Postprocess.camera_uniforms[2];
+			case 3: armory.renderpath.Postprocess.camera_uniforms[3];
+			case 4: armory.renderpath.Postprocess.camera_uniforms[4];
+			case 5: armory.renderpath.Postprocess.camera_uniforms[5];
+			case 6: armory.renderpath.Postprocess.camera_uniforms[6];
+			case 7: armory.renderpath.Postprocess.camera_uniforms[7];
+			case 8: armory.renderpath.Postprocess.camera_uniforms[8];
+			default: null;
+        	}
+	}
 }
--- a/Sources/armory/logicnode/ColorgradingGetHighlightNode.hx
+++ b/Sources/armory/logicnode/ColorgradingGetHighlightNode.hx
@@ -6,24 +6,16 @@ class ColorgradingGetHighlightNode extends LogicNode {
 		super(tree);
 	}
 
-    override function get(from:Int):Dynamic {
-        if(from == 0) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[0][2];
-        } else if (from == 1) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[0];
-        }  else if (from == 2) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[1];
-        }  else if (from == 3) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[2];
-        }  else if (from == 4) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[3];
-        }  else if (from == 5) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[4];
-        }  else if (from == 6) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[5];
-        } else {
-            return 0.0;
-        }
-    }
+	override function get(from:Int):Dynamic {
+		return switch (from) {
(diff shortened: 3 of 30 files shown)


Sources/armory/logicnode/BloomGetNode.hx
```diff
@@ -7,14 +7,11 @@ class BloomGetNode extends LogicNode {
...
-            return armory.renderpath.Postprocess.bloom_uniforms[2];
-        } else {
-            return 0.0;
-        }
-    }
+		return switch (from) {
+			case 0: armory.renderpath.Postprocess.bloom_uniforms[0];
+			case 1: armory.renderpath.Postprocess.bloom_uniforms[1];
+			case 2: armory.renderpath.Postprocess.bloom_uniforms[2];
+			default: null;
```
MoritzBrueckner, line 14: `null` and `0.0` are different values, I'm not sure which is better here. But actually, this is rather a case for an exception because this line is only triggered if the node has more outputs socket than implemented in the .hx source. But that's something for a future PR.

Sources/armory/logicnode/CameraGetNode.hx
```diff
@@ -7,26 +7,17 @@ class CameraGetNode extends LogicNode {
...
+			case 1: armory.renderpath.Postprocess.camera_uniforms[1];
+			case 2: armory.renderpath.Postprocess.camera_uniforms[2];
+			case 3: armory.renderpath.Postprocess.camera_uniforms[3];
+			case 4: armory.renderpath.Postprocess.camera_uniforms[4];
+			case 5: armory.renderpath.Postprocess.camera_uniforms[5];
+			case 6: armory.renderpath.Postprocess.camera_uniforms[6];
+			case 7: armory.renderpath.Postprocess.camera_uniforms[7];
+			case 8: armory.renderpath.Postprocess.camera_uniforms[8];
+			default: null;
+        	}
```
MoritzBrueckner, line 21: Tabs and spaces shenanigans here^^

Sources/armory/logicnode/ColorgradingGetHighlightNode.hx
```diff
@@ -6,24 +6,16 @@ class ColorgradingGetHighlightNode extends LogicNode {
 		super(tree);
 	}
 
-    override function get(from:Int):Dynamic {
-        if(from == 0) {
-            return armory.renderpath.Postprocess.colorgrading_highlight_uniforms[0][2];
```
MoritzBrueckner, line 11: This was not a typo here (`colorgrading_highlight_uniforms` is a 2-dimensional array). Also this node has 6 outputs as before, not 5 as it is in this PR.

Sources/armory/logicnode/ColorgradingGetShadowNode.hx
```diff
@@ -6,24 +6,15 @@ class ColorgradingGetShadowNode extends LogicNode {
 		super(tree);
 	}
 
-    override function get(from:Int):Dynamic {
-        if(from == 0) {
-            return armory.renderpath.Postprocess.colorgrading_shadow_uniforms[0][1];
```
MoritzBrueckner, line 11: The same goes for this, this is not a typo (check the socket types, the first one is a float output and not a vector!) and also an output got merged like above.

Sources/armory/logicnode/GetCameraFovNode.hx
```diff
@@ -11,7 +11,7 @@ class GetCameraFovNode extends LogicNode {
 	override function get(from: Int): Dynamic {
 		var camera: CameraObject = inputs[0].get();
 
-		if (camera == null) return 0.0;
+		if (camera == null) return null;
```
MoritzBrueckner, line 14: I'm not sure about this. `null` values are often forgotten and lead to errors, also they're sometimes problematic in Haxe when targeting static platforms (read about nullability in Haxe). If someone get's a value of `0.0` I think it's clear that the camera couldn't be found.

If you leave it on `null`, this needs to be documented.

Sources/armory/logicnode/GetCursorStateNode.hx
```diff
@@ -12,13 +12,13 @@ class GetCursorStateNode extends LogicNode {
...
+			if (mouse.hidden == true && mouse.locked == true) return true;
+			return false;
 
-		case 1: mouse.hidden;
-		case 2: mouse.locked;
-    default: null;
-    }
+			case 1: mouse.hidden;
+			case 2: mouse.locked;
+    			default: null;
```
MoritzBrueckner, line 21: Again some problem with tabs and spaces (sorry for being so nitpicky here)

Sources/armory/logicnode/GetVelocityNode.hx
```diff
@@ -16,11 +16,12 @@ class GetVelocityNode extends LogicNode {
...
-		if (rb != null) {
-			if (from == 0) return rb.getLinearVelocity();
-			else if (from == 1) return rb.getAngularVelocity();
+		return switch (from) {
+			case 0: rb.getLinearVelocity();
+			case 1: rb.getAngularVelocity();
+			default: null;
 		}
 #end
-		return null;
```
MoritzBrueckner, line 24: If the `arm_physics` flag is not set this function won't return anything anymore.

Sources/armory/logicnode/LenstextureGetNode.hx
```diff
@@ -7,18 +7,13 @@ class LenstextureGetNode extends LogicNode {
...
-        }  else if (from == 3) {
-            return armory.renderpath.Postprocess.lenstexture_uniforms[3];
-        }  else if (from == 4) {
-            return armory.renderpath.Postprocess.lenstexture_uniforms[4];
-        } else {
-            return 0.0;
-        }
-    }
+		return switch (from) {
+            		case 0: armory.renderpath.Postprocess.lenstexture_uniforms[0];
```
MoritzBrueckner, line 11: Again some tabs and spaces shenanigans

MoritzBrueckner (changes requested):

Sources/armory/logicnode/ColorgradingGetShadowNode.hx
MoritzBrueckner: Of course this also goes for `colorgrading_global_uniforms`, I forgot that.

Sources/armory/logicnode/BloomGetNode.hx
knowledgenude: I will leave these values as 0.0 for now, because i don't know what effect this will cause and i can't test now. Thanks!

Sources/armory/logicnode/CameraGetNode.hx
knowledgenude: By tabs and spaces shenanigans you mean a mix of tab and space at the same line or other thing? I can't found any reference about the meaning of this :/

If you can send a minimal example (with the wrong and the right) it would help

Sources/armory/logicnode/ColorgradingGetHighlightNode.hx
knowledgenude: Okay i will "unfix" this. Thanks!

knowledgenude: About this line: `armory.renderpath.Postprocess.colorgrading_highlight_uniforms[0][2];`

If i declare a variable like: `var colorgrading: String = armory.renderpath.Postprocess.colorgrading_highlight_uniforms`

It will work as this: colorgrading[0][2]

I think this is better for readability, but i don't have sure about the performance. Can you tell me if this works and if this takes some performance?

Sources/armory/logicnode/GetCursorStateNode.hx
knowledgenude: No problem about nitpicky, i also prefer the things the better as possible

Sources/armory/logicnode/CameraGetNode.hx
MoritzBrueckner: That's exactly what I mean. Here on GitHub you can see that the closing `}` is indented too much, and if you select the whitespace before that you see that there is a mix of spaces and tabs. Iron uses tabs and the Python part of Armory uses spaces.

Sources/armory/logicnode/ColorgradingGetHighlightNode.hx
MoritzBrueckner: Why `String`? Have a look at where it is defined:

https://github.com/armory3d/armory/blob/da1cf33930ce9a8b1865d35c128fe4842bef2933/Sources/armory/renderpath/Postprocess.hx#L36-L42


It's an array of arrays (2D), so instead of String you would need to use `Array<Array<Float>>` as the type. I'm not sure about the performance, it is a bit slower on one hand because it's saved in a variable (memory needs to be allocated and it has to be written into that variable), but on the other hand it's faster because the computer does not need to go from `armory` to `renderpath` and so on to read from that variable. I don't know,

knowledgenude: Hm, really interesting. 

I tough that if i put: `var ballName = ball.name.12.3.1.3.13`, the script would be interpreted like: he put `ballName` but i know that `ballName` is just a shorter name for `ball.name.123.1.3.13`. This is why i would put the variable as string.

Just a newbie :) Thanks for the explanation!

Sources/armory/logicnode/GetVelocityNode.hx
knowledgenude: Just a doubt here: the `default: null;` line of the switch is the same as if i put `return null;` without switch? This is why i deleted those lines.

In the case of `arm_physics`, i would not remove the `return`, but and for other cases? If i found some thing like that i send here to clarify this

Sources/armory/logicnode/ColorgradingGetHighlightNode.hx
MoritzBrueckner: What you describe does exist and it is called "inline variables". Haxe also has them if you explicitly declare them as `inline var`, but you can't use them inside of functions. But as I said, chances are good that Haxe automatically inlines this variable. If you're unsure, let it as it was without a variable.

Sources/armory/logicnode/GetVelocityNode.hx
MoritzBrueckner: In would be the same if there wasn't the `#end` in between. The removed `return null` was executed independently of the `arm_physics` flag (https://haxe.org/manual/lf-condition-compilation.html), now you put it inside the block that's only included in the build if that flag is set. What you can do is to replace `#end` by

```haxe
#else
	return null;
#end
```

Sources/armory/logicnode/ColorgradingGetShadowNode.hx
knowledgenude: > The same goes for this, this is not a typo (check the socket types, the first one is a float output and not a vector!) and also an output got merged like above.

I didn't understand well. I need to correct the socket type in the .py of this nodes because they are vectors?

I am also thinking about rename `ColorGrading Get Shadow` Node by `Get CG Shadow`, you think is  a good thing? Or there is some better name design for this nodes?

MoritzBrueckner: > I didn't understand well. I need to correct the socket type in the .py of this nodes because they are vectors?

No, sorry, I was just referring to the node so that you see what socket output type to expect. You changed it (because you thought this was a typo) to a returned vector instead of a float value.

> I am also thinking about rename ColorGrading Get Shadow Node by Get CG Shadow, you think is a good thing? Or there is some better name design for this nodes?

I wouldn't to this, in this context I wouldn't now what CG means. I will soon investigate if we can set minimum widths for certain nodes, maybe this will resolve some of those naming problems as well (as the name is not cut off anymore)

knowledgenude: > I will soon investigate if we can set minimum widths for certain nodes

This would be amazing!

Issue #1707, opened 2020-05-18
Collision margin is inverted


knowledgenude:
Collision margin is from object to outside, but in Armory it is to inside. Both can be useful. A way to set negative and positive values would be great.

About margin: [video](https://youtu.be/BGAwRKPlpCw)

Lykdraft:
Someone knows where to find this behaviour in the source code? I can't imagine this 
should be too difficult to fix. I just don't know where to look. But it is extremely annoying.

knowledgenude:
You can fix it in `Sources/armory/trait/physics/bullet`

There is a `setMargin` function: https://armory3d.org/api/bullet/CollisionShape.html

I don't think it should be difficult to fix (if the problem is just an inverted value). I will also take a look other time.

luboslenco:
Closed via https://github.com/armory3d/armory/pull/1907.

PR #1907, opened 2020-10-02
Bullet: Collision margin was inverted.

Lykdraft:
This should work now as expected.

Original issue:

https://github.com/armory3d/armory/issues/1707


luboslenco:
Thanks!

PR #1913, opened 2020-10-05
Improve LN descriptions

knowledgenude:
* Improve LN descriptions
* Add slot input to Get Object Material (don't had tested it yet)
* Renamed Pick Rigid Body to Pick RB to stay in accord with the other RB LNs (in the future if there is a way to set the LN size, they can be renamed back to Rigid Body).

Question: There is two nodes to set the object material, that is: `Set Object Material` and `Set Object Material Slot`. Is not better to deprecate `Set Object Material` and put the other one in its place?


luboslenco:
Thanks! Yes it looks like `Set Object Material Slot` can be used instead. `Set Object Material` sets all the slots at once though.

PR #1921, opened 2020-10-09
Merged pause/resume nodes

knowledgenude:
I think it is more user friendly and cleaner now. I had misunderstood the speaker functionality, now it is not affected.

* `Pause Trait` and `Resume Trait` merged to `Set Trait Paused`.

* `Pause Tilesheet` and `Resume Tilesheet` merged to `Set Tilesheet Paused`.

* `Pause Action` and `Resume Action` merged to `Set Action Paused`.

* Deprecated `Play Action` and `Set Object Material` by `Play Action From` and `Set Object Material Slot` respectively.

* Renamed `Action State` to `Get Action State`.

### DIFF
--- a/Sources/armory/logicnode/SetActionPausedNode.hx
+++ b/Sources/armory/logicnode/SetActionPausedNode.hx
@@ -1,0 +1,25 @@
+package armory.logicnode;
+
+import iron.object.Object;
+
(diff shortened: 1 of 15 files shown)


luboslenco:
Thanks!

PR #1922, opened 2020-10-10
Add Debug Console settings

e1e5en-gd:
1. Adding the ability to customize display, scale (size), shortcuts to the _DebugConsole_ class.
2. Adding a function to _utils.py_ to get Debug Console settings from _Render: Armory_.
3. Added Debug Console settings to the _Armory Project_ interface:
debug_console_armory_project
    - Enable (default value - False);
    - Position (Left, Center, Right, default value - Right);
    - Scale Console (from 0.3 to 10);
    - Visible.
4. Added transfer of Debug Console settings to _exporter.py_.
5. Added logical nodes to control DebugConsole while the application is running.
v2


luboslenco:
Thanks!

Issue #664, opened 2018-06-08
World Material Nodes - Math, Colors, Vectors


adriansnetlis:
In world material nodes the only nodes that work are:

- World Output
- Background Shader
- All texture nodes

However, once you add in nodes like color mix, math, vector math, normal, geometry etc., the background turns grey. Apparently support for those nodes is not coded in yet. Would be cool to be able to do more stuff in world shaders such as creating custom sky shaders for more control. Also, a light data node that would let you get incoming light vector, color, and strength of a light source would be amazing.

PR #1924, opened 2020-10-10
World nodes support and nodes parsing refactor

MoritzBrueckner:
Implements https://github.com/armory3d/armory/issues/664.

**Changes**:
- Added support for world nodes
- Moved nodes parsing into a small subpackage to make things more clearly
- Show warnings for unsupported node types
- Added a new class `ParserState` to hold information that was previously stored as global variables in cycles.py (this makes it easier to reset the state after parsing and leads to less chaos and confusion)
- Added a few docstrings and type hints (also there are now some type aliases `floatstr`, `vec3str` etc. to show that a function returns a string that has a certain type when interpreted as glsl code, maybe it's noise but I thought it might be helpful)
- Implemented two more Vector Math node operations: distance and length
- Optimized the usage of the `render.film_transparent` setting, now the node tree isn't parsed when this option is active
- Better handling of cases where there are no world or no surface nodes
- Node panels in the sidebar are now only shown when applicable to the current node tree type

**Example**

Example screenshot
Nodes system

**Still open issues to fix in the future:**
- Currently only static, view-space background image textures are supported for worlds, just as before. The current material nodes texture implementation still relies on materials.
- Because multiple sky/envmap/image texture nodes are possible now, we can't just use an already existing image to generate radiance/irradiance textures. So currently it's first come, first served until there is a better solution (@luboslenco when Kha/Kinc gets a headless option, do you think it would make sense to write a small program that renders the world to a irradiance/radiance texture while exporting? Or maybe even when changing the node tree to speed up export?)
- Because of the above issue with radiance, there is no radiance/irradiance support at all when not using a sky texture or envmap (see screenshot above)
- It's still not possible to connect group or Mix Shader nodes to the world output node directly, only Background and Emission shader nodes are supported
- The sun direction/turbidity/ground albedo values of sky textures are static/per-world, so it's undefined  from which nodes the values will be exported
- When there are problems writing to a shader file (it is opened somewhere for example), there are issues on consecutive runs where the shader is just empty (probably some module variable somewhere that isn't reset)
- It isn't possible yet to parse/follow the background/emission node strength value as Armory exports this as a static, uniform value (`world.arm_envtex_strength`).

**Ideas for the future**:
- Clouds could be ported into a custom node so that there are much more possibilities with it
- The Object Info node's color output doesn't work yet for materials, Iron maybe should implement object color (easy to do)
- We should test if it is possible to get node sockets via name and not via index without loosing too much performance to make node parsing less error-prone due to Blender changes (some nodes parsed wrong sockets because they weren't updated in a long time)
- Also, there are currently many cases when parsing nodes where there are big `if/elif/elif/elif/...` chains without a default case that could lead to errors with unsupported/changed nodes. We should probably print a warning instead.


luboslenco:
Yay!

> when Kha/Kinc gets a headless option, do you think it would make sense to write a small program that renders the world to a irradiance/radiance texture while exporting? Or maybe even when changing the node tree to speed up export?

It should be doable via Cycles as well: https://docs.blender.org/manual/en/latest/render/cycles/object_settings/cameras.html#equirectangular

Once this map is rendered the remaining steps are the same as using the environment texture now.

MoritzBrueckner:
> It should be doable via Cycles as well

The problem I see with this is that it would probably be the best to silently run another Blender instance in the background to render to that texture (in order to not show the render on screen) which could be slow, also the worlds first need to be linked or the entire project needs to be loaded again.

Another potential issue is that the feature set of Cycles and Armory renders are different but that might be acceptable as the goal is too keep both as close together as possible. An exception to that are clouds.

On the other hand, equirectangular rendering might be complicated to implement in a custom application.

I will do some testing, thank you!

Naxela:
This looks great!

As for the irradince/radiance you might want to check out the [code for environment probes in The Lightmapper](https://github.com/Naxela/The_Lightmapper/blob/a15a1059cf3e2a857cbe8563b95134e4a2d23e8a/addon/operators/tlm.py#L393), essentially it does much of the same as [Armory does when creating radiance and irradiance](https://github.com/armory3d/armory/blob/master/blender/arm/write_probes.py), only difference is that it renders out a cubemap first (thus available for both Cycles and Eevee, the latter being good if speed is an issue). Afterwards, the cubemap is stitched in CMFT, and radiance/irradiance calculated as usual - I'm currently working on using it for baked probes in Armory3D similar to Eevee, but I guess the process could be the same for the world backgrounds if a good way could figured out, whether being a background blender instance, or an empty temporary scene at export-time

billede

MoritzBrueckner:
Thanks!

I did some tests yesterday and the background rendering actually works quite good - the only big issue is that there is no way to access armory functions like `arm.utils.safestr` for file names because the background script is executed before Armory registered its SDK (that's delayed in order to access the SDK path preference).

It looks like you can also render in a running Blender session in the background without opening the render screen? Or does this happen with the Lightmapper? If it is possible, it would make things easier and faster indeed to just use a temporary empty scene.

Using Eevee for rendering might not be a good choice in this case as Eevee (AFAIK) does not support sky textures, at least in B2.8. I still have to figure out the correct camera rotation for the equirectangular projection but if it doesn't work I will reuse your cubemap code :)

The other big problem I'm having currently is that something is wrong with the radiance/irradiance colors, they seem inverted as if they would be subtracted from white? I have a small test scene with just the world normals as background colors.

Blender looks like this:


And Armory like this (the bottom face that should be blue is actually white, I changed the camera perspective so that you can see the pink color above which should not be there). Actually all the other faces are white.


The exported radiance textures look correct: 
env_World_radiance_3

All I did change was creating a custom render of the world (which looks like the texture above) and then feeding that into `write_probes.write_probes()`.

Naxela:
I see how that can be a problem, I'm not sure if there's a way to defer the process until after the SDK has been loaded somehow (maybe add a blender timer that runs until a bool property has been set by the SDK?)

Yes, the background rendering is done as a subprocess, and a timed bpy event checks in the background for a process file and continues the process once complete without locking up the Blender instance. It can be done with threading too, but threads in Blender is a pain to work with.

I agree that until Eevee support sky textures, it might not be the best option - The downside of Cycles is that it's sometimes noisy unless using a denoiser (could use the same function for the Armory lightmapper) or enough samples. - As for the camera rotation, it was a puzzle for me too, since CMFT seems to shift some orientations, but the puzzle will eventually get solved

I'm not entirely sure as for the irradiance results you get, but they're coefficients for spherical harmonics (which again is just a unintensive way of applying [spherical image based lighting values to objects](https://chetanjags.wordpress.com/2016/12/17/using-spherical-harmonics-for-ibl-probes/)). I haven't tried using the write probes function, what I usually do is just render out the radiance map and SH coefficients (in the environment probe thingy), and use Haxe to load it (just an example, it was done the same way that Armorpaint does it, but I now can't find the file that does it):

```haxe
package arm;

import kha.Image;
import kha.Blob;
import kha.graphics4.TextureFormat;
import kha.arrays.Float32Array;
import iron.data.Data;
import iron.Scene;

using StringTools; 

class EnvironmentShift extends iron.Trait {
	public function new() {
		super();

		notifyOnInit(function() {

                        //First set the irradiance from *.c file that CMFT makes, contains the SH coefficients
			Data.getBlob("Environment.c", function(blob: Blob) {
				var lines = blob.toString().split("\n");
				var band0 = lines[5];
				var band1 = lines[6];
				var band2 = lines[7];
				band0 = band0.substring(band0.indexOf("{"), band0.length);
				band1 = band1.substring(band1.indexOf("{"), band1.length);
				band2 = band2.substring(band2.indexOf("{"), band2.length);
				var band = band0 + band1 + band2;
				band = band.replace("{", "");
				band = band.replace("}", "");
				var ar = band.split(",");
				var buf = new Float32Array(28);
				for (i in 0...27) buf[i] = Std.parseFloat(ar[i]);
				Scene.active.world.probe.irradiance = buf;
			});

                        //Set the visual background to the HDR map
			Data.getImage("Background.hdr", function(image1:Image){

				Scene.active.world.envmap = image1;

			});

                        //Set the radiance for reflections, etc. - Perhaps best with mipmaps, but shown for the simplest way
			Data.getImage("Background.hdr", function(image2:Image){
				Scene.active.world.probe.radiance = image2;
			});

		});
	}
}

```

The irradiance values are used in the deferred_light shader, using [shIrradiance function](https://github.com/armory3d/armory/blob/master/Shaders/std/shirr.glsl)

Hope it helps a little

MoritzBrueckner:
Thanks again, I will try to learn a bit about spherical harmonics then :) Do I understand it correctly that there are (horizontal/vertical/crossed?) bands around a unit sphere that set the indirect color value for that "direction" the bands represent?

>  I'm not sure if there's a way to defer the process until after the SDK has been loaded somehow

That's close to impossible. Registering a timer or a handler or similar would allow using the arm module in theory but the script continues running and if it's done Blender will quit, discarding the registered timer. I just copied the required functions for now. A workaround in the future could be to write the SDK path to a config file that gets loaded directly when the addon registers.

> The downside of Cycles is that it's sometimes noisy unless using a denoiser (could use the same function for the Armory lightmapper) or enough samples.

We could allow to set the sample size in the Armory preferences or on a per-project base. In my tests even a low sample size of 16 was pretty good for just a world background.

The code you've pasted is pretty close to what Armory does, it copies the irradiance values in a json/arm array with the same logic in Python and then uses that array in Haxe. I'm not even sure if that part is wrong, it seemed to work before so the mistake might rather be with how I generate the environment texture (you can see that those CMY colors are also on the environment texture). It's difficult to see if the texture is correct if the (irr)radiance is wrong and vice versa^^

@luboslenco Maybe you have an idea why those colors seem inverted (CMY instead of RGB)?

MoritzBrueckner:
So I did some more tests and have a few more questions/conclusions:

- The hosek sky in Armory is in general a bit brighter than Armory's sky. This makes a huge difference if the sky is made darker by a certain factor for example. @luboslenco any idea why?

- Passing an input gamma numerator for the spherical harmonics generation seems to be the correct way for non-HDR surroundings (but it's just a guess) even if it is a bit darker (cycles also has that dark irradiance even if it is a bit brighter)

- In `write_probes.write_color_irradiance` the colors are multiplied with 1.13, the comment says "Adjust to cycles". So we should probably do that in `write_probes.parse_band_floats` too? Is that factor estimated or actually taken from some cycles code?

- It's currently a bit of a mess what needs to be gamma corrected and not. I don't know much about that topic, but do I understand it correctly that everything inside the world shaders apart from HDRs and hosek skies (also HDR?) needs to be gamma corrected?

- Also is there a reason that the radiance files are converted to jpgs even if we have them in hdr? Is that because of performance reasons or doesn't Iron support that or something else?

Sorry for all those questions but I need to understand it a bit more in order to hopefully take the right decisions. Thank you!

PR #1929, opened 2020-10-13
Get System Language - LN

e1e5en-gd:
Get System Language

get_system_language

Category: _Native_
To get the current system language according to the IS0 639 standard.
The _Krom_ platform always returns “_en_”.

### DIFF
--- a/blender/arm/logicnode/native/LN_get_system_language.py
+++ b/blender/arm/logicnode/native/LN_get_system_language.py
@@ -1,0 +1,13 @@
+from arm.logicnode.arm_nodes import *
+
+class GetSystemLanguage(ArmLogicTreeNode):
+    """Returns the name of the current system."""
+    bl_idname = 'LNGetSystemLanguage'
+    bl_label = 'Get System Language'
+    arm_version = 1
+
+    def init(self, context):
+        super(GetSystemLanguage, self).init(context)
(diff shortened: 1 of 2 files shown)


blender/arm/logicnode/native/LN_get_system_language.py
```diff
@@ -0,0 +1,13 @@
+from arm.logicnode.arm_nodes import *
+
+class GetSystemLanguage(ArmLogicTreeNode):
+    """Returns the name of the current system."""
```
MoritzBrueckner, line 4: I think the docstring is left over from another node?

e1e5en-gd: Yes. Sorry

luboslenco:
> The Krom platform always returns “en”.

Which OS are you on? It should work everywhere apart from Linux - that one is still missing here:
https://github.com/Kode/Kinc/blob/master/Backends/System/Linux/Sources/Kore/System.cpp#L846

e1e5en-gd:
I have Windows 10 and language ru.
Krom gave out en, so I looked at its sources (https://github.com/Kode/Kha/blob/5a489c1305b219e2ad1e9d089c32d8a53a21d4f7/Backends/Krom/kha/SystemImpl.hx#L223). I did not look into the source code of Linux.

PR #1931, opened 2020-10-14
Get Tilesheet State node

knowledgenude:
This also fixes previous commits that i did not tested.

Users must recreate the `Get Action State` node because i changed it from "Is Playing" to "Is Paused" to match the `Set Action State` node.

Requires: https://github.com/armory3d/iron/pull/106

Question: do `Action` node works also for tilesheets or there is a need for a `Tilesheet` node too?


luboslenco:
Don't think the `Action` node applies to tilesheets.

knowledgenude:
Ok, i've changed the socket names as `Play Tilesheet` works with the tilesheet name and not with actions

PR #1942, opened 2020-10-19
Two new logic nodes for the "Random" category

MoritzBrueckner:
- **Random Choice**:
  Choose a random value from a given array.
  RandomChoice Node

- **Random Output**:
  Activate a random output when the input is activated.
  RandomOutput Node


luboslenco:
Cool!

PR #1948, opened 2020-10-22
Fully implement the VectorMath node

MoritzBrueckner:
All operators are now supported and adapted to Blender's functionality. The only exception so far is the modulo operator, `mod` in GLSL works a bit different than `fmod` used by Blender/OSL when using negative values.

I also removed the `AVERAGE` operator because it was removed from Blender (misleading name, it was averaging the direction via add + normalize and wasn't just (A + B) / 2).


luboslenco:
Awesome!

PR #1950, opened 2020-10-22
New node Vector From Boolean

knowledgenude:
This can be used as a kind of input map.

Example: [vectorfromboolean.zip](https://github.com/armory3d/armory/files/5425055/vectorfromboolean.zip)

With this node, this logic setup is sufficient to make a character walk in any direction:

vectorfromboolean


luboslenco:
Thanks!

PR #1951, opened 2020-10-23
More cycles nodes + improvements

MoritzBrueckner:
- Constant factors for the MixRGB node are now inlined (actually I think we could also do this for non-constant factors, there are stored in a variable anyways I think?)
- Fixed some socket indices and replaced the indices for some nodes with string identifiers (I'm not sure about the speed loss/how Blender handles this internally)
- Fixed `Normal` output in the Texture Coordinate node for world shaders
- The Light Falloff node now follows Eevee's implementation
- Implemented the Separate HSV node
- Aligned the Light Path node to Eevee's implementation and fixed some socket indices

### DIFF
--- a/blender/arm/material/cycles_nodes/nodes_color.py
+++ b/blender/arm/material/cycles_nodes/nodes_color.py
@@ -1,5 +1,6 @@
 import bpy
 
+import arm.log as log
 import arm.material.cycles as c
 import arm.material.cycles_functions as c_functions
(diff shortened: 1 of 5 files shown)


luboslenco:
Super cool!

> Constant factors for the MixRGB node are now inlined (actually I think we could also do this for non-constant factors, there are stored in a variable anyways I think?)

Not sure if that is the case for all nodes, but if so then yes.

Issue #935, opened 2018-10-07
Bump Node not really working with procedural textures


zaethan:
While trying around with some material Library, I noticed that some of them would fail to even compile with Armory.
Looking for the cause of this issue I found that using the Bump node's height input with procedural textures was one of the problematic situations. (it seems like there was an issue about this node about a year ago, but it seems like it was about image textures: https://github.com/armory3d/armory/issues/322)
Trying out most of the procedural textures, I found out that you can plug the color output of any of the textures into the height input and the game will compile and run. But it does not use the texture at all...
If you try to use the Fac output, it will still compile for most of the nodes (you still don't see any texture in Armory), but fail at compiling for the Noise Texture and the Checker Texture.

This is my example setup I used for testing:
![image](https://user-images.githubusercontent.com/39460066/46585072-27ee2f80-ca6c-11e8-9697-53829304da15.png)

If I use the Color Output of any Texture (Brick in this case) I can see it in Cycles, but it does not appear in Armory
![image](https://user-images.githubusercontent.com/39460066/46585104-94692e80-ca6c-11e8-8708-249b64b7d408.png)
The same happens when using the Fac output.
Shader: [Material_mesh.frag.glsl](https://github.com/armory3d/armory/files/2454172/Material_mesh.frag.glsl.txt)
Looking at line 34/35, one can see that it does create the BrickTexture... but then uses the `n` variable and never touches the texture again. This seems to be caused by the `sample_bump_res` variable in the cycles.py file being empty (It also seems like the distance input is not used...)

If I use use the Fac output of either the Noise or Checker Texture, the compilation fails with this output:
> Compiling shader 1 of 1 (Material_mesh.frag.glsl).
> ERROR: /run/media/yannik/Daten/Blender/Armory3D/Tests/MaterialLibrary/build_SingleMat/compiled/Shaders/Material_mesh.frag.glsl:37: '+' :  wrong operand types: no operation '+' exists that takes a left-hand operand of type ' temp 3-component vector of float' and a right operand of type ' const 2-component vector of float' (or there is no acceptable conversion)
> ERROR: /run/media/yannik/Daten/Blender/Armory3D/Tests/MaterialLibrary/build_SingleMat/compiled/Shaders/Material_mesh.frag.glsl:37: '' : compilation terminated 
> ERROR: 2 compilation errors.  No code generated.
> SPIR-V is not generated for failed compile or link
> Compiling shader 1 of 1 (Material_mesh.frag.glsl) failed:
[...]

PR #1952, opened 2020-10-24
Fix bump node when connected to vector output

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/935.

The output still looks different in Blender, Armory samples the texture around a coordinate (which might be slow because the (procedural) texture has do be recalculated 5 times, I don't know?) and Eevee seems to use partial derivatives to get the normals from the height map which might be a better solution.


luboslenco:
IIRC derivatives were used originally but swapped in attempt to improve quality.

PR #1953, opened 2020-10-24
Vector To Object Orientation node

knowledgenude:
This node is to provide a friendly way to do https://forums.armory3d.org/t/third-peson-character-control-with-logic-nodes/4129

I would implement this in the `Get World Orientation` node but this would be less friendly.

Basically this does what `On Local Axis` checkboxes does to a vector.

Test here: [player.zip](https://github.com/armory3d/armory/files/5434853/player.zip)

Preview: b9b7a13783d287061afac6dd47e4a53d798ba948


MoritzBrueckner:
Nice!

I'm not entirely sure though if `Vector to Object Space` is the correct name here. I looks like you rotate the input vector according to the object's orientation (so you're changing the base vectors from (1,0,0),(0,1,0),(0,0,1) to the object's look/right/up vectors), but in my view the name of the node suggests that the vector is transformed into an object-space vector that points to the same point in the world as the input vector which is now relative to the object (= in object-space, just think about a coordinate system around the object where the object is always at 0,0,0 with no rotation, that's object space).

knowledgenude:
So maybe the better name is  `Object Relative Vector`?

e1e5en-gd:
> Basically this does what `On Local Axis` checkboxes does to a vector.

Perhaps the name should focus on the functionality of the node (`On Local Axis`). For example `Local Vector To Object`.

MoritzBrueckner:
Hmm, that's difficult. What about `Vector To Object Orientation` (or rotation) or `Vector Rotate By Object` (I don't like both versions that much). Or `Vector Apply Orientation`?

knowledgenude:
As Armory tries to fit Blender standards, looking in how blender names this relations, the Local makes sense (and not requires to specify that it is an object). Maybe a user can have the idea in mind like "How do i change this vector to a local way just like blender does?" and blender references it in this way:

Screenshot from 2020-10-24 15-57-47

I don't even know about the work to put normal option but following this principles:

Blender uses `Transform Orientation` - so maybe `Convert Transform Orientation` with a drop box `Global To Local` `Global To Normal` `Global To View` ? The normal should be crazy to do and If the operation is only for the object local itself, we can name it as `Global Vector To Local` or `World Vector To Local`. The camera is very possible and may be useful since the Blender camera axis is different from the other objects. What do you guys think?

Also Blender calls it `Global` instead of `World`, maybe we should change this too in the other nodes or we just stick with `World`?

knowledgenude:
I've renamed it to `Vector To Object Orientation`, in the future we may extend its functionality and rename it if needed. Thanks for the suggestions!

PR #1954, opened 2020-10-24
LN for Traits

e1e5en-gd:
1. Modification of `CanvasScript` and `WasmScript` classes: added a field to save the name of the script/file in the class.
2. Added logical nodes for working with `Traits`. 
Categories: _Trait_.

get_taits_name_v3

- `Get Object Traits` - gets an array of `Traits` for the specified object;
- `Get Trait Name` - get the `Trait` name (the name that appears in the list in `Armory Traits`) and its type (_Canvas_, _Wasm_, _Bundle_, _LogicNode_ or _Haxe_).


luboslenco:
Thanks!

PR #1966, opened 2020-10-28
Add Bounce and Elastic to Mix nodes

knowledgenude:
I don't know if there is some reason for this not available, seems to work fine.


luboslenco:
Thanks! There were added more recently by https://github.com/armory3d/iron/pull/95.

PR #1967, opened 2020-10-28
Improve handling of deprecated logic nodes

MoritzBrueckner:
Linked to https://github.com/armory3d/armory_tools/pull/5.

- Deprecated logic nodes are now also registered, but not in a menu. This makes it possible to access them for reference generation.
- There is now a `@deprecate()` decorator for the node classes to reduce duplicate code:

  ```python
  @deprecate()
  class MyNode(ArmLogicTreeNode):
     ...

  @deprecate('AlternativeNode 1', 'Alternative Node 2')
  class MyNodeWithAlternatives(ArmLogicTreeNode):
     ...

  @deprecate('AlternativeNode 1', 'Alternative Node 2', message='This node was deprecated because of XYZ')
  class MyNodeWithAlternativesAndMessage(ArmLogicTreeNode):
     ...
  ```

  Using this decorator will automatically deprecate the node ('is_obsolete = True'), set the correct icon (`ERROR`), change the label and add documentation according to the passed parameters. This way we can just keep the nodes as they are (no need to change the label, description or package) and everything will happen automatically.

  The generated reference will look like the following (without a custom message that would be appended):
  
  Screenshot of a deprecated node in the reference

  This does *not* replace the auto-replacement feature, this is something that must still be implemented for each node in question.

- Outsourced some common functionality into `utils.py` and `node_utils.py`

### DIFF
--- a/blender/arm/logicnode/arm_nodes.py
+++ b/blender/arm/logicnode/arm_nodes.py
@@ -7,6 +7,8 @@ import bpy.types
 from bpy.props import *
 from nodeitems_utils import NodeItem
 
+import arm.node_utils
+
 # When passed as a category to add_node(), this will use the capitalized
 # name of the package of the node as the category to make renaming
 # categories easier.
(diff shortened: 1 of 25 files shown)


luboslenco:
👍 👍

PR #1974, opened 2020-10-30
Replace None node

knowledgenude:
When the `value 1` is `null`, the `value 2` is returned in its place. Useful to avoid duplicates of `Set` nodes. See it in practice below:

Captura de tela de 2020-10-30 14-45-25


knowledgenude:
@MoritzBrueckner, by your alternatives i would go for `Value Or Default` with `Value In` and `Default` inputs? I would use `Value In` and `Value Out` just to avoid confusion when referencing the sockets.

The question is: if you are a beginner that is receiving (null should be...) you may search for a node that change the null value to another value, so is not more friendly `Null To Default`?

And what about the nodes that replaces the `Null` word by `None` should i rename them?

MoritzBrueckner:
Sounds good. `Default if Null` or `Value if Null` could work as well, maybe those names are even better because they imply a condition.

Regarding `Null`/`None`: I think we should change it to `Null`, as `None` is used by Python and most regular users will probably use more Haxe than Python (even if `null` is lower case in Haxe and not lower case in node titles). Nodes get compiled into Haxe code so we should stick to the Haxe naming conventions as much as possible.

luboslenco:
Nice!

Issue #870, opened 2018-09-10
Improve performance testing


luboslenco:
Create an open set of performance tests and monitor results for each Armory release. Track on all platforms and graphics APIs, detect regressions for reworked features.

tong:
https://github.com/HaxeBenchmarks/ might be an option or starting point.

PR #1977, opened 2020-10-31
Implement optional profiling for export

MoritzBrueckner:
Linked to https://github.com/armory3d/armsdk/pull/16.

Combined with the PR above, you can now choose to run [cProfile](https://docs.python.org/3.7/library/profile.html#module-cProfile) when exporting the scene. This will generate a file called `profile_exporter.prof` in the SDK directory with detailed information about the number of calls and the time taken by the functions called during export. This file can then be opened by tools such as [SnakeViz](https://jiffyclub.github.io/snakeviz/) to get a graphical representation of whats happening during export:

----
**Example use case:** 

snakeviz_exporter

This is the result when compiling the celshade example project, which takes a huge amount of time. About 3/4 of the time are taken away for the export of individual objects (and analyzing the results a bit more leads to the conclusion that [the way bone animations are exported](https://github.com/armory3d/armory/blob/master/blender/arm/exporter.py#L487-L500) is horrendously slow because for each bone the entire scene is set to each frame resulting in a complete redrawing and recalculation of the scene).

This is a very small step for the Python side of things towards https://github.com/armory3d/armory/issues/870. The new `Profile` context manager can be used for other occasions as well and can even be nested, although it will mess a bit with the results. It should always be opt-in due to performance reasons.

I've also added a new util function to get a preference setting or a default value if it doesn't exist, which makes many of the current `get_[something]()` functions obsolete (they are still used though, that's not part of this PR).

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -30,6 +30,7 @@ import arm.material.cycles as cycles
 import arm.material.make as make_material
 import arm.material.mat_batch as mat_batch
 import arm.utils
+import arm.profiler
 
 
 @unique
@@ -127,7 +128,6 @@ class ArmoryExporter:
         self.world_array = []
         self.particle_system_array = {}
 
-
         # `True` if there is at least one spawned camera in the scene
         self.camera_spawned = False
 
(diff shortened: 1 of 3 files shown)


luboslenco:
Thanks!

>  the way bone animations are exported is horrendously slow because for each bone the entire scene is set to each frame

Yes it's very ugly, unfortunately I am not aware of other way. :(

MoritzBrueckner:
I thought about manually iterating through the fcurves and combine the individual transformations to matrices, but I don't know if that works. First I have to understand Blender's animation system better. Would be really nice if Blender had a way to retrieve the matrices for arbitrary frames.

PR #1981, opened 2020-11-01
Add worldVecToOrientation method

knowledgenude:


Sources/armory/object/TransformExtension.hx
```diff
@@ -37,4 +37,12 @@ class TransformExtension {
 	public static inline function getLocalVecFromWorld(t: Transform, worldVec: Vec4): Vec4 {
 		return worldVec.clone().applymat4(Mat4.identity().getInverse(t.worldUnpack));
 	}
+	/**
+	* Returns the given world vector in transform orientation
+	* @param worldVec
+	* @return Vec4
+	**/
+	public static inline function worldVecToOrientation(t: Transform, worldVec: Vec4): Vec4 {
+		return t.world.right().mult(worldVec.x).add(t.world.look().mult(worldVec.y)).add(t.world.up().mult(worldVec.z));
```
MoritzBrueckner, line 46: Maybe this line can be simplified a bit to make it more readable (also, `Transform`s already have look/right/up functions):

```haxe
return new Vec4(t.right().mult(worldVec.x), t.look().mult(worldVec.y), t.up().mult(worldVec.z), worldVec.w);
```

Also, I changed the w coordinate to the coordinate of the `worldVec` vector to preserve if this vector is a location or direction.

knowledgenude: Looks much better, but returning a new Vec4 give `TransformExtension.hx:46: characters 19-45 : iron.math.Vec4 should be kha.FastFloat` and `TransformExtension.hx:46: characters 19-45 : For optional function argument 'x'`

What would be?

MoritzBrueckner: Oh of course, I was dumb and thought only of the standard basis. Yep, adding vectors is the way to go here, so my comment is obsolete :)

knowledgenude: It is really confuse to understand. But your commentary is not obsolete because i had to change it, as `Transforms already have look/right/up functions`. Thanks :+1:

PR #1982, opened 2020-11-01
Some more (late) fixes to the node upgrading

niacdoial:
turns out, some of the nodes's `__init__` functions didn't call their superclass's `__init__`
Also, I had to make `NodeReplacement.Identity` more generic for the detection of node properties. Not perfect, but until there's a way to track what node attributes are properties used in haxe, this might have to do.


luboslenco:
Thanks for contributing!

PR #1984, opened 2020-11-02
Update nodes that own 'On Local Axis'

knowledgenude:
* Added `On Local Axis` checkbox to `Apply Torque`, `Apply Torque Impulse` and `Get RB Velocity`
* Replaced the method in the other nodes for `worldVecToOrientation()`

This line does not makes sense for me:

`inputs.length > 3 ? inputs[3].get() : false`

If the inputs length > 3, isn't to get the input 4?

And is this verification necessary? Because i will verify it in other nodes later.

### DIFF
--- a/Sources/armory/logicnode/ApplyForceAtLocationNode.hx
+++ b/Sources/armory/logicnode/ApplyForceAtLocationNode.hx
@@ -15,9 +15,9 @@ class ApplyForceAtLocationNode extends LogicNode {
 	override function run(from: Int) {
 		var object: Object = inputs[1].get();
 		var force: Vec4 = inputs[2].get();
-		var localForce: Bool = inputs.length > 2 ? inputs[3].get() : false;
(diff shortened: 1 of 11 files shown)


MoritzBrueckner:
> If the inputs length > 3, isn't to get the input 4?

The fourth input yes, but not input 4. Computers start counting at 0, so the first element has index 1, the second has index 2, and so on.

> And is this verification necessary? Because i will verify it in other nodes later.

You don't need to verify things twice in general, but make sure that the program can't fail because of that between the first (removed) verification and the second verification. In this case the verification is in place to ensure that there is no "index out of bounds" exception when the input doesn't exist, so you shouldn't remove it here.

knowledgenude:
Looks like people does not approves this changes

Testing it in Krom with `Apply Force` and there is a performance gain from 1.8 ms to 1.2 ms here: [untitled.zip](https://github.com/armory3d/armory/files/5477940/untitled.zip)

I don't know about the other targets

What may be wrong at all?

PR #1985, opened 2020-11-02
Revert/fix for traits updating

davidbruce:
Changes compilation_server_done() back to before my change.  Updates the path it checks to krom/krom.js.temp, this should prevent it from deleting existing krom.js files on first project launch.


davidbruce:
This is related to: https://forums.armory3d.org/t/error-trait-referenced-in-object-or-scene-not-found/4298/6

PR #1994, opened 2020-11-07
Upgrading the Vector Math Node

e1e5en-gd:
Dynamically change the number of input and output parameters for the selected operation.
ln_vector_all

### DIFF
--- a/Sources/armory/logicnode/VectorMathNode.hx
+++ b/Sources/armory/logicnode/VectorMathNode.hx
@@ -4,49 +4,136 @@ import iron.math.Vec4;
 
 class VectorMathNode extends LogicNode {
 
-	public var property0: String;
-	var v = new Vec4();
+	public var property0: String; // Operation
+	public var property1: Bool;   // Separator Out
+	var res_v = new Vec4();
+	var res_f = 0.0;
 
 	public function new(tree: LogicTree) {
 		super(tree);
 	}
 
 	override function get(from: Int): Dynamic {
-		var v1: Vec4 = inputs[0].get();
-		var v2: Vec4 = inputs[1].get();
-		if (v1 == null || v2 == null) return null;
-		v.setFrom(v1);
-		var f = 0.0;
+		var p1: Vec4 = inputs[0].get();
+		if (p1 == null) return null;
+		res_v.setFrom(p1);
 		switch (property0) {
-		case "Add":
-			v.add(v2);
-		case "Subtract":
-			v.sub(v2);
-		case "Average":
-			v.add(v2);
-			v.x *= 0.5;
-			v.y *= 0.5;
-			v.z *= 0.5;
-		case "Dot Product":
-			f = v.dot(v2);
-			v.set(f, f, f);
-		case "Cross Product":
-			v.cross(v2);
-		case "Normalize":
-			v.normalize();
-		case "Multiply":
-			v.x *= v2.x;
-			v.y *= v2.y;
-			v.z *= v2.z;
-		case "Length":
-			f = v.length();
-		case "Distance":
-			f = v.distanceTo(v2);
-		case "Reflect":
-			v.reflect(v2);
+			// 1 arguments: Normalize, Length 
+			case "Normalize":
+				res_v.normalize();
+			case "Length":
+				res_f = res_v.length();
+			// 2 arguments: Distance, Reflect
+			case "Distance": {
+				var p2: Vec4 = inputs[1].get();
+				if (p2 == null) return null;
+				res_f = res_v.distanceTo(p2);
+			}
+			case "Reflect": {
+				var p2: Vec4 = inputs[1].get();
+				if (p2 == null) return null;
+				res_v.reflect(p2); 
+			}
+			// Many arguments: Add, Subtract, Average, Dot Product, Cross Product, Multiply
+			case "Add": {
+				var p2 = new Vec4();
+				var i = 1;
+				while (i < inputs.length) {
+					p2 = inputs[i].get();
+					if (p2 == null) return null;
+					res_v.add(p2);
+					i++;
+				}
+			}
+			case "Subtract": {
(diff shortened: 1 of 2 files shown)


MoritzBrueckner:
Hi, I'm not sure if this is directly related to this PR, but the replacement routine of the vector math node seems to fail for some files. In the [archer template](https://github.com/armory3d/armory_templates/tree/master/archery) the following error occurs:

```
  File ".../armory/blender\arm\logicnode\math\LN_vector_math.py", line 33, in set_bool
    self.outputs.remove(self.outputs.values()[-1]) # Result Y
IndexError: list index out of range
File ".../armory/blender\arm\logicnode\math\LN_vector_math.py", line 15, in set_bool
there were errors in node replacement
```

e1e5en-gd:
I will definitely check if possible. So far I can suggest (I'm not mistaken) that the description of nodes is saved in Blender and in order for them to update to new versions, you need to delete the old ones and add new ones. It turns out that the look is old, but the treatment is new.

e1e5en-gd:
1. `Vector Math` - fixed a bug with removing output parameters (https://github.com/armory3d/armory/pull/2010), but this does not solve the problem. The fact is that due to the different number of output parameters, the update does not work completely, and as a result, a node appears to become 2: a new connected and an old "half" connected.
error_update_nodes_two_nodes

2. The example archery template itself is not updated anyway due to other nodes.

error_update_nodes_1

I tried on version 2020.10 (when there was no update of the math nodes yet), the situation is the same - other nodes are not updated.


Maybe you should make a mistake on the update system?

3. The example starts, but when trying to shoot it hangs with an error:
```
Trace: TypeError: Cannot read property 'name' of null
    at <anonymous>:9886:19
    at <anonymous>:9950:6
    at iron_Scene.returnObjectLoaded (<anonymous>:10203:3)
    at iron_Scene.returnObject (<anonymous>:10170:9)
    at iron_Scene.createObject (<anonymous>:9944:9)
    at spawnObjectTree (<anonymous>:9875:11)
    at iron_Scene.spawnObject (<anonymous>:9891:3)
    at armory_logicnode_SpawnObjectNode.run (<anonymous>:2334:21)
    at armory_logicnode_RemoveObjectNode.runOutput (<anonymous>:1634:8)
    at armory_logicnode_RemoveObjectNode.run (<anonymous>:2089:8)
```
You must manually update the example.

MoritzBrueckner:
Regarding the node replacement system: I still have no idea how it works because I didn't take time to look into it so far. Maybe @niacdoial can help here?

Maybe the replacement system should be opt-in, so that it only takes nodes into account which explicitly state that they need a replacement for certain versions?

niacdoial:
alright so.
> I still have no idea how it works because I didn't take time to look into it so far.

I might need to comment the stuff better, I guess.

So, one of the things I've seen while looking at this thread alone is that the armory version from the file is 0.6. Between 0.6 and 2020.8 (I think), some of those nodes might have changed, and the node updater doesn't know that, because it wasn't implemented yet. I guess I could create an update function that tries to guess from the node's inputs/outputs, and does the right correction.

Also, a node's replacment routine has to be edited each time a its input/output/property list gets changed. In particular, there needs to be a `.get_replacment_node()` method implemented on that node class now.

---

> Maybe the replacement system should be opt-in, so that it only takes nodes into account which explicitly state that they need a replacement for certain versions?

so, create a way to "step up" a node's versions without the need to replace the already-existing nodes?
That would be possible, but at that point, it would be even harder to understand how a node needs to be implemented. I'm already kinda unhappy with how complex/unintuitive the `.get_replacement_node()` method is to implement, and adding the option to have it *not* give a replacement would make it worse.

Another possibility (that still adds complexity) would be to add an `arm_version_sub` property, so only the compatibility-breaking changes would change `arm_version`.

IDK, I need to think of the best way to tackle that.

I'll try and do stuff tomorrow, it's getting too late right now.
See you.

MoritzBrueckner:
Oh, I didn't see that the file is from 0.6. Maybe its better then to just update the file. No need to make things even more complicated in order to work with files that were made before the replacement version was put into place :) Maybe it is even possible to further simplify the replacement system without giving functionality away, I don't know.

What I meant with opt-in is that maybe nodes should not be replaced at all when they don't have a replacement routine. Or does it already work like that and its just spitting out a warning message? As I said, I didn't look into the replacement system yet but I often see messages when opening older files that confuse me a bit because I think those nodes weren't updated at all and it looks like there were errors/problems replacing them.

The most important thing is that the bug was fixed in https://github.com/armory3d/armory/pull/2010 :)

niacdoial:
> What I meant with opt-in is that maybe nodes should not be replaced at all when they don't have a replacement routine.  Or does it already work like that and its just spitting out a warning message?

As it is now, one of the warnings/errors consists in exactly that: telling that a replacement routine isn't defined, and gracefully skipping to the next node.
It even is the one in the second screenshot of https://github.com/armory3d/armory/pull/1994#issuecomment-727161566
Actually, I should be more clear that this is more of a warning than an error.
I am still perplexed by the duplicated nodes, though. I'll need to reproduce those problems and see for myself.

> I often see messages when opening older files that confuse me a bit because I think those nodes weren't updated at all and it looks like there were errors/problems replacing them.

In that case, please send those errors to me. There was already the problems with some nodes, fixed in https://github.com/armory3d/armory/pull/1982.
What happened there is that there is a replacement that happens on all non-versioned nodes. The replacement creates a new node that is identical, save for the fact that it is now capable of storing version information.
So, in a way, there *is* a replacement procedure that can fail, even if what it does is essentially nothing.

e1e5en-gd:
In any case, I will briefly once again describe the problems I encountered when updating the archery template project:
1. `Vector Math` - duplication has occurred, but a new node has been created and linked correctly;
2. `Play Action` - the new name of this `Play Action From` node, most likely that is why the nodes were not updated;
3. All references to objects in nodes (`Object` field) are empty.

Issue #1009, opened 2018-11-16
Request: on change single signal node


Gedankenexper1ment:
It would be very nice to have a node that sends a single pulse (red output) when a value (bool, number, etc) changes. This would be useful in many situations, but especially so for converting constant signals (like keyboard down or on overlap) to single pulses (like keyboard started). This allows us to use set variable to count, and send signals to nodes that work best with a single pulse (like print or sequence).

BlackGoku36:
You can use Event nodes.

PR #1999, opened 2020-11-08
Value Changed node

knowledgenude:
* `Changed output`: is activated when the connected value is changed.
* `Unchanged output`: is activated when the connected value returns to the initial.

Does not works with `null`, because the initial value will be defined every time until it is not null. I found this https://lib.haxe.org/p/notifier/ but i don't have the required knowledge to implement (and i don't even know if this is related to the purpose of this node).

Also does not works right for vectors in the test i do:
```
vec = inputs[1].get();
vec == inputs[1].get(); // prints false. I would need to use almostEqual here, but why?
```

Linked issue: https://github.com/armory3d/armory/issues/1009

Example: [TestNode.zip](https://github.com/armory3d/armory/files/5507150/TestNode.zip)

Captura de tela de 2020-11-08 17-46-42

### DIFF
--- a/Sources/armory/logicnode/ValueChangedNode.hx
+++ b/Sources/armory/logicnode/ValueChangedNode.hx
@@ -1,0 +1,25 @@
+package armory.logicnode;
+
+class ValueChangedNode extends LogicNode {
+
+	var initial: Dynamic;
+	var value: Dynamic;
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function run(from: Int) {
+		if (initial == null) {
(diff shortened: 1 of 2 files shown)


luboslenco:
> Also does not works right for vectors in the test i do:
> vec = inputs[1].get();
> vec == inputs[1].get(); // prints false. I would need to use almostEqual here, but why?

Vector node constructs a new Vec4 object every time: https://github.com/armory3d/armory/blob/master/Sources/armory/logicnode/VectorNode.hx#L20, because of issues mentioned at https://github.com/armory3d/armory/pull/1653.

Issue #1990, opened 2020-11-06
Bump node does not compile


knowledgenude:
I don't know if this is already supported but is returning errors:

```
build_untitled/compiled/Shaders/Material_001_mesh.frag.glsl:7: 'n' : undeclared identifier 
build_untitled/compiled/Shaders/Material_001_mesh.frag.glsl:7: '=' :  cannot convert from ' temp float' to ' temp 3-component vector of float'
build_untitled/compiled/Shaders/Material_001_mesh.frag.glsl:7: '' : compilation terminated 
```
```
SPIR-V is not generated for failed compile or link

Compiling shader 1 of 19 (Material_001_mesh.frag.glsl) failed:
Shader compiler error.
(node:7132) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Shader compiler error.
(node:7132) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

MoritzBrueckner:
Hi, the bump node is already supported, this problem occurs because of some wrong shader generation. Can you share an example file or a few steps to reproduce?

knowledgenude:
@MoritzBrueckner Hm, maybe i am using it in the wrong way... the nodes sockets are not of the same color, but in eevee it works as expected.

Example: [bump.zip](https://github.com/armory3d/armory/files/5503035/bump.zip)

Setup:
![Captura de tela de 2020-11-06 16-11-26](https://user-images.githubusercontent.com/63247726/98405284-c2a20980-204a-11eb-8ea9-9782e4bcea78.png)

MoritzBrueckner:
There are actually multiple issues here:

- The musgrave texture node was "returning" a color even if it no longer has a color output (not sure if it existed before). PR that fixes this: https://github.com/armory3d/armory/pull/1993
- The bump mapping node tries to access normals that aren't written in the shader yet. This only occurs when the `Armory Project > Export Tangents` setting is active, but if it's not there is another error also happening in the normal map parsing (`'vVec' : undeclared identifier`).

  It seems that there are many open issues with normal maps (also see https://github.com/armory3d/armory/issues/1968), the normal map parsing urgently needs an update. When I try to make bump mapping work with tangents I just get a black mesh, so I still have to figure out what I'm doing wrong.

Thanks for the report as always!

Issue #1968, opened 2020-10-29
Wrong texture color


knowledgenude:
I don't usually use node shaders, so i can't give a better name for this issue.

Basically it is a node setup that put "grass" in the top of the object. Everything works cool without textures, but when i add even a single diffuse texture, it have a wrong color:

Test file contains textured and not textured object: [GrassOnTop.zip](https://github.com/armory3d/armory/files/5455823/GrassOnTop.zip)

Eevee:
![Captura de tela de 2020-10-28 23-42-20](https://user-images.githubusercontent.com/63247726/97519049-4f780380-1977-11eb-90a6-b821bd8c8c7d.png)

Armory:
![Captura de tela de 2020-10-28 23-42-03](https://user-images.githubusercontent.com/63247726/97519067-59016b80-1977-11eb-9508-165765b40704.png)

thelabcat:
I hope I’m not putting my foot in my mouth here... 

I am viewing this by email, so apologies if this is no help. I used to be scared of the compositor and node editor, but it’s actually really easy. I recommend you look into how to do that. You probably need to just set the diffuse shader’s color.

thelabcat:
Blender basics:
A shader defines things like metallic-ness or wether it’s clear or not. Can have basic single color.
A texture is a color or value map that can be from an image or a formula, and typically goes into the color of the material. Sometimes, though, it might go into the roughness, or even the transparency of it. One material may have several textures it uses.
Bonus: a UV map has nothing to do with ultraviolet. U and V are the two axes of a coordinate system for mapping textures onto a 3D mesh. A well made UV map will look a bit like how you would cut fabric to sew/make the mesh in real life.

thelabcat:
It looks like in Armory the material disappears completely. I had that happen in BGE sometimes, mostly because BGE did not support any kind of texture coordinates but good hard UV.

I have not downloaded Armory, let alone the test file 😋 Nevertheless hope this helps 😊

MoritzBrueckner:
I think the issue is that there is currently no blending between normal maps (it is a bit more [complicated](https://blog.selfshadow.com/publications/blending-in-detail/) than it is the case for other texture maps), so the first parsed normal map is used (in this case, it is the rock normal map). The normal map is also taken into account for the color ramp (blender doesn't do this and Armory probably shouldn't do this as well) resulting in this mixture of textures.

[...]

PR #2006, opened 2020-11-12
Fix usage of normals in shaders + smaller shader improvements

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1990.

Previously it could happen that normals or other attributes like `vVec` were accessed in shaders before they were declared. Also, the w coordinate socket was mistakenly used as the scale in the musgrave texture node.

This PR was tested with various example files including normal mapping, bump mapping, displacement, etc., tangent export on/off and in forward and deferred render paths.

@luboslenco I'm interested in implementing normal map blending when I get some more time again, do you recommend one certain method? I wrote a bit about it here: https://github.com/armory3d/armory/issues/1968#issuecomment-718766196.

### DIFF
--- a/blender/arm/material/cycles.py
+++ b/blender/arm/material/cycles.py
@@ -365,7 +365,7 @@ def parse_normal_map_color_input(inp, strength_input=None):
         frag.write('mat3 TBN = cotangentFrame(n, -vVec, texCoord);')
         frag.write('n = TBN * normalize(texn);')
     else:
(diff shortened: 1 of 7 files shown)


luboslenco:
Thanks!

> I'm interested in implementing normal map blending

Sounds awesome! RNM from [the article](https://blog.selfshadow.com/publications/blending-in-detail/) you linked looks nice (+ [shadertoy example](https://www.shadertoy.com/view/4t2SzR)). We should only implement one for a start, to keep it simple.

PR #2011, opened 2020-11-14
Map Range node

knowledgenude:
I guess it  is the same but without clamp checkbox: https://docs.blender.org/manual/en/latest/compositing/types/vector/map_range.html

Use: you have the vertical rotation of the camera axis object and you need to get it in camera FOV to increase the sense of depth when the player looks up or down.

This nodes takes the vertical rotation value (i.e. from -1.5 to 1.5) and changes it to FOV value (i.e. from 1 to 2).


luboslenco:
Thanks!

Issue #864, opened 2018-09-07
Improve unit testing


luboslenco:
Going towards 0.6, it's time to add more tests to the https://github.com/armory3d/armory_ci to minimize regressions.

PR #2013, opened 2020-11-17
More renames and cleanups

knowledgenude:
* Added more specifc labels like: `While` to `While True`, `Sequence` to `Output Sequence`;
* More intuitive names for "conversion" nodes: `Vector From Boolean` to `Boolean to Vector`, `Vector From Transform` to `Transform to Vector`;
* Added more descriptions and fixed others;
* Added spaces separating outputs and inputs in the .py files;
* Some cleanups in my past authored nodes;
* Reorganized the layout of some scripts respecting the XYZ order and past (started), present (down), after (release) in input nodes and put more convenient default values;
* Fixed capitalization in all labels;
* Fixed Map Range. Source: https://gist.github.com/xposedbones/75ebaef3c10060a3ee3b246166caab56


MoritzBrueckner:
Nice! A few remarks:

- `While True` is ok, but I'm not sure if it really helps. Most people will probably just search for `While`.
- Replacing `True` by `1` is probably not a good idea. Yes, it is easier to write and you can't spell `1` wrong, but readability lacks a lot. It will make it very hard to differ between bool and int arguments when calling functions for example. Does `myFunc(1)` require an integer or a boolean?

  A good "explanation" can be found in the following Python easter egg:
  Open Python in a console and type
  ```python
  import this
  ```
  and hit enter :)

knowledgenude:
> Replacing `True` by `1` is probably not a good idea. Yes, it is easier to write and you can't spell `1` wrong, but at the cost of readability. It will make it very hard to differ between bool and int arguments when calling functions for example (does `myFunc(1)` require an integer or a boolean)?
>
Undone, thanks for the tip!

About this line in some specific nodes: 

```
import bpy

```
This line is in `Get Object by Name`. Is there some reason for this line just in this node and why the other ones don't need it?

MoritzBrueckner:
Both imports are required. `bpy` is required for `bpy.data` in the last line of this file and the other import makes all the node functions accessible (although the * import is generally rather discouraged).

Sources/armory/logicnode/GetCursorStateNode.hx
```diff
@@ -13,7 +13,7 @@ class GetCursorStateNode extends LogicNode {
 
 		return switch (from) {
 			case 0:
-				mouse.hidden ? return false : mouse.locked ? return false : return true;
+				mouse.hidden ? mouse.locked ? return true : return false;
```
MoritzBrueckner, line 16: What exactly is happening here? It looks like you return something inside of another return statement (the one before the `switch`). Also, the two nested ternary if statements might be a bit confusing, maybe you could use parentheses to make it clear how the nesting works.

knowledgenude: There is some way to test all .hx files at once without add them in the node tree? I would like to make sure that I haven't touched anything I shouldn't have.

MoritzBrueckner: It looks like it just returns whether the mouse is hidden and locked, right? Then you can just write `mouse.hidden && mouse.locked`. The `return` statement comes before the switch, so there is no need to use it again.

What I mean with nesting and parentheses is the following:

```haxe
cond1 ? (cond2 ? doSomething2() : doSomethingElse2()) : doSomethingElse1();
```

This syntax makes it clear which action belongs to which condition. If there are no brackets/parantheses, it can be a bit difficult to see what is going on.

> Is there some way to test all .hx files at once without add them to the node tree? I would like to make sure that I haven't touched anything I shouldn't have.

Yes. This is called unit testing, but Armory doesn't really do that (see https://github.com/armory3d/armory/issues/864). Unit tests usually test certain functions by giving them example inputs and test whether the output is correct ([simple example in Haxe](https://gitlab.com/koui/Koui/-/blob/master/Tests/elements/TestTextInput.hx)). This can be automated to be run on each merge request for example.

PR #2024, opened 2020-11-26
Improve `Get Mouse Movement` node

MoritzBrueckner:
- Rename misleading `Delta` socket to `Wheel Delta` (on the Discord channel there was someone asking about how to get the mouse wheel rotation, this is the main reason for this PR)
- Reordered the sockets to make a bit more sense (movement and delta values are now grouped together)
- Node output defaults to `0` now instead of `null` for unsupported operations. This prevents errors on other places
- Better docstring

getmousemovementnode

I'm not even sure if `Mouse Delta` makes sense in this node or if it should get it's own node (it is not per se a "movement" of the mouse I would say?), but I don't know. We should at least keep it for compatibility reasons.

### DIFF
--- a/Sources/armory/logicnode/GetMouseMovementNode.hx
+++ b/Sources/armory/logicnode/GetMouseMovementNode.hx
@@ -1,10 +1,8 @@
 package armory.logicnode;
 
-import iron.math.Vec4;
+import kha.FastFloat;
 
 class GetMouseMovementNode extends LogicNode {
-
-	var coords = new Vec4();
 
 	public function new(tree: LogicTree) {
 		super(tree);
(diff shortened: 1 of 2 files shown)


kaiwas:
I got lost and confused (

Scr_20201202_092459

MoritzBrueckner:
Hi, I think this is because of Blender, it doesn't update nodes that are already in a node tree. You should probably add a new `Get Mouse Movement` node and everything will look fine again.

kaiwas:
Maybe then you can come up with something that will allow you to renew the nodes with each improvement of the armory?  (if possible)
Not everyone knows how to edit the code and will face similar problems.

MoritzBrueckner:
Such a system does already exist but I can't get it to work unfortunately (@niacdoial when using a NodeReplacement with the same node as `from_node` and `to_node` it creates another duplicate node without reconnecting the sockets, do you know why?).

niacdoial:
I don't know, sorry. I need to look into that. It already was a problem last time you contacted me about that, and I didn't manage to take the time to look into that.
I will try, but I have no idea when I'll be able to. (even in log-scale, I have no idea how long it will be.)

MoritzBrueckner:
> (even in log-scale, I have no idea how long it will be.)

Nicely said :D Don't feel rushed by me, maybe I have time this weekend for a debug session then.

MoritzBrueckner:
@niacdoial Found the bug, a PR will come soon. But while I'm at it: is there a reason why the detailed report is written to a file instead of the console (maybe with colors)? I'm not talking about the popup, but if there is a reason for the files I could add another button next to "Ok" that opens the project folder with the reports.

niacdoial:
OK, thanks!

For the file, what I thought was that if someone opens the file, and saves it, any problem might get only reported once, and silently slip into the project, not to be reported until the next time armory gets updated.
So the file next to the project file was supposed to serve as a reminder, until the one who edits the project comes around and checks that the problems are dealt with.
But yeah, I guess those problems could also be reported in the console.

Issue #1398, opened 2019-09-23
Post Processs - Chromatic Aberration / Bloom VS Deferred & Forward Clustered


ControllerArts:
**Description**
- Chromatic Aberration is not working with Forward Clustered but is normal in Deferred.
- Bloom has a nice control and looking in Forward method but in Deferred varies so much with small tweaking and the behavior doesn't feel natural. 

**System**
Blender: 2.80-2019 07 29
Armory: 2019.9.0
OS: Windows 10.0.17763
Graphics card: Intel UHD Graphics 630 & Radeon RX580

PR #2027, opened 2020-11-28
Forward renderpath: add support for chromatic aberration

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/1398

This PR adds support for chromatic aberration in the forward renderpath, before this the setting would only work on deferred rendering. Also the compositor flags weren't reset between builds, so until now disabling the `Realtime postprocess` flag for example wasn't taken into account until restarting Blender.


luboslenco:
Thanks!

PR #2029, opened 2020-11-29
Added functions to disable/enable collision

knowledgenude:
Useful to remove the collision from gates.

The code was taken from here: https://pybullet.org/Bullet/phpBB3/viewtopic.php?t=11690


luboslenco:
👍👍

PR #2033, opened 2020-12-01
added lod_screen_size property in properties window

notwarp:
hi, i've added a properties that can help control the global size of the object for the lods calculation, don't know if can be a good solution but in this way i can control the lods for large objects. Hope this can help other for achieve a better solution than mine.


luboslenco:
Hello, thanks for the pull request! The lod screen size calculation is not done properly right now which is probably causing issues for you - we should fix the size calculation at https://github.com/armory3d/iron/blob/master/Sources/iron/object/MeshObject.hx#L335 instead.

notwarp:
ok, so you say that is better calculate the logic all inside that function?

Issue #2044, opened 2020-12-04
Problem with On Volume Node


chuckO99:
Thank you for contributing to Armory!

I have a racetrack which uses the On Volume Trigger to sense the car. The latest rev of Armory seems to have broken this node. The node  works in the Overlap mode, but the Enter or Leave modes won't trigger the event. No error message, the node just doesn't work. It worked fine on Armory 2011

**System**
Blender: 2.893
Armory: 2012
OS: Windows 10
Graphics card: RTX2070 Super

**Test File**
*(drag & drop the zipped .blend file here)*

knowledgenude:
Sorry, this was my fault. Will be fixed in https://github.com/armory3d/armory/pull/2045

Thanks to report! :+1:

PR #2045, opened 2020-12-04
Fix volume trigger

knowledgenude:
Fix: https://github.com/armory3d/armory/issues/2044


QuantumCoderQC:
Hi @knowledgenude , Thanks for the fix. The on Volume trigger node has inputs labeled as `RB1` and `RB2`. meaning a rigid body must be used for this node. But. this is not actually necessary. A rigid body is not needed here. So, it would be better to change RB to Object 1 and 2.  This would also help reduce misunderstandings. 

Best Regards,
QC

PR #2046, opened 2020-12-04
Fix and further improve node replacement system

MoritzBrueckner:
- Fixed a bug where the already added replacement node wasn't deleted when an exception occurred, resulting in two nodes
- Moved the replacement code into its own module because it is quite a lot of code (also we probably should soon think about how to handle those ever-growing files like `props_ui.py`, Armory needs a better package structure)
- Added a button to the error dialog that lets you quickly open the project folder with the generated report file
- The traceback of unknown exceptions is now included in the generated report
- Some general cleanup and small message improvements

@niacdoial Is it ok for you if I credit you in the `replacement.py` module docstring, now that the original git history is rather hidden because the code was moved to a new module?

### DIFF
--- a/blender/arm/logicnode/arm_nodes.py
+++ b/blender/arm/logicnode/arm_nodes.py
@@ -7,6 +7,8 @@ import bpy.types
 from bpy.props import *
 from nodeitems_utils import NodeItem
 
+# Pass NodeReplacment forward to individual node modules that import arm_nodes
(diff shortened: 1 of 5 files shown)


niacdoial:
@MoritzBrueckner No problem with being credited!
Thank you!

niacdoial:
thanks!

Issue #2014, opened 2020-11-17
Display operations in node label


knowledgenude:
![Captura de tela de 2020-11-17 19-05-51](https://user-images.githubusercontent.com/63247726/99456357-07aa2380-2908-11eb-853e-9e7bf1c3a92d.png)

![Captura de tela de 2020-11-17 19-09-06](https://user-images.githubusercontent.com/63247726/99456586-67083380-2908-11eb-81fe-92ac7878497e.png)

e1e5en-gd:
An interesting option, but you only need to do this for certain types of nodes. I would suggest only for math nodes, and naming them not `Divide`, but `Math: Divide` and `Vector Math: Divide`, respectively.
My assumptions about the difficulties that may arise for other nodes:
- `'Space' Down` - you can understand that this is `Keyboard`. But beginners can find it difficult to navigate and formulate their questions. He had a problem, how can he understand/remember from the name `‘Space’ Down` that this is the `Keyboard` node?
- `‘Up’ Down`, `’A’ Down`, `‘B’ Down` are the `Keyboard` or `Gamepad` or `Virtual Button` nodes?

If you just add, the name of the nodes can become long, for example, `Keyboard Down ‘Space’`.
You need to be very careful about the name so as not to confuse the user.

By the way, it is possible for nodes in which you can dynamically change the number of parameters, indicate how many values ​​in the name (for example, in arrays, switch and others).

knowledgenude:
@e1e5en-gd , that makes sense. Also as the keyboard node is not that much repeated in the node tree (only a few commands) then the user can name it as it want :+1:

MoritzBrueckner:
Had some ideas, open for discussion:

![nodelabel_merge](https://user-images.githubusercontent.com/17685000/99579128-0cd9a200-29de-11eb-9734-641bbf8f5490.gif) ![nodelabel_loop](https://user-images.githubusercontent.com/17685000/99579130-0d723880-29de-11eb-949b-77c04fb698d0.gif)
![nodelabel_object](https://user-images.githubusercontent.com/17685000/99579136-0ea36580-29de-11eb-8f17-8eb0e9aab1e5.gif) ![nodelabel_timer](https://user-images.githubusercontent.com/17685000/99579138-0f3bfc00-29de-11eb-8e74-8a65a27a121c.gif)

If the sockets from which the values are taken are connected, either only the regular node label is displayed, or in case of the loop node `x...5` is displayed for example if the `From` socket is connected.

[...]

PR #2058, opened 2020-12-15
Add dynamic node labels

MoritzBrueckner:
Implements https://github.com/armory3d/armory/issues/2014.

Reposting the screenshots from the linked issue (for more details please refer to that):
nodelabel_merge nodelabel_loop
nodelabel_object nodelabel_timer

- Some nodes will not display a dynamic header if the sockets are connected or only parts of the header are dynamic (for example `x...5` for a loop node where the lower value is a connected socket)
- If a user renames a node, Blender will only display that name (= no dynamic header), which gives the user more freedom.

Also, we should probably think about adding a super class for array nodes, they basically have 90% the same code.


luboslenco:
Awesome!

Issue #2054, opened 2020-12-13
On Action Marker does not work


ArmoryBlender:
Hello there!

If I try to print something at the marker's position out, it just doesn't work. I also tried it with the armory3d examples out without success.....I hope this gets fixed!

![Unbenannt](https://user-images.githubusercontent.com/48133099/102015497-49d65180-3d5c-11eb-9796-0461c07410bf.PNG)

Greetings 

Armoryblender

MoritzBrueckner:
This seems to be related to the autobake feature. When it's enabled, there seems to be another action that gets exported (the `action` variable that holds the action with the pose marker gets overridden):

https://github.com/armory3d/armory/blob/f69185a765726c37aba2706d1d45f48e392fcecc/blender/arm/exporter.py#L936

I don't know what the autobake feature does, but the solution could be to either export pose markers before autobake is handled or to save a reference to the original action that is used for marker exporting.

ArmoryBlender:
@MoritzBrueckner Thank you so much for your respond!! I literally never heard of a autobake feature. In the earlier builds of armory3d it worked very fine. This means this feature has been added somewhere in the new builds. I still don't know what you mean by exporting pose markers or save a reference to the original action. The on action node is crucial for my current project, so I'm stuck right now....

knowledgenude:
@ArmoryBlender, maybe you can use the `currentFrame()` function to get when the frame you want is reached https://armory3d.org/api/iron/object/Animation.html

ArmoryBlender:
@knowledgenude Yeessss!! Why did I forget this....thank you so much! Again!! With this method I can check if the frame is on "25".
I'm still wondering why the markers doesnt work anymore...

Greetings 

Armoryblender

MoritzBrueckner:
@ArmoryBlender I have no idea why the markers stopped working, do you loosely know in which version they worked? If I'm not mistaken the autobake feature existed for a long time (I guess it's baking all animation related things (maybe even drivers?) into a new animation to export that, but I don't know much about Blender's animation system or animation in general).

> I still don't know what you mean by exporting pose markers or save a reference to the original action.

That was rather technical (possible solutions for the code), don't worry :) I will do some tests soon to see if they work.

PR #2064, opened 2020-12-17
Fix bone markers export

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/2054.

Bone markers didn't work because the `action` object was overridden while handling the autobake setting.


luboslenco:
Thanks!

ArmoryBlender:
@MoritzBrueckner Thank you very much!! It works perfectly.

Issue #1289, opened 2019-06-05
Limiting the amount of spot light on scene


VeniGogniti:
### SHORT DESCRIPTION:
In scene I have more than 4 spot light but when I press play only 4 of them are visible, others not. After switching off one of the lights, the next ones are switched on during play.
[NearDay.zip](https://github.com/armory3d/armory/files/3257367/NearDay.zip)

Armory: 06 win64
Operating system: win 7
Graphics card & driver: Radeon RX580

indokan1945:
AFAIK it was intended to be maxed at 4. You can change the max size manually by changing the size of maxLights and maxLightsCluster in LightObject.hx, but it will slow down the rendering as the looping mechanism of lights is still need to be fixed.

luboslenco:
IIRC the main pain is handling the shadow maps and all the light types. The cap is set to 16 lights total and 4 lights per cluster. The view is divided into 16x16x16 clusters.

Haxe part:
https://github.com/armory3d/iron/blob/master/Sources/iron/object/LightObject.hx#L30

Shader part:
https://github.com/armory3d/armory/blob/master/Shaders/std/clusters.glsl#L2

A good start could be to simplify uniforms passing:
https://github.com/armory3d/armory/blob/master/Shaders/std/light.glsl#L33

```glsl
// Instead of
uniform mat4 LWVPSpot0;
uniform mat4 LWVPSpot1;
uniform mat4 LWVPSpot2;
uniform mat4 LWVPSpot3;

// Do
uniform mat4 LWVPSpot[maxLightsCluster];
```

Passing array of mat4 uniforms should nowadays work in Kha.

```glsl
// Instead of
if (index == 0) {
	vec4 lPos = LWVPSpot0 * vec4(p + n * bias * 10, 1.0);
	direct *= shadowTest(shadowMapSpot[0], lPos.xyz / lPos.w, bias);
}
else if (index == 1) {
	vec4 lPos = LWVPSpot1 * vec4(p + n * bias * 10, 1.0);
	direct *= shadowTest(shadowMapSpot[1], lPos.xyz / lPos.w, bias);
}

// Do?
vec4 lPos = LWVPSpot[index] * vec4(p + n * bias * 10, 1.0);
direct *= shadowTest(shadowMapSpot[index], lPos.xyz / lPos.w, bias);
```
The problem is the above will cause trouble with WebGL. Alternative solution is to bake shadow maps into single atlas.

Here is where light matrices are bound:
[...]

PR #2073, opened 2020-12-27
Solve artifact with clustering of lights

N8n5h:
Changed how a spot light is detected when sampling light, which caused false positives and produced
that visual artifact described in this issue https://github.com/armory3d/armory/issues/1289#issuecomment-743190855 when a cluster/pixel was detected incorrectly as being lit with a spot light when there were no spot lights in the scene.

### before: 

image

### after:

image


luboslenco:
Awesome!

Issue #2048, opened 2020-12-06
new scene shows old scene?!


zartan9:
![scene](https://user-images.githubusercontent.com/74879950/101272244-5e14c000-3758-11eb-9701-2c9e760879b7.png)
hi im still very new but making progress (found out how to go to new scene on collision!)... but the remnants of the old scene shows even though i removed active and loaded new one...any advice. PS is there a "restart scene" node to replay a level? thanks!

QuantumCoderQC:
Hi @zartan9 If you enable Debug console for Armory, you could clearly see what objects were removed and what was added.

The issue here is I believe the remove scene trait removes the trait itself. Hence, after remove scene, nothing runs anymore. You may test if this is the case by adding a simple print node after the remove scene node.

A way to re-start the scene:
![image](https://user-images.githubusercontent.com/55564981/101337337-ce683200-387b-11eb-9ccd-15448a0c35f8.png)

Hope this helps,

Best regards,
QC

zartan9:
hi, its odd cause when i publish to web the scene seems to change alright! anyway,, trying your node set up works to replays the current scene great! for "print" node what am i typing for string? thanks!

zartan9:
the debug 
![Uploading scene1.png…]()
shows only the 2 cylinders but i can plainly see the box from the prior scene. is there a refresh node or something?

MoritzBrueckner:
Can you maybe share the blend file? This makes it much easier to see what's wrong and if it's a bug we have a good reproducible example. Thanks

zartan9:
great idea! ok here it is.. 2.83 im using... when spin the scene (spacebar etc) and the ball touches plain it goes to next scene. thanks!
[object_rotate.zip](https://github.com/armory3d/armory/files/5705555/object_rotate.zip)

MoritzBrueckner:
It looks like the problem is that the scene to which the game changes has no camera. For me it shows just black without a camera and when adding a camera everything works as expected for me.

zartan9:
hi i added a camera and the previous scene is still showing... can you send a blend file that works on your end too? also how to  see which drivers i need for armory? screenshot below. Thank you.
![camera](https://user-images.githubusercontent.com/74879950/102534205-616f4c00-4074-11eb-8ef2-73051aa70486.png)

MoritzBrueckner:
[...]

PR #2075, opened 2020-12-29
Show warning when scene is missing a world and rp background is set to world

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/2048.


luboslenco:
Thanks!

Issue #2072, opened 2020-12-27
Particle emitter and instanced models - Material Popping visual issue


onelsonic:
if you have custom GLSL shaders like this example here :
https://github.com/armory3d/shader_pack/tree/master/Shaders

then when adding a particles emitter the custom shader will be broken.

Can this be related this property ? "update=assets.invalidate_shader_cache" :
https://github.com/armory3d/armory/blob/4038a67385d3cc0d92d1455f991820c62129eee3/blender/arm/props_renderpath.py#L494
The shader will pop out with wrong vertices and ghost meshes all over the place.

MoritzBrueckner:
Hi, thanks for report :) Do you have a small example file for this? Also, are there any error messages in the console (Blender or browser)?

The line you've mentioned is not likely to be related to this bug. It ensures that the shader cache is deleted when you change the particles property (=> the shaders will be recompiled upon building the game).

onelsonic:
I did others tests and exporting with Krom doesn't work neither in fact.

Below is what I mean, see the artifacts popping in the grass shader (bottom left)
![Animation_2020-12-28-12-46-36](https://user-images.githubusercontent.com/53947594/103212838-402c1c80-490c-11eb-9b62-a2f08bfdcabc.gif)

(also applying the particles system modifier does not affect anything)

Here is the Blend file :  https://github.com/onelsonic/armory_tests/blob/master/particuleswithshaderbug.7z?raw=true
Here is a live demo : https://onelsonic.github.io/particuleswithshaderbug/index.html
(Similar artifacts appears if exporting to Krom)

QuantumCoderQC:
Thanks for the example, I tried it out and I get a different behavior on Krom. In Krom, the grass disappears and appears at random when camera is at a certain position

![bugKrom](https://user-images.githubusercontent.com/55564981/103224993-31a02e00-4929-11eb-8c07-d581af63c4e1.gif)

MoritzBrueckner:
Did some tests and the flickering grass @QuantumCoderQC describes only happens when there is at least one of the particle systems in the scene, but depending on what particle system is deleted other grass objects are not drawn (the grass in front flickers when the yellow particles exist, when only blue particles exist grass in the background flicker).

What irritates me is that Renderdoc displays less draw calls when the flickering grass isn't visible but the debug console shows a constant amount of draw calls and culled objects. Maybe there is a full buffer somewhere?

onelsonic:
[...]

PR #2076, opened 2020-12-30
Show warnings for invalid instancing configurations + cleanup

MoritzBrueckner:
Closes https://github.com/armory3d/armory/issues/2072 (please refer to this for more details).

When multiple objects share one material but only some of them are instanced there can be visible flickering in the game because instance attributes like `ipos` are set with garbage values. In those cases there is now a warning printed. Is there a reason why instancing is set per-object and not per-material?

I moved the instancing part in its own method so that we can display the warning for custom shaders too while having to iterate only once over all material users when _not_ using custom shaders. The reason there is only a warning and no workaround behaviour is that users can then decide what to do (for example split materials or enable instancing for all objects with that material).

I also did some cleanup with import reorganization, more type hints and there were some lines where you could simplify the syntax a bit.

### DIFF
--- a/blender/arm/material/make.py
+++ b/blender/arm/material/make.py
@@ -1,6 +1,7 @@
 import bpy
 import arm.utils
 import arm.node_utils
+from typing import Dict
 import arm.material.make_shader as make_shader
 import arm.material.mat_batch as mat_batch
 import arm.material.mat_state as mat_state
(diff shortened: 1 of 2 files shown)


QuantumCoderQC:
> Is there a reason why instancing is set per-object and not per-material?

I think this is because the instancing data (ipos, iscl and irot) are a separate VertexBuffer, and are associated with each Mesh Object. They are set up [here](https://github.com/armory3d/iron/blob/a2d8e4280ae5ad2e50f017a2dc6d5fa3cf61301e/Sources/iron/data/Geometry.hx#L125)

PR #2077, opened 2020-12-30
[Blender 2.9] Object panel UI

MoritzBrueckner:
UI updates to the object properties panels for Blender 2.9, most of it is just to get similar results to the old 2.8 look. Some visual changes:

**Reordered the trait types to roughly match the amount of usage:**
trait_order

**The proxy menu now uses a column heading instead of a label which seems to be a new Blender convention:**
new_proxy_menu

**The logic node operators under the trait menu now have shorter labels to better fit into the layout:**
node_operators

Unfortunately (as you can see in the last image), there is a small bug in Blender's UI that places the checkbox in template lists a few pixels too much to the left. I hope this gets fixed in the future, I wasn't able to find a workaround.


luboslenco:
👍 👍

PR #2082, opened 2021-01-04
Blender 2.9 LTS support

MoritzBrueckner:
This is the main pull request for the transition to the upcoming Blender 2.9 LTS versions ([issue](https://github.com/armory3d/armory/issues/2074)). It will be in draft status as long as we're working on the migration and maybe even add some more features.

When contributing to the blender2.9 branch, please link to this pull request so that all related PRs are listed below this post. Thanks!

### Pre-Release download:
https://drive.google.com/file/d/1v0prkAMF5tud-OZcxLEhYyZ4jxIZdrSL/view?usp=sharing (*updated april 3rd*)

### Checklist of critical tasks:
- [x] Rigid bodies: `steps_per_second` was replaced by `substeps_per_frame`, leading to an error on export. Likely PhysicsWorld.hx needs to be changed as well or we need to convert the values during export if possible (https://github.com/armory3d/armory/pull/2084)
- [x] The principled BSDF node has a new emission strength input, node indices might be wrong now (https://github.com/armory3d/armory/pull/2081)
- [x] Error when using the search operator menu (`F3`) (https://github.com/armory3d/armory/pull/2086):
   ```
   [...]/armory/blender\arm\props_tilesheet.py", line 93, in poll
       trait = wrd.arm_tilesheetlist[wrd.arm_tilesheetlist_index]
   IndexError: bpy_prop_collection[index]: index 0 out of range, size 0
   ```
- [x] Error when using the search operator menu in the render properties tab (https://github.com/armory3d/armory/pull/2099)
- [x] The new search menu might require some operators to be updated, only those who are exposed in the UI will be found unless explicitely stated: https://wiki.blender.org/wiki/Reference/Release_Notes/2.90/Python_API (https://github.com/armory3d/armory/pull/2083)
- [x] There were some general changes to UI layouts, for example the trait list and Armory player panel look weird in Blender 2.9 (https://github.com/armory3d/armory/pull/2077, https://github.com/armory3d/armory/pull/2093)
- [x] Check if some icons changed (when there are some 2.93 builds released)
- [x] Update the 2.83 LTS warning to 2.9 (https://github.com/armory3d/armory/pull/2176)
- [x] Update the add-ons `bl_info` attribute (https://github.com/armory3d/armsdk/pull/21)

> The entire checklist of tasks can be found here: https://docs.google.com/spreadsheets/d/1uQsNnC2zD-NVYmVETWPj_or23QykmJkbRaSZDML5nHY/edit?usp=sharing

[...]

PR #2086, opened 2021-01-11
[Blender 2.9] Search menu fixes

MoritzBrueckner:

Some various, smaller fixes because more operators are now automatically exposed to the search menu. Some `poll()` methods only work in the correct context.


luboslenco:
👍👍

PR #2093, opened 2021-01-25
[Blender 2.9] Various UI improvements & redesigns

MoritzBrueckner:

A more in-depth discussion and overview of the changes can be found in the [2.9 forum thread](https://forums.armory3d.org/t/armory-upgrade-to-blender-2-9-x/4400).

**Notable changes**:
- Redesigned player panel: 
armory_player

- Improved `Add Trait` menu (that now also shows a checkbox `Is Object` when called from the operator search menu, and only then):
AddTraitFromMenu
AddTraitFromSearch

- Cleaner traits panel: 
traits_panel

- Improved project flags panel:
ArmoryProjectPanel

- Improved exporter panel:
ArmoryExporterPanel

- Redesign of the compositor panel, making it much cleaner:
ArmoryCompositorPanel

- Small improvements in the post process panel
- Added a small custom icon API for simplified icon usage
- Improved the material blending UI a bit
- Ensure that only logic node trees show up as node traits (before, also groups like Armory PBR were shown)
- Fixed the collision filter mask UI

---
In the forum we also discussed a separate Armory properties panel like in the following image, but is not included in this PR as it is a workaround based on a workaround and so on and we couldn't find another workaround to make the view unique for each properties editor (if the panel is opened in one editor, it is opened in all, which is bad).

properties_armory_icon


luboslenco:
Wow!

Issue #2067, opened 2020-12-22
[Android] material_shaders (not working) - Uncaught exception: Can't cast dynobj to iron.data.TSceneFormat


onelsonic:
When adding a custom shader to a material like this example here :
https://github.com/armory3d/armory_examples/tree/master/material_shaders/Shaders

you will get an error on Android:
```
2020-12-22 17:16:23.163 22316-23462/my.app.com D/hl: Uncaught exception: Can't cast dynobj to iron.data.TSceneFormat
2020-12-22 17:16:23.164 22316-23462/? V/threaded_app: android_app_destroy!
```

and this is the only reference I found on this error so far:
https://github.com/HaxeFoundation/haxe/issues/6206

Seems like we have to make sure that objects we are comparing/using are of the same type, somewhere in the code.

Issue #1772, opened 2020-07-18
Only one pointlight at a time possible


Simonrazer:
Links to #1140 #1429, reposting because these issues don't have the error posted and don't get the attention they really need.

**Description**
When using more than one pointlamp this error occurs: 
When using Deferred rendering:
```
Compiling shader 20 of 20 (World_World.vert.glsl).
C:\Users\SK\Desktop\Armory\blends\test\build_test\debug\temp\deferred_light.frag.hlsl(152,20-84): warning X3570: gradient instruction used in a loop with varying iteration, attempting to unroll the loop
C:\Users\SK\Desktop\Armory\blends\test\build_test\debug\temp\deferred_light.frag.hlsl(5,14-30): error X4500: overlapping register semantics not yet implemented 's1'
C:\Users\SK\Desktop\Armory\blends\test\build_test\debug\temp\deferred_light.frag.hlsl(7,14-30): error X4500: overlapping register semantics not yet implemented 's2'
C:\Users\SK\Desktop\Armory\blends\test\build_test\debug\temp\deferred_light.frag.hlsl(9,14-30): error X4500: overlapping register semantics not yet implemented 's3'
C:\Users\SK\Desktop\Armory\blends\test\build_test\debug\temp\deferred_light.frag.hlsl(4,19-26): error X4500: overlapping register semantics not yet implemented 't1'
C:\Users\SK\Desktop\Armory\blends\test\build_test\debug\temp\deferred_light.frag.hlsl(6,19-26): error X4500: overlapping register semantics not yet implemented 't2'
C:\Users\SK\Desktop\Armory\blends\test\build_test\debug\temp\deferred_light.frag.hlsl(8,19-26): error X4500: overlapping register semantics not yet implemented 't3'

Compiling shader 8 of 20 (deferred_light.frag.glsl) failed:
Shader compiler error.
(node:8276) UnhandledPromiseRejectionWarning: Shader compiler error.
(node:8276) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:8276) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
ERROR: Build failed, check console
```
When using Forward rendering:
```
Compiling shader 19 of 19 (World_World.vert.glsl).
[...]

PR #2102, opened 2021-02-04
Add support for shadow map atlasing

N8n5h:
With this it will now possible to enable atlasing of shadow maps, which solves the existing limitation of 4 lights in a scene: https://github.com/armory3d/armory/issues/1289

This is done by grouping the rendering of shadow maps, that currently are drawn into their own images for each light, into one or several big textures. This was done because the openGL and webGL version Armory targets do not support dynamic indexing of shadowMapSamplers, meaning that the index that access an array of shadow maps has to be know by the compiler before hand so it can be unrolled into if/else branching. By instead simply using a big shadow map texture and moving the dynamic part to other types of array that are allowed dynamic indexing like vec4 and mat4, this limitation was solved.

The premise was simple enough for the shader part, but for the Haxe part, managing and solving where lights shadow maps should go in a shadow map can be tricky. So to keep track and solve this, ShadowMapAtlas and ShadowMapTile were created. These classes have the minimally required logic to solve the basic features needed for this problem: defining some kind of abstraction to prevent overlapping of shadowmaps, finding available space, assigning such space efficiently, locking and freeing this space, etc. This functionality it is used by drawShadowMapAtlas(), which is a modified version of drawShadowMap().

### How to use: https://github.com/N8n5h/armory/wiki/Shadow-map-atlas-for-users

image

Blender files I've used for testing: 
[files-for-testing.tar.gz](https://github.com/armory3d/armory/files/5928744/files-for-testing.tar.gz)

Notes:

As explained in the detailed article for maintainers, this is completely governed by the `#arm_shadow_map_atlas` compiler flag to isolate issues related to it within it until it's mature enough, so it should be relatively safe to push even if it wasn't thoroughly tested as I initially planned... but not up to me to decide.
 
* Tested mostly in Krom Linux because of time constraints. 
### **Please test it if you can.** 

* Tried to test it on Krom C but I get this when I try to launch even the default cube https://github.com/armory3d/armory/issues/2067
* I tried to test it on windows but unfortunately my windows installation is acting up so couldn't confirm if it works there, 
I have the suspicion that point lights might bring problems because of the inverted faces, but since the mapping is done manually all platforms should have the same "cubemap standard" so maybe it's safe. Also I suspect this https://github.com/armory3d/armory/issues/1772 might bring problems too but maybe the issue it's circumvented by the fact that no array of shadow map samplers is used for atlases.

* HTML5 seems to not work with multiple point lights for whatever reason. One light works but more than one breaks shadows. Don't seem to be connected to the shadow map size from my tests but could be wrong. I've tried to debug it but unfortunately couldn't make Spector nor webgl-inspector to work, so not sure what is going on in the shader side. 

* Tried to implement dual paraboloid but unfortunately I got stuck and couldn't make it work, here is my progress in case anyone want to check it out [armory](https://github.com/N8n5h/armory/tree/light-fix-2-dpsm) / [iron](https://github.com/N8n5h/iron/tree/light-fix-setfloats-dpsm). There is no doc for this but it should be pretty clear since it's an (attempt of) implementation of the things suggested by the articles and gamedev post linked in the commit message.

This depends on this from iron https://github.com/armory3d/iron/pull/111

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -2,7 +2,6 @@
 
 #include "compiled.inc"
 #include "std/gbuffer.glsl"
-#include "std/light.glsl"
 #ifdef _Clusters
 #include "std/clusters.glsl"
 #endif
@@ -80,14 +79,11 @@ uniform mat4 invVP;
 #ifdef _ShadowMap
 	#ifdef _SinglePoint
 	//!uniform sampler2DShadow shadowMapSpot[1];
-	//!uniform mat4 LWVPSpot0;
+	//!uniform mat4 LWVPSpot[1];
 	#endif
 	#ifdef _Clusters
 	//!uniform sampler2DShadow shadowMapSpot[4];
-	//!uniform mat4 LWVPSpot0;
-	//!uniform mat4 LWVPSpot1;
-	//!uniform mat4 LWVPSpot2;
-	//!uniform mat4 LWVPSpot3;
+	//!uniform mat4 LWVPSpotArray[4];
 	#endif
 #endif
 #endif
@@ -109,21 +105,36 @@ uniform vec2 cameraPlane;
 #ifdef _SinglePoint
 	#ifdef _Spot
 	//!uniform sampler2DShadow shadowMapSpot[1];
-	//!uniform mat4 LWVPSpot0;
+	//!uniform mat4 LWVPSpot[1];
 	#else
 	//!uniform samplerCubeShadow shadowMapPoint[1];
 	//!uniform vec2 lightProj;
 	#endif
 #endif
 #ifdef _Clusters
-	//!uniform samplerCubeShadow shadowMapPoint[4];
+	#ifdef _ShadowMapAtlas
+		#ifdef _SingleAtlas
+		uniform sampler2DShadow shadowMapAtlas;
+		#endif
+	#endif
+	#ifdef _ShadowMapAtlas
+		#ifndef _SingleAtlas
+		//!uniform sampler2DShadow shadowMapAtlasPoint;
+		#endif
+		//!uniform vec4 pointLightDataArray[4];
+	#else
+		//!uniform samplerCubeShadow shadowMapPoint[4];
+	#endif
 	//!uniform vec2 lightProj;
 	#ifdef _Spot
-	//!uniform sampler2DShadow shadowMapSpot[4];
-	//!uniform mat4 LWVPSpot0;
-	//!uniform mat4 LWVPSpot1;
-	//!uniform mat4 LWVPSpot2;
-	//!uniform mat4 LWVPSpot3;
+		#ifdef _ShadowMapAtlas
+		#ifndef _SingleAtlas
+		//!uniform sampler2DShadow shadowMapAtlasSpot;
+		#endif
+		#else
+		//!uniform sampler2DShadow shadowMapSpot[4];
+		#endif
+	//!uniform mat4 LWVPSpotArray[4];
 	#endif
 #endif
 #endif
@@ -132,7 +143,13 @@ uniform vec2 cameraPlane;
 uniform vec3 sunDir;
 uniform vec3 sunCol;
 	#ifdef _ShadowMap
+	#ifdef _ShadowMapAtlas
+	#ifndef _SingleAtlas
+	uniform sampler2DShadow shadowMapAtlasSun;
+	#endif
+	#else
 	uniform sampler2DShadow shadowMap;
+	#endif
 	uniform float shadowsBias;
 	#ifdef _CSM
 	//!uniform vec4 casData[shadowmapCascades * 4 + 4];
@@ -158,6 +175,8 @@ uniform vec3 pointCol;
 uniform sampler2D texClouds;
 uniform float time;
 #endif
+
+#include "std/light.glsl"
 
 in vec2 texCoord;
 in vec3 viewRay;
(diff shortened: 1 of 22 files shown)


MoritzBrueckner:
This is absolutely epic! Must have been a ton of work...

> I tried to test it on windows but unfortunately my windows installation is acting up so couldn't confirm if it works there,
> I have the suspicion that point lights might bring problems because of the inverted faces, but since the mapping is done manually all platforms should have the same "cubemap standard" so maybe it's safe. Also I suspect this #1772 might bring problems too but maybe the issue it's circumvented by the fact that no array of shadow map samplers is used for atlases.

Tested it and it "solves" #1772 when the new atlasing option is enabled. But, unfortunately point lights indeed seem to be problematic:

This is how the scene looks in Blender:



This is how it looks in Krom on Windows (there is no visible difference when using shadow LOD or a single atlas map):



And with spot lights:



Very similar artifacts are visible in the browser (Firefox) when running with html5.

Test file and RenderDoc captures (exported to two formats because I don't know if sharing a capture works): [lights.zip](https://github.com/armory3d/armory/files/5933413/lights.zip)

N8n5h:
> This is absolutely epic! Must have been a ton of work...
> 
> > I tried to test it on windows but unfortunately my windows installation is acting up so couldn't confirm if it works there,
> > I have the suspicion that point lights might bring problems because of the inverted faces, but since the mapping is done manually all platforms should have the same "cubemap standard" so maybe it's safe. Also I suspect this #1772 might bring problems too but maybe the issue it's circumvented by the fact that no array of shadow map samplers is used for atlases.
> 
> Tested it and it "solves" #1772 when the new atlasing option is enabled. But, unfortunately point lights indeed seem to be problematic:
> 
> This is how the scene looks in Blender:
> 
> 
> 
> This is how it looks in Krom on Windows (there is no visible difference when using shadow LOD or a single atlas map):
> 
> 
> 
> And with spot lights:
> 
> 
> 
> Very similar artifacts are visible in the browser (Firefox) when running with html5.
> 
> Test file and RenderDoc captures (exported to two formats because I don't know if sharing a capture works): [lights.zip](https://github.com/armory3d/armory/files/5933413/lights.zip). I hope it helps a bit :)

Ok thanks, yeah I was afraid that directx would be problematic. I have to install windows and play with it to sort it out.

knowledgenude:
I don't know about rendering, but the "artifact" in the second gif reminds me a similar bug that happens to sun lamps. Looks like some wrong calculation of the camera clip start, but i am not stating nothing. It occurs on OpenGL (not tested in DirectX), here is an example file: [test.zip](https://github.com/armory3d/armory/files/5935620/test.zip)

Sun lamp (sorry for the brightness):
Captura de tela de 2021-02-05 19-12-54

The same bug does not occur with other types of lamps, see point lamp:
Captura de tela de 2021-02-05 19-13-23

Maybe it is not a bug of the atlas map but of Armory?

N8n5h:
> Maybe it is not a bug of the atlas map but of Armory?

If the issue can be reproduced with the `Shadow Map Atlasing` option disabled, then it's an issue with armory. Toggling that option off should remove all code related to the feature :)

N8n5h:
Ok, it took a full evening but I managed to reinstall windows and be able to test on it :smile:. The first thing I've tried is just one point light in the scene and surprise, surprise, it works. So it's the same thing that happens in HTML5. Now I've to guess and find whatever is wrong with the implementation that breaks multiple lights for not-opengl targets :sweat_smile:

N8n5h:
After some poking around, I figure that the issue might have to do with this... 

image
(from http://thedev-log.blogspot.com/2012/07/texture-coordinates-tutorial-opengl-and.html)

So I suppose it works with one light because of the coincidence that the face was just in the perfect spot when mirrored

N8n5h:
Pushed a commit that seems to solve the issue with spot and point lights on Windows from my brief tests. Please test if you can to confirm it's finally working :)

MoritzBrueckner:
Works like a charm now, thank you so much!

Regarding html5: if I'm not mistaken, webgl uses a screen coordinate system in the range [-1, 1] with (0, 0) being the center. Maybe that's what causing the issues there?

N8n5h:
> Regarding html5: if I'm not mistaken, webgl uses a screen coordinate system in the range [-1, 1] with (0, 0) being the center. Maybe that's what causing the issues there? That would explain why the artifacts look similar but different.

Will check it out!

> Edit: the issue when the camera is too close to an object still persists unfortunately (can be seen in the gif above when the camera is close above the plane), is that due to a wrong setting somewhere or is it a bug?

Yeah, I noticed it when testing it too, but it seems it's not related to atlas from what I've seen. Can you confirm this too?

MoritzBrueckner:
> Will check it out!

I think it's just the screen/canvas coords, texture coords should be in the usual [0, 1] range.

> Yeah, I noticed it when testing it too, but it seems it's not related to atlas from what I've seen. Can you confirm this too?

Oh, indeed. Maybe there is something wrong with the clustering algorithm in general?

N8n5h:
> I think it's just the screen/canvas coords, texture coords should be in the usual [0, 1] range.

Yes, I tested it by simply commenting this line https://github.com/N8n5h/iron/blob/91728e210b0703822ce42c38b3a79840a9452b72/Sources/iron/object/LightObject.hx#L550 which I think should have been enough to make spot lights work, but it didn't :(

> Oh, indeed. Maybe there is something wrong with the clustering algorithm in general?

Could be, but also could be the algorithm that computes lighting for spot lights too.

I will try playing around with the shaders to see if clusters have anything to do with it, but getting a file where the issue can be reliably reproduced so it's easy to quickly test for changes probably is a better priority. Also I've noticed when using windows that renderdoc allows debugging direct3d, so that probably could be worth looking too, to compare what was going on on a broken pixel versus a working one...

MoritzBrueckner:
Hi, I found a small issue with the new atlasing option: when there is a transparent material (like ArmoryPBR with opacity < 1), the following glsl compiler error is raised:

```
ERROR: [...]\build\compiled\Shaders\std/clusters.glsl:6: 'clusterNear' : undeclared identifier
ERROR: [...]\build\compiled\Shaders\Material_translucent.frag.glsl:3: '#' : preprocessor directive cannot be preceded by another token
ERROR: [...]\build\compiled\Shaders\Material_translucent.frag.glsl:3: '' : compilation terminated
ERROR: 3 compilation errors.  No code generated.
```

When `Shadow Map Atlasing` is disabled, everything works fine. Should I open an issue (+ example file) for this?

N8n5h:
> Hi, I found a small issue with the new atlasing option: when there is a transparent material (like ArmoryPBR with opacity < 1), the following glsl compiler error is raised:
> 
> ```
> ERROR: [...]\build\compiled\Shaders\std/clusters.glsl:6: 'clusterNear' : undeclared identifier
> ERROR: [...]\build\compiled\Shaders\Material_translucent.frag.glsl:3: '#' : preprocessor directive cannot be preceded by another token
> ERROR: [...]\build\compiled\Shaders\Material_translucent.frag.glsl:3: '' : compilation terminated
> ERROR: 3 compilation errors.  No code generated.
> ```
> 
> When `Shadow Map Atlasing` is disabled, everything works fine. Should I open an issue (+ example file) for this?

This should solve it https://github.com/armory3d/armory/pull/2109. Didn't spot that `compiled.inc` gets no special treatment (which I think it's smelly since it has macros for other shaders to work (?)) so with `add_include_front()` (which solves other issue with parsing cluster shaders for forward) I unknowngly let `cluster.glsl` to be on top of `compiler.inc`, and that's what's causing the shader compiling error.

MoritzBrueckner:
Thanks!

PR #2106, opened 2021-02-13
Remove array allocation from SendEventNode

SunDaw:
Remove entries member variable from SendEventNode and SendGlobalEventNode


luboslenco:
Thanks!

PR #2111, opened 2021-02-19
Console improvements

tong:
Minor improvements to terminal output:
- Shows scene name in missing camera warning
- Format scene export time to human readable string
- Format build flags to be copy/paste able

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -2152,7 +2152,7 @@ class ArmoryExporter:
 
         # No camera found
         if not self.camera_spawned:
-            log.warn('Scene "'+self.scene.name+'" is missing a camera')
+            log.warn( f'Scene "{self.scene.name}" is missing a camera')
 
         # No camera found, create a default one
         if (len(self.output['camera_datas']) == 0 or len(bpy.data.cameras) == 0) or not self.camera_spawned:
@@ -2186,7 +2186,7 @@ class ArmoryExporter:
         if self.scene.frame_current != current_frame:
             self.scene.frame_set(current_frame, subframe=current_subframe)
 
-        print('Scene exported in {:0.3f}'.format(time.time() - profile_time) + 's')
+        print('Scene exported in {:0.3f}s'.format(time.time() - profile_time))
 
     def create_default_camera(self, is_viewport_camera=False):
         """Creates the default camera and adds a WalkNavigation trait to it."""
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -2109,7 +2109,7 @@ class ArmoryExporter:
                 self.output['camera_ref'] = self.scene.camera.name
             else:
                 if self.scene.name == arm.utils.get_project_scene_name():
-                    log.warn('Scene "'+self.scene.name+'" is missing a camera')
+                    log.warn(f'Scene "{self.scene.name}" is missing a camera')
 
             self.output['material_datas'] = []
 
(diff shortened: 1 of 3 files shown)


blender/arm/log.py
```diff
@@ -2,8 +2,8 @@
 
 DEBUG = 36
 INFO = 37
-WARN = 35
-ERROR = 31
+WARN = 93
```
MoritzBrueckner, line 5: I'm not sure if it makes sense to use light yellow for warnings (at least that's how it looks like on Windows). If there are large blocks of white text in the console, a yellow warning in between might get overlooked.

blender/arm/make.py
```diff
@@ -432,7 +431,7 @@ def compilation_server_done():
         log.error('Build failed, check console')
 
 def build_done():
-    print('Finished in ' + str( math.ceil((time.time() - profile_time) * 100) / 100) + 's')
+    print('Finished in {:0.3f}'.format(time.time() - profile_time) + 's')
```
SunDaw, line 434: I think you can get rid of the `+ 's'` by moving it into the string (right after `{0:3f}`)

blender/arm/exporter.py
```diff
@@ -2186,7 +2186,7 @@ def execute(self):
         if self.scene.frame_current != current_frame:
             self.scene.frame_set(current_frame, subframe=current_subframe)
 
-        print('Scene exported in ' + str(time.time() - profile_time))
+        print('Scene exported in {:0.3f}'.format(time.time() - profile_time) + 's')
```
SunDaw, line 2189: You can move the `'s'` into the string instead of having it appended after

blender/arm/exporter.py
```diff
@@ -2152,7 +2152,7 @@ def execute(self):
 
         # No camera found
         if not self.camera_spawned:
-            log.warn('No camera found in active scene layers')
+            log.warn('Scene "'+self.scene.name+'" is missing a camera')
```
SunDaw, line 2155: I'd use `f` strings like you did above

blender/arm/make.py
```diff
@@ -62,10 +62,9 @@ def remove_readonly(func, path, excinfo):
 def export_data(fp, sdk_path):
     wrd = bpy.data.worlds['Arm']
 
-    print('\n' + '_' * 10 + '  [Armory] Compiling  ' + '_' * 10)
+    print('Armory v{0} ({1})'.format(wrd.arm_version, wrd.arm_commit))
     if wrd.arm_verbose_output:
-        print('\nArmory v{0} ({1})'.format(wrd.arm_version, wrd.arm_commit))
-        print('OS: ' + arm.utils.get_os() + ', Target: ' + state.target + ', GAPI: ' + arm.utils.get_gapi() + ', Blender: ' + bpy.app.version_string)
+        print('Blender: ' + bpy.app.version_string + ', Target: ' + state.target + ', GAPI: ' + arm.utils.get_gapi())
```
SunDaw, line 67: An `f` string would make this a bit nicer to read

blender/arm/exporter.py
MoritzBrueckner: f-strings are also [quite fast](https://grski.pl/fstrings-performance.html) and in many cases faster than simple concatenation with `+`.

tong:
Thanks for your reviews!

blender/arm/exporter.py
```diff
@@ -2109,7 +2109,7 @@ def execute(self):
                 self.output['camera_ref'] = self.scene.camera.name
             else:
                 if self.scene.name == arm.utils.get_project_scene_name():
-                    log.warn('No camera found in active scene')
+                    log.warn('Scene "'+self.scene.name+'" is missing a camera')
```
SunDaw, line 2112: You missed this one (should also be an f-string) 😄

Issue #1615, opened 2020-03-15
Blending throws error


Naxela:
Thank you for contributing to Armory!

**Description**
Blending doesn't work in Armory with neither Deferred or Forward path, throwing a setPipeline error in Deferred:

![Error](https://user-images.githubusercontent.com/5429817/76711471-b4e0f200-6710-11ea-923f-230b4fc20c27.jpg)

In Forward I get no result:

![Blending_Empty](https://user-images.githubusercontent.com/5429817/76711687-5157c400-6712-11ea-8bde-5a94f748ed7e.jpg)

I tried to look back into earlier releases, and it seems the problem goes back for quite some time, but I managed to get it working in a version from oktober (10-2019) by setting the target to "tex" right before the blend-drawmesh function in the RenderPathDeferred file 

`path.setTarget("tex");`

https://github.com/armory3d/armory/blob/252e0dc7f3ab0ba7db0af0d3c212b6bdbbb25ffb/Sources/armory/renderpath/RenderPathDeferred.hx#L649

![Blending](https://user-images.githubusercontent.com/5429817/76711624-f756fe80-6711-11ea-8fc6-226813e5a274.jpg)

Although this doesn't seem to be a solution in the current version of Armory - While I don't get the setPipeline error, I get the same empty result as the Forward path

**To Reproduce**
Activating material blending, or turning on Blending state under Armory Render path > Renderer > Blending

**Expected behavior**
Expected result above with Blending working

**System**
Blender: 2.82
Armory: 2020-3
OS: Windows 10, 64-Bit
Graphics card: Nvidia 1060 GTX

Error also occurs for example file located under the translucent folder: https://github.com/armory3d/armory_examples/blob/master/material_translucent/blend.blend

PR #2114, opened 2021-02-20
Blending fix + ignore irradiance per material

Naxela:
I thought I had made a pull request about this earlier, but turns out I had forgotten to.

In any case, it's a fix for https://github.com/armory3d/armory/issues/1615 which involves issues with blending. On deferred blending causes a setPipeline error, whilst forward displays nothing.

Before on forward:

Forward

After:

Fixed

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -21,6 +21,7 @@
 uniform sampler2D gbufferD;
 uniform sampler2D gbuffer0;
 uniform sampler2D gbuffer1;
+uniform sampler2D gbuffer2;
 
 #ifdef _VoxelAOvar
 uniform sampler3D voxels;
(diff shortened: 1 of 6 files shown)


Naxela:
Additional commit:

- Add option to ignore irradiance per material

For baked environments, the option to ignore irradiance is nice to have, since when SH irradiance is applied to baked objects the static environments end up looking a bit too saturated and contrasted with wrong lighting:

image

I'm unsure of whether the implementation is good though, since it's using the available space in gbuffer2 (if I understood it correctly, object motion blur uses the rg values, the unused b value is then used for designated which object to be ignored by the environment lighting). I'm not sure if there's some overhead when using the existing buffer or if it would be better to create a separate rendertarget for this, since it's not available for forward rendering either (similar to object motion blur).

With static environments + irradiance applied - Notice the wrong darkening on the walls:

image

and without it looks more natural. This is more apparent in high contrast scenes:

image

In short, the point was to differentiate between static environments with lightmaps and mobile/dynamic objects inside the environment which has the SH coefficients from the interior rendered out and applied on runtime.

luboslenco:
Looks great, thanks!

PR #2115, opened 2021-02-22
Update custom trait icons

tong:
Monotone icons (haxe, bundle, wasm) for better integration into the blender 2.8+ ui.

Old:
screenshot_2

New:
screenshot_1

There is room for further improvement but I haven't found any documentation/specification for sizes, colors, ….
Though, already an improvement, i think.


MoritzBrueckner:
Nice! A few remarks:

- The down-scaled haxe and wasm icons now look a bit blurry, especially the WA in the wasm icon is unreadable. This could be because of the 10px border which makes even vertical/horizontal lines no longer pixel perfect when scaled down?

- I'm not sure if the Armory icon is really better suited for bundled traits. Of course, it makes sense, but for me it is a bit less verbose than the actual "package" icon. But that's personal opinion, both icons look good :)

As for the icon sizes, I think they are scaled to 16 or 32 px ([source](https://devtalk.blender.org/t/icon-development-guide-lines/289/2)). But Blender seems to be ok with larger sizes as well.

Maybe we should color *all* trait icons a bit in the future (like for the properties panel sidebar icons, not only Haxe and wasm as before to make it consistent) to look more appealing and distinguishable, even if it might not be that blender-ish or modern looking. For lists with many traits, colors could help you finding a particular trait faster. Again, this is just opinion-based.

tong:
Indeed, they look better when downscaled to 32px.
The wasm icon looks way too bulky without any border (imho), but the 'WA' is readable now (on my screen).
For the bundle icon i am not sure (/)

32px:
screenshot_3

PR #2124, opened 2021-03-10
Fix for material NoneType and fix for Blender exit error

Naxela:
First commit is a fix where unallocated material slots would cause an error with the arm_ignore_irradiance check, as reported by a user on Discord:

image0

image

Second commit is a fix for the annoying winman error - It's already fixed in the upcoming lightmapper update for the 2.9 branch, but I'm not sure how far out the branch is yet, so I've fixed it for the main branch as well:

image


luboslenco:
Thanks!

PR #2125, opened 2021-03-13
Fix light visual artifact when close to the camera

N8n5h:
This fixes a current problem with lights that occurs when getting too close to the camera. The issue is related to the clustering algorithm and the "cluster near solution".

> #### "Cluster near solution" explanation for context:
> Because of the logarithmic nature of how the Z slice is calculated, clusters slices get "burned" too quickly at the start of the frustum, which is not ideal. So to solve this an offset is applied to the start of the cluster frustum, which is clusterNear.

Cluster near solves one problem but creates another one: There is a gap between the near plane and the cluster near offset that gets completely ignored when calculating the Z slice. And this results in the visual artifact of a visual space not being affected by light whatsoever.

image

The proposed solution is to "stretch" the first slice of Z clusters to contain the gap.

### Before:

image

### After:

image

### DIFF
--- a/Shaders/std/clusters.glsl
+++ b/Shaders/std/clusters.glsl
@@ -8,6 +8,10 @@ int getClusterI(vec2 tc, float viewz, vec2 cameraPlane) {
 		float z = log(viewz - cnear + 1.0) / log(cameraPlane.y - cnear + 1.0);
 		sliceZ = int(z * (clusterSlices.z - 1)) + 1;
 	}
+	// address gap between near plane and cluster near offset
+	else if (viewz >= cameraPlane.x) {
+		sliceZ = 1;
+	}
 	return int(tc.x * clusterSlices.x) +
 		   int(int(tc.y * clusterSlices.y) * clusterSlices.x) +
 		   int(sliceZ * clusterSlices.x * clusterSlices.y);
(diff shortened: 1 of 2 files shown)


QuantumCoderQC:
Awesome! Thanks for the fix.

And a very nice illustration of the problem. 

I have a question though: When you say "make the gap resolve to the first slice of Z clusters" don't the slices then "burn out" by the time it gets a bit far from the `near` plane as stated in the cluster near solution context?

N8n5h:
> And a very nice illustration of the problem.

Thanks :+1: Took some time to make but I figure it was worth it to make it more understandable for future reference.

> I have a question though: When you say "make the gap resolve to the first slice of Z clusters" don't the slices then "burn out" by the time it gets a bit far from the near plane as stated in the cluster near solution context?

That shouldn't happen, since the original behavior is unmodified. What I basically did is I've "stretched" the first slice to so it starts from `near` by simply: "catching" the pixels with Z coordinate that are in the visible gap and setting their Z slice to be 1.

Regarding burning, they are still burned regardless, but it's not so close to the camera so the results are a bit more "acceptable". To properly solve this, there needs to be implemented a more elaborate solution like the one proposed by the Avalanche Studios dev in the Clustering presentation. Problem is the solution was really designed for big open world projects like the game they were making (just cause 3 I think), so it requires a new approach to make it more flexible for most or any kind of project like you would see in a game engine.

luboslenco:
Thanks for the great fixes!

PR #2128, opened 2021-03-15
Fix shadows not working when there are lights present with cast_shadow=false

N8n5h:
With this it's possible to combine lights that cast no shadows (cast_shadow=false/"Shadow" toggled off on Blender's UI) with shadow casting lights, otherwise previously broken because of the hardcoded nature of `ReceiveShadows` when invoking `sampleLight()`.

This was done by exposing cast_shadows to the shaders through pointLightData.

Note: 
Because there was not enough space I had to increment the requirement for lights from 2 to 3 vec4 in the lightsArray, so this solution leaves 3 unused spaces per light (for now). I've tried to change lightsArray to be an array of vec3 to avoid wasting space, but after testing I've noticed it doesn't play nice on Windows with Direct3D, so I had to revert it to vec4.

Tested with Krom Linux / Krom Windows, with point lights and spot lights with deferred and forward, 
Test file: [cast_shadow_test.blend.zip](https://github.com/armory3d/armory/files/6138318/cast_shadow_test.blend.zip)

Depends on this https://github.com/armory3d/iron/pull/117

### before (points):

before_points

### after (points) :

after_points

### before (spots):

before_spots

### after (spots):

after_spots

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -94,7 +94,7 @@ uniform vec3 eye;
 uniform vec3 eyeLook;
 
 #ifdef _Clusters
-uniform vec4 lightsArray[maxLights * 2];
+uniform vec4 lightsArray[maxLights * 3];
 	#ifdef _Spot
 	uniform vec4 lightsArraySpot[maxLights];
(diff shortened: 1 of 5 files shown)


luboslenco:
Awesome!

PR #2129, opened 2021-03-16
Fix clamp node

knowledgenude:
https://forums.armory3d.org/t/clamp-node-causes-failed-build/4523

### DIFF
--- a/Sources/armory/logicnode/ClampNode.hx
+++ b/Sources/armory/logicnode/ClampNode.hx
@@ -1,18 +1,18 @@
 package armory.logicnode;
 
+import kha.FastFloat;
+
 class ClampNode extends LogicNode {
 
 	public function new(tree: LogicTree) {
 		super(tree);
 	}
 
-	override function get(from: Int): Dynamic {
-		var value: kha.FastFloat = inputs[0].get();
-		var min: kha.FastFloat = inputs[1].get();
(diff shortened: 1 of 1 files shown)


Sources/armory/logicnode/ClampNode.hx
```diff
@@ -1,18 +1,18 @@
...
 		super(tree);
 	}
 
-	override function get(from: Int): Dynamic {
-		var value: kha.FastFloat = inputs[0].get();
-		var min: kha.FastFloat = inputs[1].get();
-		var max: kha.FastFloat = inputs[2].get();
-
-		if (value == null || min == null || max == null) return null;
+	override function get(from: Int): FastFloat {
```
SunDaw, line 11: You are changing the return type with the `override` here (I didn't even know the compiler would accept that) - is that something we should be doing? (I don't remember seeing it being done on any of the other nodes)

knowledgenude: As far i know the change will not affect in nothing regarding compatibility, because previously the inputs were limited to `FastFloats`, so anyway it would return a `FastFloat`. Of course the outputed value can't be connected to an `Int` input without a `Dynamic` return. Since the `min/max` values in the node would be set in the node itself, the values would be of `Float` type, then no problem because the returned value will never be an `Int`.

PR #2139, opened 2021-03-24
Add map range node, vector rotate node, missing math operations and blackbody node

Naxela:
- The map range node should now be fully supported
- The vector rotate node should now be fully supported
- The blackbody node has been reimplemented in GLSL (rather than being a constant), so animating the value is possible now
- The math node should now be fully supported (the remaining missing operations was added)


luboslenco:
Awesome!

PR #2149, opened 2021-04-02
[Blender 2.9] Nishita sky model & (ir)radiance support for all world shaders

MoritzBrueckner:

This PR adds an implementation of the Nishita sky model which was introduced in Blender 2.9. It was ported from https://github.com/wwwtyro/glsl-atmosphere which is licensed under the [Unlicense](https://github.com/wwwtyro/glsl-atmosphere/blob/master/LICENSE).

nishita_cycle_small

- To make it work fast, I moved the computation of the inner integral to a precalculated 2D LUT texture that only needs to be recalculated if the density parameters change. In the future we could improve things further by (optionally) rendering the sky to a lower resolution texture first, but that's not part of this PR. The LUT also makes it possible to store presets for different atmospheres.

- Also, I added support for all the node settings such as air, dust and ozone density and a sun disk with limb darkening.

- For clouds, there is now an option to (artificially) darken them based on the sun direction, this should probably be replaced in the future with a more physically accurate representation (but before the clouds would just remain bright):

  nishita_cycle_clouds_small
  (The artifacts in the clouds are caused by the gif compression)

- In addition to the new sky model, worlds are now rendered in the background to (ir)radiance maps when no other node already set those maps (environment texture node for example). This takes about ~1s per world on my quite old mid-level machine from 2015, even with very complex node setups because worlds only need one sample. Actually I wanted to implement world caching in this PR, but it turns out to be more complicated than I thought so I will postpone it to a future PR.

- There is also a new API for setting texture parameters in shaders.py based on https://github.com/armory3d/iron/pull/118.

### DIFF
--- a/Shaders/std/sky.glsl
+++ b/Shaders/std/sky.glsl
@@ -29,6 +29,8 @@
 #define nishita_mie_dir 0.76 // Aerosols anisotropy ("direction")
 #define nishita_mie_dir_sq 0.5776 // Squared aerosols anisotropy
 
+#define sun_limb_darkening_col vec3(0.397, 0.503, 0.652)
+
 /* ray-sphere intersection that assumes
  * the sphere is centered at the origin.
  * No intersection when result.x > result.y */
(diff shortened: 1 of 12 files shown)


luboslenco:
Wow!

PR #2153, opened 2021-04-08
Correctly detecting the system

onelsonic:
I was unable to compile the API documentation on windows (https://github.com/armory3d/armsdk/tree/master/api)
until editing this node.
Need to check if this does not break anything else.


MoritzBrueckner:
Are you sure that you're using the latest SDK? The first output of the node (index 0) is a string and the node has 6 outputs, so the current (now removed/changed) code actually looks totally correct. What was the error message you got when building the API docs?

PR #2154, opened 2021-04-08
Fix GetSystemName node

MoritzBrueckner:
Fixed `Unknown identifier : systemName` error as observed by @onelsonic in https://github.com/armory3d/armory/pull/2153.


onelsonic:
Hi @MoritzBrueckner, just seeing your question now I was AFK,
yes that was with the lastest SDK it returned an error "Unknown identifier : systemName" when trying to compile the API doc with Haxe (4.1.3). I see you figured it out, I was not sure if my commit would break something, I will close my PR.
Thank you

PR #2157, opened 2021-04-14
Add shadow map atlases info to the debug console

N8n5h:
This is a first step to add debug info to help understand what is going on with shadow map atlases by displaying some information like tile size, current size, max size, free tiles, etc. 

When single map is enabled

Screenshot from 2021-04-13 14:43:22

No single map: 

Screenshot from 2021-04-14 01:07:04


luboslenco:
Awesome!

PR #2160, opened 2021-04-18
Add visual representation of shadow map atlases to the debug console

N8n5h:
This was done to make it easier to understand what is going on with atlases to help users and also the maintenance of the feature.

No LOD

LOD


luboslenco:
Wow, so cool!

Issue #2189, opened 2021-05-10
[Feature request] Integer output from the merge node


ItsCubeTime:
![image](https://user-images.githubusercontent.com/20190653/117674691-2ea52c00-b1ac-11eb-989c-41f897b80247.png)

The integer output being counted from 0 and representing which input it was called from. 

A possible usecase would be for a character that you might want to be able to move forwards and backwards (and hence reuse the "move" code only multiplying the speed by -1 for reverse/moving backwards).

ItsCubeTime:
Input 0 = 0
Input 1 = 1
Input 2 = 2
etc

Like a reverse switch

ItsCubeTime:
Something else that I think could be useful just in general for the node system when it comes to naming, would be having some extra keywords for the nodes.

Like a node like this could also have the search keywords "join", "execute", "flow", "merge".

ItsCubeTime:
> > A possible usecase would be for a character that you might want to be able to move forwards and backwards (and hence reuse the "move" code only multiplying the speed by -1 for reverse/moving backwards).
> 
> Hello dear, did u already see the `Vector From Boolean` node? Just connect WASD keys on it and multiply the outputed vector by the desired velocity. Probably it needs to be normalized, i don't remember.
> 
> I suppose @MoritzBrueckner is right in his first affirmation, see that behavior (search for "merge"): https://forums.armory3d.org/t/weird-issue/4211/5
> 
> What happens is that the object is moved faster when more inputs are active.
> 
> Merge node actually can be activated multiple times as far i know, so an array of integers makes more sense. Otherwise another node should be created with the `any input` function.

Hi, the node is working like I would expect it to at the moment so the issue you linked is not something g that bothers me, I were indeed looking to execute my move node several times per tick.

For the other issue I think a separate "do only once" node might be better to give the user better control over from when they want extra executions per tick to get filtered away

MoritzBrueckner:
But this behaviour is actually the right one for having this output as the node is executed for each individual input, thus making it possible to get a unique value for the last active input. As I said, I originally confused the functionality of the node a bit.

Maybe, to lower possible confusion, it would be a good idea to keep the name `Merge` and instead add a node switch for either individual execution or per-frame execution if an input is active? Something like `Once Per Input` (default for backwards compatibility) and `Once Per Frame`?

ItsCubeTime:
[...]

PR #2210, opened 2021-05-21
Merge node: add execution mode and output for input index

MoritzBrueckner:
Closes https://github.com/armory3d/armory/issues/2189 and requires https://github.com/armory3d/armory_tools/pull/8 for the documentation.

Node screenshot

There is now a dropdown that lets you choose whether the node should activate its output for each active input individually or only once per frame for the first active input. If the first mode is selected, there is a new output that lets you access the index of the input that activated the node.

### DIFF
--- a/Sources/armory/logicnode/MergeNode.hx
+++ b/Sources/armory/logicnode/MergeNode.hx
@@ -2,11 +2,31 @@ package armory.logicnode;
 
 class MergeNode extends LogicNode {
 
+	/** Execution mode. **/
+	public var property0: String;
+
+	var lastInputIndex = -1;
+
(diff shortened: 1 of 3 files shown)


QuantumCoderQC:
hey, great job!

I was also working on the exact same thing last night. I was trying to add node replacement feature to it as well, where I got stuck ;-)

MoritzBrueckner:
Thanks, I hope I didn't invalidate too much of your work 😨 You're probably right that it needs node replacement, I didn't implement it because the default settings are still the same (no compatibility issues) but I guess it still needs an "Identity" replacement to automatically update the UI? I have to look this up every single time I work on it...^^

Issue #2186, opened 2021-05-10
[BUG] Unhandled promise rejection error during shader compilation. Causing the game from failing to build/launch.


ItsCubeTime:
```
MeasureIt: Cleaning data
Info: Saved "untitled.blend"

Armory v2021.5 ($Id: e4b950f78eb7c1ac9e65f2ad87920132e33cc61d $)
Exporting Scene
Scene exported in 10.466s
C:\Users\Oliver\Desktop\Armory test 3.0.0\build_untitled\debug\temp\Cloud_003_shadowmap.frag.hlsl(98,43-58): error X4000: variable 'dotNV' used without having been completely initialized

Compiling shader 4 of 27 (Cloud_003_shadowmap.frag.glsl) failed:
Shader compiler error.
(node:9632) UnhandledPromiseRejectionWarning: Shader compiler error.
(node:9632) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:9632) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Finished in 17.466s
ERROR: Build failed, check console

```

(This is my Blender console from when hitting F5)

Heres the blend file:
[untitled.zip](https://github.com/armory3d/armory/files/6451497/untitled.zip)
https://u.teknik.io/59JyH.blend
(Its the same file on both links)

You can find "Cloud.003" in the object shader editor.

QuantumCoderQC:
Hi, from your file, I was able to run it without any errors. I use the latest git version of the SDK without the 2.9 branch commits. Try cleaning the project once and then try. BTW, Which version of Blender are you using?

![image](https://user-images.githubusercontent.com/55564981/117657783-44a9f100-b19a-11eb-87c0-d458cf7d6296.png)

ItsCubeTime:
> Hi, from your file, I was able to run it without any errors. I use the latest git version of the SDK without the 2.9 branch commits. Try cleaning the project once and then try. BTW, Which version of Blender are you using?
> 
> ![image](https://user-images.githubusercontent.com/55564981/117657783-44a9f100-b19a-11eb-87c0-d458cf7d6296.png)

[...]

PR #2216, opened 2021-05-27
Fix Layer Weight node and multiple cycles shaders in one material

MoritzBrueckner:
This fixes https://github.com/armory3d/armory/issues/2186.

The issue was caused by the Layer Weight node which needs to access normals that were not available in the generated shadowmap shaders. Because the Layer Weight node might change the opacity of objects depending on normals and thus also should affect shadowmap rendering, the normals are now added to the shader if this is not yet the case (e.g. this is already done when displacement is enabled).

@luboslenco do you remember what the following comment (on now removed code) meant?

https://github.com/armory3d/armory/blob/138386f02a031e958daec1d0dca1ab2af1cf7f45/blender/arm/material/make_depth.py#L35-L37

How is something discarded at compile time (particularly by adding `vec3 n`)? Or is this just related to the discarding of fragments below a certain opacity threshold? If there is some fancy compile time optimization, I might need to revise my solution.

There also is [another comment](https://github.com/armory3d/armory/blob/138386f02a031e958daec1d0dca1ab2af1cf7f45/blender/arm/material/make_finalize.py#L11-L12) in `make_finalize` that I'm not sure how it is meant. I had an idea (that I yet have to test) that might make the shader export simpler and remove some code duplicates, which sounds a bit like the comment. The idea is to add a set of variables/values that a material can "require" while parsing it. Requirements can add other requirements transitively and at the end, code for all of those is added to the shader, for example a shader requires `dotNV` which also needs `n` which needs `wnormal` to exist. What do you think of this? Is this what was meant by the comment?

During fixing of the above, I found another issue that could happen if multiple cycles shaders were combined in one material. It's just a few lines of code, so I included it in this PR as well as some minor code cleanups.

### DIFF
--- a/Shaders/std/sky.glsl
+++ b/Shaders/std/sky.glsl
@@ -60,7 +60,7 @@ float random(vec2 coords) {
 
 vec3 nishita_lookupLUT(const float height, const float sunTheta) {
 	vec2 coords = vec2(
-	sqrt(height * (1 / nishita_atmo_radius)),
+		sqrt(height * (1 / nishita_atmo_radius)),
 		0.5 + 0.5 * sign(sunTheta - HALF_PI) * sqrt(abs(sunTheta * (1 / HALF_PI) - 1))
 	);
 	return textureLod(nishitaLUT, coords, 0.0).rgb;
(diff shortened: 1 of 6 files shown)


luboslenco:
Thanks!

> Discard at compile time 

I think `ved3 n` was only needed in some specific cases there, in other cases it was supposed to be removed by shader compiler.

> There also is another comment in make_finalize that I'm not sure how it is meant. I had an idea (that I yet have to test) that might make the shader export simpler and remove some code duplicates, which sounds a bit like the comment. The idea is to add a set of variables/values that a material can "require" while parsing it. Requirements can add other requirements transitively and at the end, code for all of those is added to the shader, for example a shader requires dotNV which also needs n which needs wnormal to exist. What do you think of this? Is this what was meant by the comment?

Yes, exactly! 👍 That would make it much better.

PR #2220, opened 2021-06-05
Add Select node

### LINKED ISSUE #2200: [Logic nodes] New node "select by index".
Best explained here.

https://youtu.be/lcN-hRux4cU?t=381

### LINKED ISSUE #2201: [Logic nodes] New node "Select by last called execution pin".
I hope you can excuse my charming dots representing node inputs/outputs.
image

This should work similarily to the select by index node described in #2200, except it should remember which pin was last executed and every time the "Return value" needs to be accessed it will return the cooresponding "Option" based on which of those exec pins was last called.

MoritzBrueckner:
Closes #2200, closes #2201.

This PR adds a new `Select` node with two different modes: it can either select an input value based on a given index, or it can forward an input value based on the index of a corresponding action input:

Screenshot

Unfortunately the colors of the Int and Dynamic type sockets are very very similar, maybe this should be adjusted in the future.

### DIFF
--- a/Sources/armory/logicnode/SelectNode.hx
+++ b/Sources/armory/logicnode/SelectNode.hx
@@ -1,0 +1,37 @@
+package armory.logicnode;
+
+class SelectNode extends LogicNode {
+
+	/** Execution mode. **/
+	public var property0: String;
+
+	var value: Dynamic = null;
+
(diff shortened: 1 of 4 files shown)


luboslenco:
Thanks!

PR #2221, opened 2021-06-07
Extra logic nodes

### LINKED ISSUE #2202: [Logic nodes] New node "Once per tick" or "Once per frame". Filters exec output to only execute once every frame maximum.
Quite well described by the title maybe but so there would be one input execution pin and one output execution pin. The node will remember if it has been called before at the current frame. If it has not then it will activate its output exec pin, if it has been executed once before that frame then it will simply "do nothing", essentially making sure any nodes after it only runs once every frame maximum (if theres nothing calling the "once per tick" node, ofc then the nodes after wouldnt be executed either).

QuantumCoderQC:
Introduces a new node "Once Per Frame". Fixes #2202 

image

### DIFF
--- a/Sources/armory/logicnode/OncePerFrameNode.hx
+++ b/Sources/armory/logicnode/OncePerFrameNode.hx
@@ -1,0 +1,20 @@
+package armory.logicnode;
+
+class OncePerFrameNode extends LogicNode {
+
+	var c = false;
+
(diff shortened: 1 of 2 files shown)


ItsCubeTime:
Thank you @QuantumCoderQC ! <3

PR #2232, opened 2021-06-14
Improve set material parameter nodes

QuantumCoderQC:
Requires also https://github.com/armory3d/iron/pull/128

This PR improves `set material parameter` nodes. This enables setting material parameters globally or on a per object basis. 

Also fixes issue where the value is not exported when `Armory parameter` is enabled.

Adds a new class `UniformsManager` in iron to improve setting material parameters.

Credits to @MoritzBrueckner, @ wuaieyo and @ danieljackson over at discord for helping with this.


blender/arm/logicnode/material/LN_set_material_image_param.py
```diff
@@ -1,17 +1,43 @@
 from arm.logicnode.arm_nodes import *
 
 class SetMaterialImageParamNode(ArmLogicTreeNode):
-    """TO DO."""
+    """Set an image value material parameter to the specified object. 
+    
+    @seeNode Get Scene Root
+    
+    @input Object: Object whose material parameter should change. Use `Get Scene Root` node to set paramter globally.
```
MoritzBrueckner, line 8: Great work 👍  Is "Use `Get Scene Root` node to set paramter globally" a leftover from older experiments? It looks like unchecking `Per Object` alone is sufficient to set the parameter globally. Also, there is a small typo in "paramter".

QuantumCoderQC: Thanks!

That get scene root node is added intentionally. We need a way to remove this object from the map when we want it to follow global values again. For example, consider an objects A using the same material as others.

Say we need to set A's parameter only, then we enable `per object` and set the property. This adds the object to the map. 

Now, to remove the object from this map, and revert back to global value, we disable `per object`. 

But we also need a way to set the global parameter without removing this object from the map. 

For this, we use scene root object. This way, we do not affect any objects with their own parameters, but directly modify the global parameter.

Also thanks for pointing out the typo, will make a commit for that!

Issue #2241, opened 2021-06-27
[(likely) Unintended behaviour] "Get object location" returns world space location whilst "Set object location" sets the local position (relative to the parent object)


ItsCubeTime:
Essentially this makes it really hard to use data fetched with the get object location node to later on set the location.

An ideal solution would be adding a bool input "Set relative location" & "Get relative location" which determines whether the variable returned/set is in world space or in relative space to its parent object.

ItsCubeTime:
https://github.com/armory3d/armory/blob/master/Sources/armory/logicnode/SetLocationNode.hx `object.transform.loc.setFrom(vec);`

https://github.com/armory3d/armory/blob/master/Sources/armory/logicnode/GetLocationNode.hx `return object.transform.world.getLoc();`

niacdoial:
darn it, just an hour before I managed to get my head together and make good on what I said on discord.
on the plus side, I have a MWE that exploits this?

here it is:
control position of the parent cube with up/down/left/right/shift/ctrl, and observe the child cube drift.
[badvariable.blend.gz](https://github.com/armory3d/armory/files/6721884/badvariable.blend.gz)

knowledgenude:
I've open a PR for that, in case someone want to use another object as relative, use **World Vector to Local Space** node

ItsCubeTime:
> I've open a PR for that, in case someone want to use another object as relative, use **World Vector to Local Space** node

Fantastic 👍 I will wait with closing this until its merged

knowledgenude:
If you still have problems on that feel free to reopen :+1:

knowledgenude:
@ItsCubeTime  feel free to reopen if the issue was not solved for you :+1:

PR #2242, opened 2021-06-27
Add "Parent Relative" option to location nodes

knowledgenude:
Proposal fix for https://github.com/armory3d/armory/issues/2241

* Improved docs for **World Vector to Local Space** and **Vector to Object Orientation** nodes to be used in cases where don't have parent.
* Added Subtract option for World Vector to Local Space node

### DIFF
--- a/blender/arm/logicnode/transform/LN_get_object_location.py
+++ b/blender/arm/logicnode/transform/LN_get_object_location.py
@@ -1,7 +1,14 @@
 from arm.logicnode.arm_nodes import *
 
 class GetLocationNode(ArmLogicTreeNode):
-    """Returns the current location of the given object in world coordinates."""
+    """Get the location of the given object in world coordinates.
+    
+    @input Parent Relative: If enabled, transforms the world coordinates into object parent local coordinates
+
+    @seeNode Set Object Location
+    @seeNode World Vector to Local Space
+    @seeNode Vector to Object Orientation
+    """
     bl_idname = 'LNGetLocationNode'
     bl_label = 'Get Object Location'
     arm_section = 'location'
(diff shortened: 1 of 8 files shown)


blender/arm/logicnode/transform/LN_get_object_location.py
```diff
@@ -10,5 +17,6 @@ class GetLocationNode(ArmLogicTreeNode):
     def init(self, context):
         super(GetLocationNode, self).init(context)
         self.add_input('ArmNodeSocketObject', 'Object')
+        self.add_input('NodeSocketBool', 'Parent Relative')
```
MoritzBrueckner, line 20: Hi, nice work! I think you might need to add a node replacement routine so that this input is automatically added to older nodes. The same holds for the other changed nodes.

knowledgenude: Sorry, i didn't know about that! Any example of how to do so?

MoritzBrueckner: You can probably just use this:

```python
def get_replacement_node(self, node_tree: bpy.types.NodeTree):
    if self.arm_version not in (0, 1):
        raise LookupError()
    return NodeReplacement.Identity(self)
```

This will update the node but keep all socket connections the same as there is only a new socket but nothing else has changed. Don't forget to also increment the `arm_version` attribute of the node.

You can try the changes by creating a file without the changes in this PR, then apply the changes again and then you can execute the "Replace Nodes" operator in the F3 search menu to test if the replacement works.

knowledgenude: Done! Thanks. Also fixed many typos i left and removed the subtract option i've added. Now the **World Vec To Local** node works as it should by default to avoid confusion. I didn't incremented its arm version because it is more a bug fix, but if you think i should increment then no problem :)

Also i've concluded that i could remove the input length verification from the nodes. Any case you can tell me what is better to do!

Issue #2166, opened 2021-04-22
Occlusion texture overlay the screen on skinned meshes with normal map


knowledgenude:
With normal map
<img src="https://user-images.githubusercontent.com/63247726/115779704-adb5fa00-a38e-11eb-9de6-8b0abf0e18b2.png" width=25% height=25%>

<img src="https://user-images.githubusercontent.com/63247726/115779756-bc041600-a38e-11eb-888e-5168ed5c0791.png" width=25% height=25%>

Without normal map
<img src="https://user-images.githubusercontent.com/63247726/115779713-b27aae00-a38e-11eb-9381-16d7e4271369.png" width=25% height=25%>

MoritzBrueckner:
Isn't the "with normal map" case the correct one? The occlusion texture states which pixels should not be affected by ambient light, thus darkening them. On the bottom image, this darkening seems missing. Or is the problem something else?

knowledgenude:
@MoritzBrueckner , thanks for the answer! The problem is if i put an occlusion map with a normal map together, this artifacts (probably the occlusion texture) are rendered in the screen:

![Screenshot from 2021-04-22 17-09-37](https://user-images.githubusercontent.com/63247726/115889617-79901700-a42a-11eb-97de-0d94751be71c.png)

It follows the camera, just like an screen overlay with some texture (i suppose the occlusion map texture).

If i use only the occlusion map without normal map or vice-versa, then it works. Both together is causing this issue (don't know on Windows). Maybe i can create an example file later for test in Windows, i fear for the license of these mixamo models.

N8n5h:
Could reproduce it too, so it may be Linux related... and from my tests it seems the issue goes away if you replace the `Principled BSDF` node for an `Armory PBR` node. Can you try that?

MoritzBrueckner:
I couldn't reproduce this so far on Windows with a skinned mesh, a normal map and the Armory PBR node. How did you use a occlusion texture with the Principled BSDF node?

knowledgenude:
>Could reproduce it too, so it may be openGL/Linux related... and from my tests it seems the issue goes away if you replace the Principled BSDF node for an Armory PBR node. Can you try that?

I've used Armory PBR

I don't remember what I was doing that I didn't answer that, but today i got an example of a similar issue happening:

[Test.zip](https://github.com/armory3d/armory/files/6734332/Test.zip)

What is happening is that if i use opacity in the material, then appear this artifact:
[...]

PR #2268, opened 2021-07-10
Fix visual artifacts caused by invalid gbuffer2 on OpenGL

MoritzBrueckner:
This PR fixes the second example from issue https://github.com/armory3d/armory/issues/2166 and probably also the original issue. @knowledgenude can you please test this if you have the project still around?

The reason for the artifacts was that `gbuffer2` could be accessed even if it was not used, thus it was filled with left-over values from other usages on OpenGL. On DirectX this was not an issue (probably undefined behaviour).


knowledgenude:
Many thanks for the fix, @MoritzBrueckner ! All working perfectly now :)

Issue #1693, opened 2020-05-12
Very low frame rate in Mali GPU only


knowledgenude:
Doing some tests in different Android devices, one of them with a Mali GPU (samsung smartphone) have very low frame rate in a basic scene. A same specs smartphone, with the same Android version but with Adreno GPU, runs with high fps. Is there some feature in Armory that can cause a low frame rate in Mali GPU? Can a simple logic do this? Because i already tested irradiance, textures, physics, and nothing lagged in the device. It seems that the issue is caused by some other stuff in my specific blend file. I will keep doing tests around this, but if someone can give me some light to go i will be grateful.

• Already disabled all unused stuff, like skinning, displacement, etc. and nothing gives a improvement in the performance in this case.

• The device don't over heat nor consumes too much energy (just a little bit more).

Edit: Also the HDR world background turns black in this specific device, but the irradiance works fine. Already tryed different resolutions but not get results.

knowledgenude:
Tested in more Mali GPU devices and the world keeps black. After some search i found that the problem is with Mali, but i tested Avakin (that have good graphics for a android game) and it runs good even in a weak device, probably i should downgrade more the graphics.

QuantumCoderQC:
Hi @knowledgenude,

Did you try and compare Mali GPU devices VS others using the default cube scene?
How is the frame rate then?

knowledgenude:
Hi @QuantumCoderQC, i have tested in three Mali devices until now. The low specs device have very low frame rate with a basic scene (that have 15 rigid bodies with spherical shape, 2 materials with textures, hdr world and irradiance enabled) but strangely, this device runs Avakin very well at 45/60 fps with a lot of alphas and skinning. I already tested with lamps and shadows and the problem is not this.

When i rotate the camera, the black world seems that is flicking. I see a problem like this in the GTA SA 0Android versions when it was released in Samsung's devices (related to GPU).

But the mainly focus is performance as we are talking about mobile. If you tell me a way to calculate the framerate directly in the device, i will make a lot of builds, each for a element (physics, graphics, logic, etc.) as i can't publish with debug console to see exactly what is lagging.

Also, the Sky Texture appearly works fine in Mali unlike the world hdr.

Everything runs very well in Adreno GPU, even with shadows and lamps (tested in 2 devices and i will test in more of these devices soon too). What do you think about it? Maybe the problem is not in GPU type but on changes in the OS? I particulary think that is the GPU.

QuantumCoderQC:
[...]

PR #2274, opened 2021-07-17
texture format changes in order to support iOS devices

onelsonic:
changing texture format from RGBA64 to RGBA48 so it is supported on both iOS and Android devices.

### DIFF
--- a/Sources/armory/renderpath/Inc.hx
+++ b/Sources/armory/renderpath/Inc.hx
@@ -369,7 +369,7 @@ class Inc {
 		t.width = 0;
 		t.height = 0;
 		t.displayp = getDisplayp();
-		t.format = "RGBA48";
+		t.format = "RGBA64";
 		t.scale = getSuperSampling();
 		t.depth_buffer = "main";
 		path.createRenderTarget(t);
@@ -524,7 +524,7 @@ class Inc {
 
 	public static inline function getHdrFormat(): String {
 		#if rp_hdr
-		return "RGBA48";
+		return "RGBA64";
 		#else
 		return "RGBA32";
 		#end
--- a/Sources/armory/renderpath/Inc.hx
+++ b/Sources/armory/renderpath/Inc.hx
@@ -369,7 +369,7 @@ class Inc {
 		t.width = 0;
 		t.height = 0;
 		t.displayp = getDisplayp();
-		t.format = "RGBA64";
+		t.format = "RGBA48";
 		t.scale = getSuperSampling();
 		t.depth_buffer = "main";
 		path.createRenderTarget(t);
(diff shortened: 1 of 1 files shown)


MoritzBrueckner:
Hi, there is no such format as `RGBA48` in Kha ([API docs](http://api.kha.tech/kha/graphics4/TextureFormat.html)). As you can see [here](https://github.com/armory3d/iron/blob/aeae2f4644089bead90d5ff7dedef2b75e0c046b/Sources/iron/RenderPath.hx#L641), using `"RGBA48"` will just result in `RGBA32`, which is no HDR at all (and the changes in this PR will affect all devices, not only mobiles). Instead (if there really is no HDR support on iOS) HDR should probably only be used on non-iOS devices then, you can use Kha's compiler flags for that.

knowledgenude:
I made a few tests months ago and noticed that the world background is black in devices that have Mali GPUs, such as Samsung's ones.

Maybe changing the textures from RGBA64 to RGBA32 for Android also could solve this issue, but i don't have a way to build and test it now.
 
See some screenshots in the last comment: https://github.com/armory3d/armory/issues/1693#issuecomment-628066069

onelsonic:
well not really sure what is going on really but I have the following cases,
if I change the texture format as I suggested to :
RGBA48 : 3D canvas and Bloom effect works on both iOS and Android
RGBA64 : 3D canvas and Bloom works only on Android, I only have a black screen on iOS
RGBA32 : 3D canvas works on both iOS and Android but no Bloom effect

Not sure how to explain this, maybe the shaders are picking up the specific texture format somehow?

onelsonic:
@knowledgenude 
> 
> 
> I made a few tests months ago and noticed that the world background is black in devices that have Mali GPUs, such as Samsung's ones.
> 
> Maybe changing the textures from RGBA64 to RGBA32 for Android also could solve this issue, but i don't have a way to build and test it now.
> 
> See some screenshots in the last comment: [#1693 (comment)](https://github.com/armory3d/armory/issues/1693#issuecomment-628066069)

try to desactivate the HDR setting and check if your texture is rendering.

MoritzBrueckner:
> So what we need is a way to have the option to pass the same conditions as the RGBA64 textures format (with post processing/HDR possible) but set the texture format as RGBA32 so we can have all working on iOS as well.

Yes, that's what I meant with "Kha's compiler flags", you can include code depending on some compiler defines (like in C for example). It's explained [here](https://github.com/Kode/Kha/wiki/System-defines) in greater detail.

This way you could write something like `#if (rp_hdr && !(kha_android || kha_ios))` for the two lines you changed in this PR. I didn't test it but that could work. Then `RGBA64` is only used if hdr is active and you're not on mobile devices.

onelsonic:
...I cannot delete the commits in this pull request I will close it and submit another one.

Issue #2279, opened 2021-07-20
mutilple UV's makes project crash on mobile


onelsonic:
Some realtime shader effects needs multiple UV's.
Multiple UV's works fine on desktop but not on mobile, the project will not even compile.
Is there a way to activate multiple UV's on mobile on the mobile materials as well?

It uses the shading node "UV Map": 
https://docs.blender.org/manual/en/latest/render/shader_nodes/input/uv_map.html
![render_shader-nodes_input_uv-map_node](https://user-images.githubusercontent.com/53947594/126352014-dcfd7508-b7a7-41fd-a944-961dbfb79e8b.png)

>***On Desktop you will get multiple UV's working fine :***
<img width="303" alt="multiple UV's working" src="https://user-images.githubusercontent.com/53947594/126346983-136ac220-bd8a-4cce-a916-41dbc21cc5dc.png">

>***On Mobile it will crash with the following:***
```
Armory v2021.7 ($Id: 5533755fad0369c580ed2760e777e6d88341d0fb $)
Armory: Using CPU for environment render (might be slow). Enable CUDA if possible.
Exporting Scene
Scene exported in 0.280s
ERROR: C:\multi-UV-suzannebravehearts\build_multi-UV-suzannebraveheart\compiled\Shaders\left_mesh.frag.glsl:20: 'texCoord1' : undeclared identifier
ERROR: C:\multi-UV-suzannebravehearts\build_multi-UV-suzannebraveheart\compiled\Shaders\left_mesh.frag.glsl:20: '' : compilation terminated
ERROR: 2 compilation errors.  No code generated.


SPIR-V is not generated for failed compile or link

Compiling shader 1 of 8 (left_mesh.frag.glsl) failed:
Shader compiler error.
(node:3276) UnhandledPromiseRejectionWarning: Shader compiler error.
(node:3276) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:3276) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Finished in 1.967s
ERROR: Build failed, check console
```

Here is a Blend file: 
https://github.com/onelsonic/armory_tests/blob/master/multi-UV-suzannebraveheart/multi-UV-suzannebraveheart.blend?raw=true

PR #2280, opened 2021-07-20
Fix usage of multiple UV maps on mobile render path

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/2279. The necessary code for exporting `texCoord1` was simply missing on mobile.


onelsonic:
Great, I see other things differing for correct effects on the mobile target.
I will do some more tests.

PR #2283, opened 2021-07-25
Live patch: refactor and logic nodes support

MoritzBrueckner:
_TL;DR: here is a video that showcases some of the changes: https://www.youtube.com/watch?v=mxklu7TfwBw_

This PR improves the live patch system:

- Support for light color and strength settings, more could be implemented easily by adapting the current code.
- Many smaller changes to improve the overall stability
- Added a list of operators that shouldn't trigger a rebuild
- Support for live patching logic nodes:
  - Nodes now have their own property constructors for properties with a Haxe equivalent, e.g. `IntProperty` becomes `HaxeIntProperty` with an additional first parameter equal to the property's name. This was required to be able to listen to property changes, and in the future we could name properties different than `property0` - `property9` with only some minor adjustments needed in the code.
    Old node trees automatically get updated when a file saved before SDK 2021.08 is opened. External node libraries need to update this to support live patch, but outside of live patching the nodes will continue to work without getting updated.
  - Added custom socket types even for standard Blender types to be able to listen to socket value change events. This is also updated automatically for old files and external libraries should update this.
  - Supported operations:
    - Adding/removing links
    - Adding nodes
    - Duplicating nodes
    - Deleting nodes
    - Change properties and socket default values
    - Operators that change the amount of sockets
  - At the cost of slightly more memory, the way links between logic nodes are stored has changed to reduce complexity in the algorithms used for live patch updates (and other areas like `node.runOutput()` benefit as well).

**Limitations:**
- The current way of sending live patch data to Krom is to write it to a file called `krom.patch` that is opened by the running game in regular time steps. This means that we can't reliably send multiple updates directly after each other in a short time span because some of them would simply go missing. As a result there are some limitations:

  - When dragging a link between nodes to another target (no deletion and then adding a new link, but re-target a link), only one related node is updated, which node is more or less randomly.
  - If multiple nodes are deleted at once with box select, only one of them is deleted in the running game.
  - If the game is re-exported because some changes unknown to the live patch system, the changed node setup is reset to the version in krom.js until the game is properly restarted. It would be possible to recreate the entire tree but not with the current file-based IPC for said reasons, there would be too many commands in a short time span.

  I tried to implement a named pipe as a node.js extern but unfortunately didn't succeed, maybe something like this could get implemented in Armorcore directly?

- Also, there is no live patch support yet for reroutes, support for that will follow in a future PR (shouldn't be much work but I will have almost no time in August).

So using live patch for logic nodes remains rather experimental at the moment, it's likely that there might be some errors left as well although I thoroughly tested the changes.


tong:
Why not use a network socket for sending data ?

MoritzBrueckner:
Yeah that would probably work as well. I have no experience whatsoever in network programming, so feel free to implement it if you want. The data is sent [here](https://github.com/MoritzBrueckner/armory/blob/8a758bbe267c13ecc686dcb05fda4505026a644c/blender/arm/live_patch.py#L92-L98) and received [here](https://github.com/MoritzBrueckner/armory/blob/live-patch/Sources/armory/trait/internal/LivePatch.hx#L24-L33).

tong:
Just checked and there seems to be no support for sockets in krom :(

MoritzBrueckner:
Can [this](https://api.haxe.org/js/html/WebSocket.html) be used on the Krom target?

tong:
I don't think so.
But this would allow to have live-patch functionality for the html5 target.

luboslenco:
Amazing. Would be cool to eventually get sockets in Krom.

PR #2304, opened 2021-08-17
Shift click play just run

tong:
Adds _shift_ option to the _Play_ button to just run last build (without compiling).


luboslenco:
Thanks!

PR #2305, opened 2021-08-19
Added a rotation node-socket

niacdoial:
consequences:
- revamped all nodes that work with rotations
- removed the notion of a quaternion socket (even though it was a vector socket by lender standards, which made it especially awkward)
- rotation is now internally represented as a quaternion, not a XZY-ordered euler, which allows better performance in almost all cases

Some checking was done, both for haxe-time bugs, and for node-upgrade-time bugs. Hopefully this testing was complete.

NOTE: this pull request requires changes in iron, in the form of one file added under path `iron/Sources/iron/math/Rotation.hx`. The file is (hopefully) attached below.


niacdoial:
here: [Rotation.hx.gz](https://github.com/armory3d/armory/files/7012917/Rotation.hx.gz)

PR #2309, opened 2021-08-26
Add assert() macro

MoritzBrueckner:
This PR adds an `assert()` macro to Armory in order to make the handling of some error cases easier and optional for better performance. This is especially helpful for checking correct user input in logic nodes for example, but this PR doesn't introduce any code yet that actually uses the macro. Assertions are _completely_ turned off and omitted from the code when publishing a build.

Please note that assertions don't replace checks that are required during runtime! For example they should not replace important safety checks in some nodes where the input is allowed to be invalid during runtime (e.g. unavoidable `null` values).

For debugging, there are two different assert levels available: `Warning` (only shows a message but doesn't throw an exception) and `Error` (throws an exception). Error assertions are also active when the warning level is set by the user. The user can also decide whether to ignore assertions completely even in debug builds and whether the game should quit if an exception happens (helpful when in fullscreen for example). I'm not sure if there is a need for more levels, but they can be easily added later.

### **Usage**
`assert()` is imported automatically in the armory directory but has to be imported for user scripts.

```haxe
// If the second "condition" parameter evaluates to false, the assertion fails
// The last message parameter is optional
assert(Warning, 1 < 0, "Test assertion");

// Leads to

// Sources/armory/logicnode/PrintNode.hx:12: Failed assertion:
//        Message: Test assertion
//        Expression: (1 < 0)
```

```haxe
assert(Error, 1 < 0, "Test assertion");

// Leads to

// Trace: Error:
// Failed assertion:
//         Message: Test assertion
//         Expression: (1 < 0)
//     at Function.throwAssertionError (<anonymous>:24213:9)
//     at armory_logicnode_PrintNode.run (<anonymous>:16643:24)
//     at armory_logicnode_MergedKeyboardNode.runOutput (<anonymous>:886:19)
//     at armory_logicnode_MergedKeyboardNode.update (<anonymous>:14860:9)
//     at update (<anonymous>:24822:28)
//     at kha_TimeTask.task (<anonymous>:51806:4)
//     at Function.executeTimeTasks (<anonymous>:51689:76)
//     at Function.executeFrame (<anonymous>:51639:18)
//     at renderCallback (<anonymous>:52540:17)
```

Unfortunately, the Blender UI is kind of ugly, but I didn't find a way to move the dropdown label above the dropdown while keeping it aligned with the column layout:

Screenshot of the new assertion settings in the Blender UI

### DIFF
--- a/Sources/armory/import.hx
+++ b/Sources/armory/import.hx
@@ -1,0 +1,1 @@
+import armory.system.Assert.*;
--- a/Sources/armory/system/Assert.hx
+++ b/Sources/armory/system/Assert.hx
@@ -12,8 +12,8 @@ class Assert {
 
 	/**
 		Checks whether the given expression evaluates to true. If this is not
-		the case, an `ArmAssertionException` with additional information is
-		thrown.
+		the case, an `ArmAssertionException` is thrown or a warning is printed
+		(depending on the assertion level).
 
 		The assert level describes the severity of the assertion. If the
 		severity is lower than the level stored in the `arm_assert_level` flag,
--- a/Sources/armory/system/Assert.hx
+++ b/Sources/armory/system/Assert.hx
@@ -1,0 +1,131 @@
+package armory.system;
+
+import haxe.Exception;
+import haxe.PosInfos;
+import haxe.exceptions.PosException;
+import haxe.macro.Context;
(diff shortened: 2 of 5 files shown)


luboslenco:
Thanks!

Issue #2314, opened 2021-09-01
Bloom exemple issues on Android : Kinc: Uniform PPComp10 / PPComp11 not found.


onelsonic:
Trying to compile the Bloom example to Android.
I get the following error with a black screen, then the App crashing :
```
W/Kinc: Uniform PPComp10 not found.
W/Kinc: Uniform PPComp11 not found.
```

https://github.com/armory3d/armory_examples/tree/master/render_bloom

knowledgenude:
I don't think bloom is supposed to work in Android, but did you tried any other postprocess filter? IRC if you check the **postprocessing checkbox** in the **renderpath settings** is sufficient to get a black screen in Android (even with everything turned "off"). Maybe that is the main cause of the issue. I would test it again by myself but Android Studio doesn't likes me...

MoritzBrueckner:
Fixed in https://github.com/armory3d/armory/pull/2315.

PR #2315, opened 2021-09-02
Fix "Uniform PPCompXX not found" warnings

MoritzBrueckner:
Potential fix for https://github.com/armory3d/armory/issues/2314.

When `Realtime postprocess` was turned off, some unused uniforms were removed by the compiler but were still referenced from Iron. This led to warnings and maybe even to the blackscreen reported in the above issue. @onelsonic can you please check whether the blackscreen is fixed by this PR?


onelsonic:
I confirm that the black screen and crash is now corrected on Android. Thanks 👍

Issue #2323, opened 2021-09-05
Realtime postprocess not working


tong:
Since #2315 realtime postprocessing (`rp_pp`) stopped working for me for all uniforms wrapped with `#ifdef _CPostprocess` condition. Not sure why this is happening, `_CPostprocess` is correctly defined when `rp_pp` is enabled.

onelsonic:
Try rebuild a project from an empty Blender file (default cube).
Not sure if it is the case but sometimes it seems like the ArmoryUI properties that were setup are kept in some cached info within the Blender file itself.

tong:
Doesn't help.

Reproduce:
- Create empty project
- Enable realtime postprocessing
- Enable Chromatic Aberration
- Set Chromatic Aberration type to Spectral

Results in black screen.

MoritzBrueckner:
Can confirm this. I think this is due to the fact that defines seem to be added to the `compiled.inc` include that some shaders include ([source](https://github.com/armory3d/armory/blob/55cfead0844923d419d75bf4bd677ebed714b4b5/blender/arm/write_data.py#L543-L546)), but not all shaders include that file. Now the question is whether that should be the case, or if those defines should be put in another defines.inc for example that is always included, or if Kha is able to compile the shaders correctly when the flag is also set in Khamake.

PR #2332, opened 2021-09-07
Fix realtime postprocess & cleanup arm.lib.make_datas

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/2323. This issue happened because `PPCompXX` uniforms were used in shaders which were not part of the compositor pass, however they were only correctly considered in that pass. Because the uniforms don't interfere with any non-compositor uniforms, it is safe to simply allow them for non-compositor shader passes.

I also cleaned up make_datas.py a bit, the changes are mostly semi-automatic IDE stuff, some type hints and comments and two big duplicate code blocks were moved into a function.

### DIFF
--- a/blender/arm/lib/make_datas.py
+++ b/blender/arm/lib/make_datas.py
@@ -1,14 +1,15 @@
-import os
 import arm.utils
 import arm.assets as assets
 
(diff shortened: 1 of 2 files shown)


tong:
Works for me.

PR #2338, opened 2021-09-15
Remove haxe --times param per default

tong:
How about removing this flag per default ?
There isn't that much useful information in there and one could add it manually to the khafile.js if necessary.


MoritzBrueckner:
What about adding it back as an option in the Armory developer settings? I guess it could be helpful to debug macro times, especially if the new assert macro should get used more.

tong:
Sounds good too. @luboslenco opinions ?

luboslenco:
Sounds good as an option in developer settings.

tong:
Added.
armsdk change is here: https://github.com/armory3d/armsdk/pull/35

PR #2340, opened 2021-09-17
Vscode open error position

tong:
This change opens vscode|kodestudio at error position (line-column) if haxe compilation fails, but only if `open_script_error` is set in prefs.

It's only tested on linux and vscode, but i assume kodestudio has the same cli.

The change in armsdk: https://github.com/armory3d/armsdk/pull/34

### DIFF
--- a/blender/arm/make.py
+++ b/blender/arm/make.py
@@ -73,7 +73,7 @@ def run_proc(cmd, done: Callable) -> subprocess.Popen:
         else:
             done()
 
-    p = subprocess.Popen(cmd)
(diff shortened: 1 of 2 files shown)


tong:
Just recognized that there are issues with the subprocess stdout pipe if compilation server is not enabled.
…

tong:
I close this for now and come back to it later.

From what i see we should integrate the haxe compilation server via writing/reading to and from stdin/stdout and not using the network stack, which would resolve the difference of using haxe-server/no-haxe-server and therefore also resolve the issue i am having here.

PR #2359, opened 2021-10-11
add logic nodes: set/get canvas font size and set progress bar color

t3du:
image


luboslenco:
Thanks for contributing!

PR #2367, opened 2021-10-25
Fix exporting textures from linked libraries

MoritzBrueckner:
Fixes an issue that could happen if a material from a linked library used a texture with a path relative to the library. In this case the relative path would be copied to the khafile and Iron would then try to load a nonexisting texture.

@luboslenco Currently `krom_load_image()` will return `null` when an image doesn't exist which will result in a Kha image object which neither has a `texture_` nor a `renderTarget_` set. Because of that, Amorcore will crash as soon as Iron tries to generate mipmaps for the image. Should there be handling for this in the mipmap generating function? Should `krom_load_image()` instead trigger a failure callback so that those failures can be more easily handled in Iron? I'm not sure what the "design philosophy" of Armorcore or Iron is in this case.

### DIFF
--- a/blender/arm/material/cycles.py
+++ b/blender/arm/material/cycles.py
@@ -737,7 +737,9 @@ def node_name(s: str) -> str:
 
 ##
 
+
 def make_texture(image_node: bpy.types.ShaderNodeTexImage, tex_name: str, matname: str = None) -> Optional[Dict[str, Any]]:
+    """Creates a texture reference for the export data for a given texture node."""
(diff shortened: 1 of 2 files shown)


luboslenco:
Thanks! At https://github.com/Kode/Kha/blob/master/Backends/Krom/kha/LoaderImpl.hx#L14 `loadImageFromDescription` should call `failed` instead of `done` in that case, which could be enough to at least prevent the crash for now?

Issue #225, opened 2017-08-17
Shape keys


Xristoskr:
shape key  didnt work in armory
*(optional - drag & drop zipped .blend file here)*

ItsCubeTime:
Honestly this is a requirement for any game engine. You cant use bones for everything and its far from ideal.

Skehmatics:
This has been a long requested item, so I'm opening a bounty ($75 to start):

https://www.bountysource.com/issues/48360495-shape-keys

QuantumCoderQC:
Hi!

Yes, I had just started working on this. I might have a basic shapekey support sometime early next week. Then required extra features can be added as we go.

MoritzBrueckner:
Closing this issue as shape keys are now implemented in https://github.com/armory3d/armory/pull/2375 by the awesome @QuantumCoderQC!

PR #2375, opened 2021-11-02
Implementing Shape keys

QuantumCoderQC:
This PR implements GPU and texture based shape keys, more commonly known as "morph targets" in game-dev terminology. 
Solves  https://github.com/armory3d/armory/issues/225

Requires https://github.com/armory3d/iron/pull/145

**What is possible:**
- Supports a maximum of 32 shape keys per object.
- Weights of each of the shape keys can be individually set. The resulting mesh is a sum of all shape keys, just like in Blender.
- Supports displacement and normal maps.
- Shape keys can be used along with skinned mesh too.
- Any shape key can be excluded form export by "muting" that shape key in Blender.
- Shape keys can be enabled or disabled for the entire project in Armory render path settings.
- Added a new `Set Object Shape Key` Node to make it easier to set shape keys.
- The node also supports live patching

**How it works:**
- At export, if a mesh has shape keys, displacement of each vertex in a shape key is calculated relative to the `Base` mesh. Vertex normals are also calculated relative to the ´Base` mesh.
- This displacement and normal data is scaled to best possible range between 0 to 1 and stored in RGBA32  format
- The pixels are then packed into best suited power-of-two png image in a dedicated directory `MorphTargets`
- At export the mesh containing shape keys will add an additional UV map `UVMap_shape_key` to read this texture. This does not conflict the already supported maximum of 2 UV maps.
- This image is read at runtime as a texture in a dedicated morph target shader in the vertex shader.

**Missing Features and known issues:**
- Tangent maps are not yet exported. So displacement along the tangents do not work too well.
- Shape key animations are not yet implemented. Only relative shape keys are supported.
- Maximum image size for shape key texture is hard coded to be 4096 X 4096 pixels.

Thanks to people at Discord for providing valuable inputs and testing the code.


luboslenco:
Wow, amazing!

PR #2382, opened 2021-11-07
Fix morph target exporting on Unix-based systems

Skehmatics:
The new morph-target exporting had explicit backwards-slash directory separators, and while that works fine on Windows, we need to use forward-slash separators for the Unix-based systems (Linux, MacOS). I also received an issue due to some parameters not being ints in the NLA code, not sure if this is platform dependent or not. 

To fix this, this PR uses `os.path.join` to automatically join the two with the correct separators as well as casts the problem paramaters to int. This also cleans up some extra spaces lying around in the file since my editor automatically removes them.

(Also, a million thanks to @QuantumCoderQC for this feature!!)

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -985,7 +985,7 @@ class ArmoryExporter:
                             for _o in sel:
                                 _o.select_set(False)
                             skelobj.select_set(True)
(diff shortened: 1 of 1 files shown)


luboslenco:
Thanks!

PR #2410, opened 2021-12-23
Improve node library handling

MoritzBrueckner:
- External node libraries can now use Armory's `init_nodes()` function that registers all nodes inside of a given package (recursively)
- If node categories had special characters in their name (e.g. spaces), the menu class name and bl_idname would also contain those characters which would lead to warnings (interestingly, there was no _error_ when using spaces before, just a warning...?)
- Also fixed a small typo that is unrelated to this PR, but I didn't want to annoy you with a separate PR for this. @luboslenco it seems that I have write access to this repo, I guess it's ok for you if I would commit such trivial typo fixes directly to master? For everything else I will of course open a PR, but I think stuff like this is just unnecessary work for you.

### DIFF
--- a/blender/arm/logicnode/__init__.py
+++ b/blender/arm/logicnode/__init__.py
@@ -64,17 +64,25 @@ def init_categories():
     arm_nodes.add_category_section('default')
 
 
-def init_nodes():
-    """Calls the on_register() method on all logic nodes in order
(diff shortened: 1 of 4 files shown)


luboslenco:
> I guess it's ok for you if I would commit such trivial typo fixes directly to master?

Perfectly ok. 🙂

PR #2411, opened 2021-12-28
OnCanvasElementNode: fix elements values for scaled canvas

t3du:
this node wasnt working as expected if the canvas was scaled with the set canvas global scale factor... the values considered the initial canvas values defined in the json file instead of the actual ones modified by scale.


luboslenco:
Thanks!

PR #2416, opened 2022-01-04
add and remove objects from collections

t3du:
improve the create collection node by adding object arrays when creating a collection

add and remove object from a collection

image

image

requires: https://github.com/armory3d/iron/pull/151


t3du:
updated with getGroup instead of add and remove functions from iron in a simpler way, thanks!

luboslenco:
Thanks a lot!

PR #2418, opened 2022-01-08
Math node - add precision for round

t3du:
image

### DIFF
--- a/blender/arm/logicnode/math/LN_math.py
+++ b/blender/arm/logicnode/math/LN_math.py
@@ -4,7 +4,7 @@ class MathNode(ArmLogicTreeNode):
     """Mathematical operations on values."""
     bl_idname = 'LNMathNode'
     bl_label = 'Math'
-    arm_version = 1
+    arm_version = 2
 
     @staticmethod
     def get_enum_id_value(obj, prop_name, value):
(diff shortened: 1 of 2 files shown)


blender/arm/logicnode/math/LN_math.py
```diff
@@ -63,6 +63,9 @@ def set_enum(self, value):
                 while (len(self.inputs) > 1):
                     self.inputs.remove(self.inputs.values()[-1])
         self['property0'] = value
+        if (self.property0 == 'Round'):
+            self.inputs[1].name = 'Precision'
+        else: self.inputs[1].name = 'Value 1'
```
MoritzBrueckner, line 68: Hi, this will fail if you switch from "Round" to some operation that has only one input, e.g. "Sine". Better change the `else` to `elif len(self.inputs) > 1`. Did not see this while we were talking about it on Discord, sorry.

MoritzBrueckner (changes requested):

blender/arm/logicnode/math/LN_math.py
t3du: you are right i did not see the error in the console log... i just see it working as expected when i tested all the cases,

thanks im gonna make the change

MoritzBrueckner (approved):

Issue #2434, opened 2022-02-03
math node didn't work fine with v 2022.2.0


mhreez:
Thank you for contributing to Armory!

**Description**
math node didn't work fine

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**
What you expected to happen.

**System**
Blender:2.93.1
Armory:2022.2.0
OS:
Graphics card:1050

**Test File**
[test.zip](https://github.com/armory3d/armory/files/7990947/test.zip)

*(drag & drop the zipped .blend file here)*

MoritzBrueckner:
The problem seems to be caused indirectly by the "GET storage" nodes. If these nodes exist, the integer node will be replaced with a dynamic node during export which stores an empty object `{}` upon initialization:

https://github.com/armory3d/armory/blob/68fb7b114766e6d5586f46910c12e9d704994af1/Sources/armory/logicnode/DynamicNode.hx#L9

Then the math node continues to calculate with that value and because you're running the game on a dynamic target (Krom), there is no error but instead javascript starts to append the numbers to the object as if it was a string...

This particular issue is probably fixable by correctly initializing the dynamic nodes in case there is a getter/setter node somewhere, however I don't know how to handle the input socket for the variable nodes. Are they even used in general?

For now you can just replace the "GET storage" nodes with another integer node that shares the same ID (sidebar > Armory > Armory Logic Node > ID) as the other integer node. Then both nodes are actually only one node during runtime.

mhreez:
`however I don't know how to handle the input socket for the variable nodes. Are they even used in general?`

what did you mean ?

`For now you can just replace the "GET storage" nodes with another integer node that shares the same ID (sidebar > Armory > Armory Logic Node > ID) as the other integer node. Then both nodes are actually only one node during runtime.`

I tried it it works but the storage system still doesn't work

MoritzBrueckner:
> what did you mean ?

Node inputs are followed at runtime, not during compilation. But variable nodes that are linked together are handled at compile time, which makes this a bit complicated. But I'm currently working on a fix (or better: on a re-design of the handling of linked variable nodes) that will solve this, because linked variable nodes will simply show no input sockets anymore. It also won't be possible anymore to use differently typed nodes for the same ID, so there won't be any race conditions anymore that cause incorrect behaviour.

I guess the PR for this will be ready in the next days, here is a little sneak peek (WIP):
[...]

PR #2439, opened 2022-02-21
New logic tree variable system

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/2434.

What started as an attempt to fix a simple issue became an almost complete rewrite of the logic ID system. Previously it was easily possible to create a few different situations where the correctness of the exported logic tree was completely violated or at least depending on external things such as iteration orders when exporting nodes:
- if multiple variable nodes shared the same values, only one random node was exported, so the initial value was undefined if the nodes had different values.
- it was possible to use the same logic ID for nodes with different types. Because all linked nodes were replaced with the exported node of another type, this could lead to runtime exceptions/type errors in other places of the tree.
- the above also happened with the dynamic getter/setter nodes, which caused the linked issue.

New Tree Variables panel

Now, the logic ID is no longer directly accessible, instead the user has to promote variable nodes to tree variable nodes to let them share data. So if a variable node is linked to a tree variable, it shares its values with all other variable nodes that are linked to the same tree variable. A variable node does not have to be linked to a tree variable, in this case it acts for its own, just as before if the logic ID was unset.

**In one sentence: if the variable node has no logic ID, all stays as it was before. If it has one set, the logic ID acts as a reference for a _tree_ variable.**

Files that are saved before SDK 2022.03 will be automatically upgraded to the new system (assuming that the version number of the SDK in the code will be bumped for the release), apart from the fact that the initial value may be affected by said race conditions.

## Technical explanation of the new system:
 A tree variable stores the type of the variable node to ensure correctness, and it uses one linked node (internally called the "master node") to actually store the variable values. This node is also used to show the variable values in the UI in the side panel, and the master node automatically synchronizes its values with all other nodes that are linked to the same variable (called "replica nodes" in the code). The user doesn't need to know about this concept, it is all done in the background and it even works to copy tree variable nodes between trees, there is always only one master node for each tree var, and for each variable node with a logic ID in the tree there will be a tree variable automatically. If there is no linked node left for a tree variable, the tree variable is removed.

In case this short paragraph is too briefly explained, there is a more detailed explanation in arm_nodes.py :)

## User perspective:
The following operators are relevant for the user:
- **New Var From Node**: creates a new tree variable for the selected variable node. This operator is not visible in the screenshot above, as it shares its position based on the selected node with the next operator, namely...
- ...**Make Node Local**: unlinks the selected variable node from the selected tree variable. If the selected node was the last to link to the tree variable, the tree variable is removed.
- **Assign To Node**: link the selected variable node to the selected tree variable.
- **Add Getter**/**Add Setter**: as before, just that the getter/setter nodes are no longer dynamic but typed the same as the variable.

The tree variables have a color that is automatically synchronized with their linked nodes, and the name of the tree variable is shown in the node's body and header instead of the input sockets and other settings (which are now only editable in the sidebar if the node is linked, because the master node needs to be edited):

Linked tree variable node

The name of a tree variable can be edited by double clicking on it in the sidebar list.

Thanks a lot to @QuantumCoderQC for the helpful discussions and ideas on Discord!

### DIFF
--- a/blender/arm/handlers.py
+++ b/blender/arm/handlers.py
@@ -106,9 +106,9 @@ def send_operator(op):
 
 def always() -> float:
     # Force ui redraw
-    if state.redraw_ui and context_screen is not None:
-        for area in context_screen.areas:
-            if area.type == 'VIEW_3D' or area.type == 'PROPERTIES':
+    if state.redraw_ui:
+        for area in bpy.context.screen.areas:
+            if area.type in ('NODE_EDITOR', 'PROPERTIES', 'VIEW_3D'):
                 area.tag_redraw()
         state.redraw_ui = False
     # TODO: depsgraph.updates only triggers material trees
--- a/blender/arm/logicnode/__init__.py
+++ b/blender/arm/logicnode/__init__.py
@@ -90,7 +90,7 @@ def init_nodes(base_path=__path__, base_package=__package__, subpackages_only=Fa
                 _module = importlib.reload(sys.modules[module_name])
 
             for name, obj in inspect.getmembers(_module, inspect.isclass):
-                if name == "ArmLogicTreeNode":
+                if name in ("ArmLogicTreeNode", "ArmLogicVariableNodeMixin"):
                     continue
                 if issubclass(obj, arm_nodes.ArmLogicTreeNode):
                     obj.on_register()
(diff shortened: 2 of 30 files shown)


MoritzBrueckner:
Ah no, this wasn't intentional, thanks for mentioning this :) I will add another commit tomorrow.

Issue #1480, opened 2019-11-14
Volumetric light shader issue when multiple lights are present


Sandy10000:
**Description**
If a volumetric light is set in an environment with multiple lights, an error occurs and compilation fails. The error contents change depending on the combination of lights.

Report to Armory Forum by Daniel Sobrinho(http://forums.armory3d.org/t/problems-with-volumetric-lights-using-lamps-was-not-sun/3689)

1.  A combination of one "Sun" and one "Spot"(#define _Sun,_Spot._SinglePoint)
ERROR: volumetric_light.frag.glsl:104: 'lPos' : redefinition
1.  Combination of "Sun" and "Point" or "Area"(#define _Sun,_SinglePoint)
ERROR: volumetric_light.frag.glsl:112: 'visibility' : redefinition
1.  Combination of two "Spot" or two "Point"(#define _Clusters)
ERROR:volumetric_light.frag.glsl:119: 'visibility' : undeclared identifier

| combination | #define | line | message |  
---- | ---- | ---- | ----   
| One "Sun" and one "Spot" | _Sun,_Spot._SinglePoint | 104 | 'lPos' : redefinition |  
| "Sun" and "Point" or "Area" | _Sun,_SinglePoint | 112 | 'visibility' : redefinition |
| Two "Spot" or two "Point" | _Clusters| 119 | 'visibility' : undeclared identifier |

**System**
Blender:2.8
Armory:git

**Test File**
http://forums.armory3d.org/uploads/short-url/11Uvk4PBtITImTFV6uFAWl1e8dY.blend

MoritzBrueckner:
This has been fixed by https://github.com/armory3d/armory/pull/2445 and https://github.com/armory3d/armory/pull/1945.

PR #2445, opened 2022-03-02
Fix redefinition in volumetric light shader when using both sun and single spot light

MoritzBrueckner:
Fixes the first error mentioned in https://github.com/armory3d/armory/issues/1480, the other two errors seem to have been fixed already by https://github.com/armory3d/armory/pull/1945. I'm assuming the shader compiler is clever enough to remove unused variables in case neither `_Sun` nor `_SinglePoint` are defined :)

---

@luboslenco out of interest (because I'm trying to fix another lighting issue), what is the reason for supporting `ifdef` and `ifndef` for links in the accompanying .json files for shaders when the links are only resolved when the uniform is used in the shaders anyway (see [blender/arm/lib/make_datas.py](https://github.com/armory3d/armory/blob/fe19efe56a3831df43a9392d9c301c6ef8dc60ae/blender/arm/lib/make_datas.py#L82))? It looks like the links are checked twice this way, but I probably miss something important here.

The issue itself is more or less unrelated to this, some uniforms aren't correctly linked because I guess [this syntax](https://github.com/armory3d/armory/blob/fe19efe56a3831df43a9392d9c301c6ef8dc60ae/Shaders/deferred_light/deferred_light.json#L216-L239) for linking individual array elements isn't supposed to work at all. I have no idea how to fix this yet, I'll probably split the array into individual uniforms which is how it was before the shadow atlas PR.

### DIFF
--- a/Shaders/volumetric_light/volumetric_light.frag.glsl
+++ b/Shaders/volumetric_light/volumetric_light.frag.glsl
@@ -100,12 +100,13 @@ void rayStep(inout vec3 curPos, inout float curOpticalDepth, inout float scatter
 	curOpticalDepth *= exp(-tExt * stepLenWorld * density);
 
 	float visibility = 0.0;
+	vec4 lPos;
 
 #ifdef _Sun
 	#ifdef _CSM
 	mat4 LWVP = mat4(casData[4], casData[4 + 1], casData[4 + 2], casData[4 + 3]);
 	#endif
-	vec4 lPos = LWVP * vec4(curPos, 1.0);
+	lPos = LWVP * vec4(curPos, 1.0);
 	lPos.xyz /= lPos.w;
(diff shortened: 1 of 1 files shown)


luboslenco:
Hm maybe a remnant of the past, perhaps we can remove those.

> I guess [this syntax](https://github.com/armory3d/armory/blob/fe19efe56a3831df43a9392d9c301c6ef8dc60ae/Shaders/deferred_light/deferred_light.json#L216-L239) for linking individual array elements isn't supposed to work at all

Yeah, right now it seems to work only for [array of samplers](https://github.com/Kode/Kinc/blob/master/Backends/Graphics4/Direct3D11/Sources/kinc/backend/graphics4/pipeline.c.h#L315).

PR #2446, opened 2022-03-03
Add support for VS2022

MoritzBrueckner:
Should VS2022 also be the new default since it is default in Kha, or is it to early for that?


luboslenco:
Thanks! Happy to go with VS2022 as a default, to keep consistent with Kha.

MoritzBrueckner:
Implemented in https://github.com/armory3d/armory/commit/ae677e7a9429aa04cc5d7aae241d0d1b0f4e9223, I didn't want to open a PR for an already approved one-liner :)

luboslenco:
Super cool! 🙂

Issue #2136, opened 2021-03-19
Just few bug pictures


mushroomeo:
hi just wanted to try armory again and did on LTS and finding all times bugs . i too lazy to put blend file or any other except these screenshot because i think this is too obvious for devs to not find such bgus themself. anyway ill wait again for many more month untill i try armory again to see its its bugfree mostly and at least not every time crashing live patch armory session. good bye

![ARMORY BUGSS](https://user-images.githubusercontent.com/9019563/111723358-acf7e880-8863-11eb-893e-a7d3b9684def.jpeg)
![Armory Bug after remesh modifier](https://user-images.githubusercontent.com/9019563/111723360-ad907f00-8863-11eb-8bd3-0c1c7f346bf9.jpeg)
![Armory Bug if not unchecked export tangents](https://user-images.githubusercontent.com/9019563/111723361-ad907f00-8863-11eb-9726-a43ccfaa48fb.jpeg)
![Armory Bugs](https://user-images.githubusercontent.com/9019563/111723363-ae291580-8863-11eb-9f04-0c903a25a31c.jpeg)

N8n5h:
From the second and third pic it seems you encountered this issue: https://github.com/armory3d/armory/issues/1411#issuecomment-536114520, which you can workaround it by not having n-gons (faces with more than 4 vertices) if you are using some kind of normal mapping. 

The other seems to be related to materials, and (since there is no file to confirm it) I guess that it may have to do with some unsupported node: https://github.com/armory3d/armory/wiki/supported_nodes.

MoritzBrueckner:
@mushroomeo Thanks for the report. It would help a lot if you could upload the blend files anyway as it makes things a lot simpler. The error messages that weren't fixed in the linked PR (see above) are so generic that without a file it's difficult to see what went wrong. The `TBN` matrix for example that is mentioned in one of the error messages occurs in every shader with normal maps for example, so there is a lot that can go wrong there. 

> i too lazy to put blend file or any other except these screenshot because i think this is too obvious for devs to not find such bgus themself

I think you underestimate how diffcult it is to check every possible combination of nodes that could somehow break. Many nodes are just not completely supported and the code is written by many people, so it is actually pretty likely that there are many bugs and situations one forgets when testing. If it would be so simple, there wouldn't be any bugs.
Also, a simple looking bug can be caused by some setting on the user side that looks completely unrelated at first, so it's always best to upload a minimal blend file.

[...]

PR #2450, opened 2022-03-07
Various export fixes for subsurface scattering

MoritzBrueckner:
- The "Auto" option for SSS could lead to a missing `sssWidth` variable in the shader which is one of the issue mentioned here: https://github.com/armory3d/armory/issues/2136
- If multiple SSS nodes were present in a material but the first to be parsed was not used, the shader didn't use SSS. It's still far away from Blender's subsurface scattering, but it's a bit more reliable now.
- When compiling with SSS disabled and then setting it to "Auto" without clearing the cache, it could happen that the generated materials didn't activate SSS because the `_SSS` flag wasn't set. I'm not completely happy with the solution, but I'm not yet confident enough with the material exporter code to be able to refactor the code in a way that prevents the dependency/order issues that can happen. It might be a good idea to first check all the requirements of a material, store them temporary with the material and only then create the shaders.

@luboslenco Do you remember why the material ID is handled via a uniform whose value is defined during export (see [this](https://github.com/armory3d/armory/blob/8e91e5e84c15ed4d9ddc16f85a2bb00178ba0f91/blender/arm/material/make_mesh.py#L243-L251) and [this](https://github.com/armory3d/armory/blob/8e91e5e84c15ed4d9ddc16f85a2bb00178ba0f91/blender/arm/material/make.py#L90-L95))? Is it for context variants or to be able to merge similar shaders for less context switches?

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -2009,7 +2009,7 @@ Make sure the mesh only has tris/quads.""")
         overlays_used = False
         blending_used = False
         decals_used = False
-        # sss_used = False
+        sss_used = False
 
         for material in self.material_array:
             # If the material is unlinked, material becomes None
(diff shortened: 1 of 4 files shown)


luboslenco:
Thanks!

> Is it for context variants or to be able to merge similar shaders for less context switches?

Yeah looks like it was done for less context switches.

PR #2457, opened 2022-03-11
Remove unused code that substantially slowed down texture export

MoritzBrueckner:
There was some leftover code in the texture export that, according to a comment, led to Blender loading the textures into memory even though its no longer used. After removing the code, I got from 2.6s to 0.2s overall spent in `make_texture()` in a file with 15 textures on my machine.


luboslenco:
Awesome!

Issue #1134, opened 2019-01-31
String input for On Event node


Didiercode:
It's just a request to upgrade the On Event node of actual this
 
![image](https://user-images.githubusercontent.com/20455592/52047245-e4529080-2548-11e9-86da-3a0702d62bff.png)

to this 
![image](https://user-images.githubusercontent.com/20455592/52047211-d00e9380-2548-11e9-8c0e-1cc16eab92df.png)

*(drag & drop zipped .blend file here)*

PR #2466, opened 2022-03-29
Upgrade OnEvent node

rpaladin:
Upgrade for the `OnEvent` node as suggested in #1134.

### Enum Menu

* `Init` type
* `Update` type
* `Custom` type

### Screenshot

image

### Test Blend
<https://github.com/rpaladin/A3D-LN-Reviews/blob/main/A3D_LN_On_Event.blend>

### Special Thanks

I would like to especially thank @MoritzBrueckner & @QuantumCoderQC for their time, advice, help & review!


luboslenco:
Thanks for contributing!

PR #2477, opened 2022-04-08
key Interpolation Node

QuantumCoderQC:
This node allows to linearly interpolate between `0` and `1` based on a condition, For example keyboard/ mouse key down. It can be used to smooth character movement/translations that are activated on key down.

image


luboslenco:
Thanks!

PR #2489, opened 2022-05-16
Added a new class to handle editor time warning messages through Blenders UI and also added an operator to reload the Python side of Armory

ItsCubeTime:
I recorded a little video to demo:
https://user-images.githubusercontent.com/20190653/168518459-d0f05d63-b851-4ca9-ae96-2266c19f2ac0.mp4
image

The ideology is that we should now be able to in an organized way add warning messages if we for some reason believe the users configuration settings are potentially unstable. Like in this case its known that having "Shadow map atlasing" disabled on Windows based systems with more than 2 scene lights causes some really obscure error - this PR now adds handy little messages to make it easier for the user to know about the issue, even before hitting play and allows for more warnings to be added later on with minimal effort.

Then, to ease UI development, I created a new simple little "Armory Reload Blender Addon" operator.

### This PR also requires the following to be merged: https://github.com/armory3d/armsdk/pull/45


luboslenco:
Hi, looks like there is an issue with line endings, making the diff too big?

ItsCubeTime:
@luboslenco I believe I fixed the line endings now

blender/arm/props_ui.py
```diff
@@ -1423,7 +1529,24 @@ def draw(self, context):
         col.prop(rpdat, 'arm_pcfsize')
         layout.separator()
 
+        def label_multiline(text: str, context: bpy.types.Context, parent):
```
MoritzBrueckner, line 1532: I think this function and the comments a few lines below (1546 onwards) are no longer required, right?

blender/arm/props_ui.py
```diff
@@ -43,6 +44,94 @@
...
+    LOW = -1
+    NORMAL = 0
+    HIGH = 1
+    SEVERE = 2
+class EditorTimeWarnings():
+    "Do not instanciate me. Im responsible for keeping track of warnings that will later be displayed under the Armory Player panel."
+    class EditorTimeWarning():
+        "Instanciate me and append the instance to __listOfWarnings__. Im castable to string to get the formatted warning message."
+        def __init__(self, warningHeaderOrSettingsName: str, warningDescription: str, severityLevel: EditorTimeWarnings.SeverityLevels = WarningSeverityLevels.NORMAL) -> None:
+            self.warningHeaderOrSettingsName = warningHeaderOrSettingsName
```
MoritzBrueckner, line 89: Armory uses snake_case for Python variable names (see [this](https://github.com/armory3d/armory/wiki/Code-Style#python)), maybe you can change this. It's pretty clean and understandable code nonetheless!

blender/arm/props_ui.py
```diff
@@ -1423,7 +1529,24 @@ def draw(self, context):
...
+            import textwrap
+            for text in textListSplitByNewLines:
+                chars = int(context.region.width / 7)   # 7 pix on 1 character
+                wrapper = textwrap.TextWrapper(width=chars)
+                text_lines = wrapper.wrap(text=text)
+                for text_line in text_lines:
+                    parent.label(text=text_line)
         layout.prop(rpdat, 'rp_shadowmap_atlas')
+        if (not rpdat.rp_shadowmap_atlas) and os.name == 'nt':
+            EditorTimeWarnings.registerWarning("Armory Render Path > Shadows > Shadow Map Atlasing", "Known to prevent Armory from starting on Windows when there are more than 2 scene lights.", True, context, layout)
```
MoritzBrueckner, line 1543: This sounds as if the Shadow Map Atlas option would prevent Armory from compiling whereas the opposite is the case – enabling this option is a workaround to fix shader compilation on Direct X (which may be better wording than "Windows" since it runs perfectly fine when targeting html5 on Windows for example, it's not the OS that's the problem).

I really hope that we'll find that nasty shader compilation bug soon so that we can remove this warning again.

blender/arm/props_ui.py
```diff
@@ -43,6 +44,94 @@
...
+            wrapper = textwrap.TextWrapper(width=chars, initial_indent=f" {bulletSymbol} ", subsequent_indent=amountOfSpaceNecessary)
+        else:
+            wrapper = textwrap.TextWrapper(width=chars)
+        text_lines = wrapper.wrap(text=text)
+        for text_line in text_lines:
+            parent.label(text=text_line)
+
+
+from enum import Enum
+class WarningSeverityLevels(Enum): # Unfortunately this cant go inside EditorTimeWarnings as Python wont then let me reference it for default values in function parameters within EditorTimeWarnings.
```
MoritzBrueckner, line 78: I haven't tested this, but the enum looks like it could introduce a similar issue than what was fixed by https://github.com/armory3d/armory/pull/2392.

MoritzBrueckner (changes requested):
Wow! It's a great idea to show the warnings both in the corresponding UI and in a central place :)

Is there a reason why this PR is targeted towards the 3.3 LTS branch?

blender/arm/props_ui.py
ItsCubeTime: Correct, removing both the duplicate method and the comments

blender/arm/props_ui.py
ItsCubeTime: Ah yes, typo, fixing.

blender/arm/props_ui.py
ItsCubeTime: Im not a big fan of snake case :P Youre welcome to rename my variables if you like

blender/arm/props_ui.py
ItsCubeTime: The downside of intEnum is that I dont think it can be converted to strings - unlike regular enums.

I will switch the class to use string variables instead for now.

MoritzBrueckner: Using IntEnum is just one possible solution, the other would be to only define the enum once (don't redefine it when reloading) by moving the declaration into the else branch of the `if arm.is_reload(__name__):` block in the file header. Using strings is also fine for me but you loose a bit of type safety in return.

blender/arm/props_ui.py
MoritzBrueckner: Me neither, but that's how it is done in Armory and in a lot of Python projects. We should keep the project's code style as consistent as possible. Maybe VSCode has a little helper to change this automatically, otherwise I'll see if Pycharm (my Python IDE) has such a function and I can change it later.

MoritzBrueckner:
Don't forget to push your commits, I still don't see the new changes :)

> Yes, because I like working in recent versions of Blender 🤣Let me know if its problematic,

It's not problematic since the 3.3 LTS branch will eventually get merged into master, but it means that this change will not be available until then, which seems to be a bit against the intent of this change. In general, commits the 3.3. branch should be related to the version change, but now and then the master is merged into that branch to keep it up to date so that people can start using it already. It's ok for me, I was just curious :)

MoritzBrueckner (approved):

MoritzBrueckner:
Huh, apparently this PR was automatically closed when I deleted the now-merged Blender 3.3 branch. I'm not sure though whether we still need this PR to be honest, since the lights issue was [fixed a while ago](https://github.com/armory3d/armory/issues/1772). I can't reopen it since the target branch no longer exists, but if you still think that Armory should have some real-time warnings like you implemented for other things, feel free to open a new PR.

PR #2499, opened 2022-06-04
Several Draw nodes: cameras, images, string, geometric figures, etc

t3du:
sorry ive been away, i did these nodes in February, this upload was long due:

both camera nodes draw camera in the scene considering position size and windows size. Nodes can be activated and deactivated (sorry for the misspelling in the codes, im gonna fix it later)

image

image

### DIFF
--- a/Sources/armory/logicnode/OnRender2DNode.hx
+++ b/Sources/armory/logicnode/OnRender2DNode.hx
@@ -1,7 +1,5 @@
 package armory.logicnode;
 
-import armory.trait.physics.PhysicsWorld;
-
 class OnRender2DNode extends LogicNode {
 	/**
 		The current kha g2 object if the executed code is called inside the
--- a/Sources/armory/logicnode/OnRender2DNode.hx
+++ b/Sources/armory/logicnode/OnRender2DNode.hx
@@ -1,0 +1,28 @@
+package armory.logicnode;
+
+import armory.trait.physics.PhysicsWorld;
+
+class OnRender2DNode extends LogicNode {
+	/**
+		The current kha g2 object if the executed code is called inside the
+		render2D callback of a OnRender2DNode, else `null`.
+	**/
+	public static var g(default, null): Null<kha.graphics2.Graphics> = null;
+
+	public function new(tree: LogicTree) {
+		super(tree);
+		tree.notifyOnRender2D(onRender2D);
+	}
+
+	function onRender2D(g: kha.graphics2.Graphics) {
+		OnRender2DNode.g = g;
+		runOutput(0);
+		OnRender2DNode.g = null;
+	}
+
+	public static inline function ensure2DContext(nodeClassName: String) {
+		assert(Error, OnRender2DNode.g != null,
+			'$nodeClassName must be executed inside of a render2D callback. Please consult the documentation of this node.'
+		);
+	}
+}
--- a/Sources/armory/logicnode/DrawArcNode.hx
+++ b/Sources/armory/logicnode/DrawArcNode.hx
@@ -2,42 +2,34 @@ package armory.logicnode;
 
 import kha.Color;
 
+using kha.graphics2.GraphicsExtension;
+
 class DrawArcNode extends LogicNode {
 
-	var w: Int;
-	var h: Int;
-	
 	public function new(tree: LogicTree) {
 		super(tree);
-
 	}
 
 	override function run(from: Int) {
-	
-		w = iron.App.w();
-		h = iron.App.h();
-				
-		tree.notifyOnRender2D(render2D);
-		
-		runOutput(0);
+		OnRender2DNode.ensure2DContext("DrawArcNode");
 
-	}
-	
-	function render2D(g:kha.graphics2.Graphics) {
-		
-		if(inputs[1].get()){
-		
-			var sw = iron.App.w()/w;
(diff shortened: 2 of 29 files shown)


MoritzBrueckner:
Wow, that's a really good and useful idea!

One question though, what happens if the user activates the "In" input multiple times? Doesn't this result in multiple render callbacks being registered (i.e. the old render callbacks aren't removed)? Currently it seems that you can draw multiple things with one node but you can't "un-draw" them. It may make more sense to only allow one drawing per node (then you could even move the onRender2D registration into the constructor) and maybe you don't even need the "In" input then.

t3du:
im sure there are a lot of ways to do it... i did it this way because i needed the nodes to be activated at an specific event and i wanted to be able to change the parameters at running in order to achieve animation for those nodes...  and it was useful to me so i thought others could use it too. Feel free to discard or change the code as you want =)

MoritzBrueckner:
What about adding an "On Render2D" node and then these new draw nodes work like the corresponding Kha API calls (immediate-mode-like)? This way you can still change parameters dynamically (and decide what is drawn in the first place, which would not work that easily with my first proposal), and since there could be only one render callback for all draw nodes you don't have the problem of accidentally registering dozens of callbacks over and over. Maybe the nodes then need to check whether they're in a 2D context (a static variable that the Render2D event node would set) so that they don't do anything if used outside of the render callback. What do you think of this proposal?

QuantumCoderQC:
> What about adding an "On Render2D" node and then these new draw nodes work like the corresponding

Yes, this is a better option in my opinion.

> Maybe the nodes then need to check whether they're in a 2D context (a static variable that the Render2D event node would set)

Yes, this would be necessary. But how/ when would one un-set the `context` static variable? Also, there are a few nodes such as `Draw Camera` and `Draw Camera To Texture` node that also need `notifyOnRender` callback. How would this work in these cases?

MoritzBrueckner:
> But how/ when would one un-set the context static variable?

Since the render 2D callbacks don't run concurrently it should be easily doable like this:

```haxe
class OnRender2DNode extends LogicNode {
	public static var(default, null): inRender2DContext = false;

	public function new(tree: LogicTree) {
		super(tree);
		tree.notifyOnRender2D(render2D);
	}

	function render2D(g:kha.graphics2.Graphics) {
		OnRender2DNode.inRender2DContext = true;
		runOutput(0);
		OnRender2DNode.inRender2DContext = false;
	}
}
```

> Also, there are a few nodes such as Draw Camera and Draw Camera To Texture node that also need notifyOnRender callback. How would this work in these cases?

I only glanced through their sources, but it looks like it should be possible for them to register their own render callback. This would be completely independent of the 2D render callback. The drawback of this would be that it would be called each frame, independent of whether the rendertarget is actually used or not... Yet another alternative would be to not use a render 2D callback and instead of drawing directory add each draw command to a command queue so that a procedure can decide later when to draw the scene first and when not.

QuantumCoderQC:
> Since the render 2D callbacks don't run concurrently it should be easily doable like this:

Cool. This would work. 



> The drawback of this would be that the render callback is called each frame, independent of whether the rendertarget is actually used or not

How about this: Only two of the nodes here(Draw Camera and Draw Camera To Texture) require camera rendering. Why not replace the `in` socket with `start` and `stop` input sockets instead just for these two nodes. The rest of the 2D nodes can be modified as you suggested. This might be simpler to both understand and implement.

MoritzBrueckner:
Sounds good! I would also propose to use assertions to check whether the other nodes are called in a 2D context and else display a warning. It's a bit unfortunate that it's a runtime problem then instead of compile/export-time problem, but I currently don't know of any better way.

@t3du do you want to implement this? Otherwise I can give it a try and add some commits to this PR (I think it's possible as a collaborator to append commits to PRs).

t3du:
Be my guest, i couldnt... i didnt even understand what you were talking
about.. i did my best to do what i did with the knowledge i have... you are
in another dimension for me...

Il Lun 18 Lug 2022, 9:57 AM Moritz Brückner ***@***.***> ha
scritto:

MoritzBrueckner:
I think it's ready to merge now :)

Notable changes:
- Implemented the changes discussed above. The camera and image sequence nodes now have start/stop inputs and the other draw nodes must be called from the new `On Render2D` node. I wasn't sure about the design of the image sequence node, it's a bit complicated right now and I think maybe it makes more sense to let the user implement image sequences by using `Draw Image` nodes in a loop. This way there is more fine-grained control which is not possible with the current, very general solution.
- Coordinates are now mostly in pixels instead of fractions of the screen. Not all applications want fractions, and if they do it is possible to simply multiply the screen size or whatever total area they want to use. This issue came up because in some nodes the metrics were scaled by the average of the height and width which may not be expected in every case.
- Added documentation for all nodes.

Sources/armory/logicnode/OnRender2DNode.hx
```diff
@@ -0,0 +1,28 @@
+package armory.logicnode;
+
+import armory.trait.physics.PhysicsWorld;
```
QuantumCoderQC, line 3: I think PhysicsWorld is not needed here.

MoritzBrueckner: Good catch, thanks :)

PR #2509, opened 2022-06-27
FirstPersonController.hx anti chess fool cheat.

e2002e:
The first controller node allows player to walk faster when using a combination of two keys,
because vectors are added as is without modifications ; I changed the code to get the correct speed. Also sped up the movement overall.

### DIFF
--- a/Sources/armory/trait/FirstPersonController.hx
+++ b/Sources/armory/trait/FirstPersonController.hx
@@ -1,4 +1,4 @@
-package armory.trait;
+package arm;
 
 import iron.math.Vec4;
 import iron.system.Input;
@@ -13,7 +13,7 @@ class FirstPersonController extends CameraController {
 	public function new() { super(); }
 #else
 
-	var head: Object;
+	var head : Object;
 	static inline var rotationSpeed = 2.0;
 
 	public function new() {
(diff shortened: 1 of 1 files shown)


MoritzBrueckner:
Thanks! Is there a reason for not simply normalizing the `dir` vector after the directional components have been set?

Also, please make sure to use tabs instead of spaces :)

e2002e:
You can't avoid using geometry in the circle in this case;
adding the same force on X and Y does reach the corner of a square, which is the hypotenuse and bigger that going in one single direction.
By using sin/cos we can calibrate the speed to be always the same regardless of pressing one or two keys.

QuantumCoderQC:
What @MoritzBrueckner means is that if the direction/force vector is normalized, it always lies on the circle. No need to use polar coordinate system.

image

e2002e:
I thought normalizing was scaling a value in the range 0-1
I tried to do as you said and (to my surprise) it acts just like my complex script.
Also fixing an issue that the player had to first look in some direction to make the solution effective.

MoritzBrueckner:
Normalizing a vector always sets its length to 1 (so it lies on the unit circle) while keeping the direction. The vector is divided by its length.

It looks like you accidentally pushed a commit related to the water shader to this PR, maybe you can revert that (after saving it of course!) and then optionally squash the commits starting from that commit so that the history is clean?

e2002e:
Sorry about that squash I really did not get this thing, anyway the water is set back as original, I'm trying to get it to reflect as if it was part of the ssr pass, no luck until now.

luboslenco:
Thanks!

Issue #2506, opened 2022-06-23
Variable of Rotation resets to 0,0,0,1 after cleaning project


JAWS81:
Thank you for contributing to Armory!

**Description**
When setting and printing afterwards of a variable for rotation. The variable is set with the correct values. But after cleaning and restarting the values shown are 0,0,0,1.  

All done with Nodes. Haven´t seen this behavior with vectors or floats yet.

**To Reproduce**
This setup on a default cube.
![grafik](https://user-images.githubusercontent.com/80793695/175402908-563f8f41-e6a7-47ed-afb9-cd0efd83f182.png)

To get the actual values again the print node needs to be connected ones to the "Get rotation" node and then the project needs to be startet again. Cleaning, and the setup can be reconnected to the shown one.

**Expected behavior**
The variable shall be set with the actual rotation. 

**System**
Blender: 2.93.6
Armory:2022-06
OS:Windows 11
Graphics card: RTX2080 Super

**Test File**
*(drag & drop the zipped .blend file here)*

e2002e:
What's wrong with the factor being at 1 ?

MoritzBrueckner:
I can confirm the issue, it seems that the rotation variable node counts its inputs at a point in time where it doesn't have any inputs yet. As a result of this it then prevents values from being written to it.

@niacdoial Hi, can you remember why you added the `input_length` checks in `RotationNode.hx`? Node inputs are added after a node's constructor is called, so I'm wondering if this has worked at all at some point in time.

niacdoial:
uuugh, yeah OK, I see.
I think the `input_length` variable itself was needless optimisation (especially since `input.length()` might optimise to a single in-memory fetch as well).

Though TBH I also had/have a very bad understanding of how setting variables is supposed to work in the case where the variable node in question has inputs (or worse, some but not all inputs, which, how to even distinguish this in code since non-value variable nodes create(d) phantom value variable nodes as their inputs).

Every time I look at what I contributed here, I see more jank I've caused you. Sorry.

MoritzBrueckner:
All right, so I guess it's safe to remove the checks then. I will try to do that tomorrow.

[...]

PR #2513, opened 2022-07-03
Fix RotationNode.set(): get rid of invalid input length check

MoritzBrueckner:
Fixes https://github.com/armory3d/armory/issues/2506.

The rotation node counted its input sockets in the constructor, so the count was always 0 because no sockets existed yet.


luboslenco:
👍👍

PR #2517, opened 2022-07-13
Introduce Pulse Node

knowledgenude:
@ **TriVoxel** in DC was needing help for creating an automatic weapon logic, so I propose a node to further facilitate this

The node allows to **limit the rate** of the signal

PulseNodePreview

Most common uses:
- Prevent spam
- Control weapon fire rate

*Refactored. Thanks to @ QuantumCoderQC by the new layout idea


luboslenco:
Thanks!

Issue #2520, opened 2022-07-18
Transparency and lights


e2002e:
I have a several transparent shaders that will not compile when using any other light type than 'sun'.
The error is:
/compiled/Shaders/Flame_translucent.frag.glsl:162: 'sampleLight' : no matching overloaded function found 

Voxel shadows is enabled, and it does work neither with it disabled nor with shadow map atlas (these are not compatible).
The reason I want to add another light is because I have emissive materials and they don't light the scene.

Also this should be another issue, but since this regards also transparent materials I'm posting here:
When using volume light and having a transparent material that is untouched by light but a shaft goes in front of it. The material stands out:

![Capture d’écran de 2022-07-18 12-27-53](https://user-images.githubusercontent.com/29781855/179493677-b05297b3-4d8d-4526-b184-a89ba3307ed0.png)

Forward renderer does manage this issue but then I cannot find the correct settings for the glass.

MoritzBrueckner:
Hi, could you please share the blend file or at least the generated `/compiled/Shaders/Flame_translucent.frag.glsl`? What and how many lights do you have in the scene?

e2002e:
[church.zip](https://github.com/armory3d/armory/files/9137582/church.zip)
I packed the textures it should display.
By default it illustrates the issue (volume light and deferred rendering)

e2002e:
I'm sending you the current version of the church,
https://ufile.io/dljk3ove
I had an issue previously with tesselation and had to uncheck 'precompute tangents'. This one's fine though.

MoritzBrueckner:
Does the original error message from your first post

```
/compiled/Shaders/Flame_translucent.frag.glsl:162: 'sampleLight' : no matching overloaded function found
```

still happen for you in that exact file you've shared? Then I'm afraid I can't reproduce it. The volumetric lighting issue should be fixed in https://github.com/armory3d/armory/pull/2523, apart from that I can only reproduce an issue with the tesselation in your file that I will try to fix once I get more time (it's a very weird error message I've never seen before in a similar way).

e2002e:
Yes, same error with a point lamp added. Can you add one ?
Really good work with the fix. I'm going to try this today.

MoritzBrueckner:
Thanks, now I can confirm it.  I will investigate :)

e2002e:
To pinpoint the issue, material set to an oppacity of less than one will yield this error, whatever blend mode is used.

e2002e:
[...]

PR #2526, opened 2022-07-22
Fix for #2520

e2002e:
Translucency and volume lights fixed by Moritz and the other issue that I spotted was that the macros for the sample light require an adequate treatment at the level of the shader creation.

### DIFF
--- a/Sources/armory/renderpath/RenderPathDeferred.hx
+++ b/Sources/armory/renderpath/RenderPathDeferred.hx
@@ -651,28 +651,9 @@ class RenderPathDeferred {
 			path.setTarget("tex");
 			path.bindTarget("_main", "gbufferD");
 			path.bindTarget("buf", "tex");
+			path.bindTarget("gbuffer0", "gbuffer0");
+			path.bindTarget("gbuffer1", "gbuffer1");	
 			path.drawShader("shader_datas/water_pass/water_pass");
-		}
-		#end
-
-		#if rp_volumetriclight
-		{
-			path.setTarget("singlea");
-			path.bindTarget("_main", "gbufferD");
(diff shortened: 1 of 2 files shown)


MoritzBrueckner:
Thanks :)

Are you sure that `gbufferD` is available from the translucency shader? When I looked at this earlier today it didn't seem so to me, but maybe I'm mistaken here. `occ` however is not a uniform but defined in the shader itself, so there is no need to declare it as such.

Also, it looks like you accidentally included the changes from https://github.com/armory3d/armory/pull/2523 as well (probably doesn't matter for Git) as well as some changes relating the water pass?

e2002e:
It's hard to debug because the alpha seems to depend on the illumination of the fragments, and despite setting it to 0.999 i have a strong transparency overall.
Capture d’écran du 2022-07-23 01-37-13

e2002e:
No, no contact shadows on translucents.
I looked in the engine and what I see is that there are several type of shading passes : meshes and translucency are what we'r concerned with line ~90 in make_shader.py:
        elif rp == 'mesh':
            con = make_mesh.make(rp, rpasses)

        elif rp == 'shadowmap':
            con = make_depth.make(rp, rpasses, shadowmap=True)

        elif rp == 'translucent':
            con = make_transluc.make(rp)
The blend values are set to constants in make_transluc, and that might be the issue.
I tried other values with no succes yet.
When activating blending in the material, make_mesh.make deals with it and then unlit fragments appear. But the colors are pushed up. I am going to play with these blending to see if possible to get the correct transparency.
I opened an issue for that.

e2002e:
Closing this for now as it contains also your work.

Issue #2498, opened 2022-05-27
Armory PBR used in more advanced effects - material not compiling 


onelsonic:
Blender 2.93.9
last github repo
windows 10
Project example :  https://github.com/onelsonic/armory_tests/raw/1789f62f04f64a53cd7acfb8a13f4b86aeeb29aa/armoryPBRmat.7z

In Blender if we use a "Principled BSDF" material node everything works fine.
But when using the "Armory PBR" material node when there is a need to move around the "Armory PBR" material node in a few effects the project will simply not compile.
While the "Armory PBR" material node is required for proper normal mapping we cannot use it currently in more advanced material.
When trying so the compilation will through errors.
Some variables corresponding to other material nodes present in the material in Blender (like : Seperate RGB or MIX RGB) in the GLSL code are used before their own declaration which throw some errors at the compilation.

```
Scene exported in 0.018s
Info: Publishing project, check console for details.

ERROR: C:\armoryPBRmat_mesh.frag.glsl:13: 'Mix_Color_res' : undeclared identifier
ERROR: C:\armoryPBRmat_mesh.frag.glsl:13: 'assign' :  cannot convert from ' temp float' to ' temp 3-component vector of float'
ERROR: C:\armoryPBRmat_mesh.frag.glsl:13: '' : compilation terminated
ERROR: 3 compilation errors.  No code generated.
```

if you look at the GLSL code : 
https://github.com/onelsonic/armory_tests/blob/master/armoryPBRmat/build_armoryPBRmat/compiled/Shaders/armoryPBRmat_mesh.frag.glsl

Mix_Color_res is first used at line 13
but only declared later at line 25 ??

It seems there are some issues on the way material are exported from Blender.

e2002e:
I could run your file by setting the rendered to forward and un-setting the factor of the materials mix.
By simply doing the second step and leaving the rendered to deferred, it would compile but crash instantly.

e2002e:
The strange matter with that is there is not a single line that sets n to n = normalize(TBN * n);
The line just above is the issue and should be written after the declaration of Mix_Color_res, or deleted since it looks absurd to set the normal to the color of the object unless you plugged color output to the normal of the shader node.

onelsonic:
@e2002e what happened here did you find what was causing the issue?

PR #2531, opened 2022-07-25
Fix for point lights and translucency

e2002e:
There was a bug with translucent materials, when calling make_forward_base() and a point light is used, the number of arguments passed to sampleLight in the shader doesn't correspond with the definition.
Because the definition takes into account the SSRS and Microshadowing options, but not the part that I had to modify.
I'm not sure this is the proper way of setting the options, since it does only concern the point lights, I don't know about spots nor areas.
Seems effective though, with a glass candlestick and tesselation it looks like this:
None:
Capture d’écran du 2022-07-25 13-54-18
Microshadowing:
Capture d’écran du 2022-07-25 13-54-40
Microshadowing and ssrs:
Capture d’écran du 2022-07-25 13-55-02
ssrs without microshadowing:

Thing is microshadowing with ssr on seems to have no effect, we can see the wood being more lit hence some of it we can see through the glass ; but not on the glass itself.
Capture d’écran du 2022-07-25 13-58-58

### DIFF
--- a/blender/arm/material/make_mesh.py
+++ b/blender/arm/material/make_mesh.py
@@ -590,7 +590,6 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
 
     frag.write_attrib('vec3 vVec = normalize(eyeDir);')
     frag.write_attrib('float dotNV = max(dot(n, vVec), 0.0);')
-
     sh = tese if tese is not None else vert
     sh.add_out('vec3 eyeDir')
     sh.add_uniform('vec3 eye', '_cameraPosition')
@@ -708,6 +707,21 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
             else:
                 frag.add_uniform('vec2 lightProj', link='_lightPlaneProj', included=True)
                 frag.add_uniform('samplerCubeShadow shadowMapPoint[1]', included=True)
+        if '_MicroShadowing' in wrd.world_defs:
+            frag.add_include('std/gbuffer.glsl')
+            frag.add_uniform('sampler2D gbuffer1')
+            frag.add_uniform('sampler2D gbufferD')
+            frag.add_uniform('vec3 eyeLook')
+            frag.add_uniform('vec2 cameraProj')
+            frag.add_in('vec2 texCoord')
+            frag.add_in('vec3 viewRay')
+            frag.write('vec4 g1 = textureLod(gbuffer1, texCoord, 0.0);')
+            frag.write('vec2 occspec = unpackFloat2(g1.a);')
+            frag.write('float depth = textureLod(gbufferD, texCoord, 0.0).r * 2.0 - 1.0;')
+            frag.write('vec3 p = getPos(eye, eyeLook, normalize(viewRay), depth, cameraProj);')
+            frag.write('vec3 v = normalize(eye - p);')
+            frag.write('occspec.x = mix(1.0, occspec.x, dotNV); // AO Fresnel')
+
         frag.write('direct += sampleLight(')
         frag.write('  wposition, n, vVec, dotNV, pointPos, pointCol, albedo, roughness, specular, f0')
         if is_shadows:
@@ -717,12 +731,11 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
         if '_VoxelShadow' in wrd.world_defs and '_VoxelAOvar' in wrd.world_defs:
             frag.write('  , voxels, voxpos')
         if '_MicroShadowing' in wrd.world_defs:
-            frag.add_uniform('float occ')
-            frag.write(' , occ')
+            frag.write(' , occspec.x')
         if '_SSRS' in wrd.world_defs:
-            frag.add_include('std/ssrs.glsl')
+       	    frag.add_include('std/ssrs.glsl')
+            frag.add_uniform('sampler2D gbufferD')
             frag.add_uniform('mat4 invVP')
-            frag.add_uniform('sampler2D gbufferD') 
        	    frag.write(' , gbufferD, invVP, eye')
         frag.write(');')
 
--- a/blender/arm/material/make_mesh.py
+++ b/blender/arm/material/make_mesh.py
@@ -216,7 +216,6 @@ def make_deferred(con_mesh, rpasses):
                 else:
                     vert.write('prevwvpposition = prevWVP * spos;')
             else:
-                tese.add_out('vec4 wvpposition')
                 tese.add_out('vec4 prevwvpposition')
                 tese.write('wvpposition = gl_Position;')
                 if is_displacement:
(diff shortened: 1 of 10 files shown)


blender/arm/material/make_mesh.py
```diff
@@ -716,6 +730,13 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
             frag.write('  , true, spotData.x, spotData.y, spotDir, spotData.zw, spotRight')
         if '_VoxelShadow' in wrd.world_defs and '_VoxelAOvar' in wrd.world_defs:
             frag.write('  , voxels, voxpos')
+        if '_MicroShadowing' in wrd.world_defs:
+            frag.write(' , occspec.x')
+        if '_SSRS' in wrd.world_defs:
+       	    frag.add_include('std/ssrs.glsl')
```
MoritzBrueckner, line 736: What's the reason for including `ssrs.glsl` here? Also this line and line 739 have one space too much indentation (Python is sometimes very nitpicky about this).

blender/arm/material/make_mesh.py
```diff
@@ -708,6 +707,21 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
...
+            frag.add_include('std/gbuffer.glsl')
+            frag.add_uniform('sampler2D gbuffer1')
+            frag.add_uniform('sampler2D gbufferD')
+            frag.add_uniform('vec3 eyeLook')
+            frag.add_uniform('vec2 cameraProj')
+            frag.add_in('vec2 texCoord')
+            frag.add_in('vec3 viewRay')
+            frag.write('vec4 g1 = textureLod(gbuffer1, texCoord, 0.0);')
+            frag.write('vec2 occspec = unpackFloat2(g1.a);')
+            frag.write('float depth = textureLod(gbufferD, texCoord, 0.0).r * 2.0 - 1.0;')
```
MoritzBrueckner, line 720: What are all these calculations for if you only use  `occspec` later?

MoritzBrueckner (changes requested):
Nice!

Have you tested whether `gbufferD` actually contains valid values inside of the translucent shaders and not some garbage values or just (0,0,0)? I'm still not sure whether it does. It's still better this way than the current error message, but if you're unsure as well then maybe you want to add a little `# TODO` comment in the code so that someone debugging this code in the future knows to keep an eye on this?

blender/arm/material/make_mesh.py
MoritzBrueckner: Seems like the GitHub previews is bit misleading, I meant `float depth`, `p`, `v` and everything above that is only required for those calculations.

e2002e:
Yes, non needed computations, also right for the include.
It would be nice to set the sampler2D texture in the renderpath file but then I don't know how to set the shader as target.
rp_translucent uses 'translucent_resolve.frag' via Inc.hx

e2002e:
Microshadowing works as expected:
without it and the lamp on the left
Capture d’écran du 2022-07-26 04-57-17
And activated:
Capture d’écran du 2022-07-26 04-57-40
What gets darker are the bottom parts so there is no aberration here.
The issue is that is does not seem to be of any effect when ssrs is turned on.

blender/arm/material/make_transluc.py
```diff
@@ -18,7 +18,7 @@
 
 
 def make(context_id):
-	con_transluc = mat_state.data.add_context({ 'name': context_id, 'depth_write': False, 'compare_mode': 'less', 'cull_mode': 'clockwise', \
+	con_transluc = mat_state.data.add_context({ 'name': context_id, 'depth_write': False, 'compare_mode': 'less', 'cull_mode': 'none', \
```
MoritzBrueckner, line 21: What does this change fix?

e2002e: Nothing, I was trying to set up the glass material and forgot that.

MoritzBrueckner (approved):

e2002e:
check the related issue sir.

e2002e:
See, picture one is the candle's tray without microshadowing:

Capture d’écran du 2022-07-28 08-29-57

picture two shows lights from the flame, with microshadowing on:
Capture d’écran du 2022-07-28 08-29-44

e2002e:
I made another change which is non-related and for which I had opened an issue.
Ocean needs to reflect something else than the sky, and this last commit fixes that.
It was a big pain because I though the code needed to be changed, but the issue was from the uninitialized buffers.
So that's a copy/paste, roughly put.
Capture d’écran du 2022-07-29 19-33-10

QuantumCoderQC:
So does the ocean also reflect when SSR is disabled?

e2002e:
> So does the ocean also reflect when SSR is disabled?

Yes. To not do so I guess we need to add an extra parameter to the water, maybe differentiate environment reflection from ssr.
I just pushed a commit addressing #2498
I don't know what the lines I've removed were for but that was the solution.

MoritzBrueckner:
> I don't know what the lines I've removed were for but that was the solution.

Those lines ensure that the normals-related code is included in the shader before it's used by other parts of the shader. If I remember it correctly from my tests a while ago, the solution for this issue seemed to be a bit more complicated, but I currently have no time for it, maybe it is indeed that simple. If you haven't already, you should check the Git blame for changes around these lines and check whether your change reintroduces any of the old bugs that were fixed by code used in that function.

---

As I already mentioned in one of the related issues, please keep a single PR related to a single fix or feature, otherwise it complicates the repository's history and the discussion in the PR, makes it harder for Lubos to pick which changes to merge if he decides so, and all changes in this PR only show up as "Fix for point lights and translucency" in the release notes.

e2002e:
Okay. I'm closing this one.

e2002e (commented):
Issue appeared when an rgb mix's color is used for ArmoryPBR's normal input;
The normal is set with the texture before the later being declared.
I tried many modifications before noticing these two lines (I don't know what increasing a function does in python) and removing them fixes the issue.

e2002e (commented):
The water pass is not working, I don't know how to reestablish it. I'm wiping it from the changes.

MoritzBrueckner:
> I don't know what increasing a function does in python

`write_normal` is not a function, but a member variable of the `Shader` class ([source](https://github.com/armory3d/armory/blob/82ee3ae5efeabb0f508f646aa5a326e51f2d53ec/blender/arm/material/shader.py#L223)) which is kind of a stack-like counter in order to ensure that normals related code is added at the right position ([source](https://github.com/armory3d/armory/blob/82ee3ae5efeabb0f508f646aa5a326e51f2d53ec/blender/arm/material/shader.py#L350)). The counter is used instead of a simple boolean to make sure that leaving an "inner" normal parsing node doesn't reset the normal-parsing context of an outer node. So I guess that removing the counting and the positioning may break some other material setups in return...

e2002e:
Ok, i'll check that once more. For now i've rolled back.
So the pull request is fixing voxel shadows and shadowmap atlas, and transculent related issues

MoritzBrueckner:
> Something is puzzling me, the sampleLight() function is not used anywhere else than in make_clusters and make_mesh.

It's also used in the deferred light shaders (both standard render path and mobile), maybe that helps to resolve your confusion?

PR #2535, opened 2022-07-31
Water pass ssr

e2002e:
I'm starting to make several branches with my changes.
There it's for the water.
Also there will be one for microshadowing and ssrs with point lights and translucency.
Another for shadow map atlas with voxel shadows.
There was also a change related to tesselation and translucency. I'll probably leave it with the fork concerning ssrs/microshadowing.
And the last about the normal from a color mix.


e2002e:
This is good now, the ocean gets an 'ssr' checkbox.
I fixed a mistake in the ssr pass.

```
vec3 start = hitCoord;
for (int i = 0; i < numBinarySearchSteps; i++) {
  dir *= 0.5;
  hitCoord -= dir;
  ddepth = getDeltaDepth(hitCoord);
  if (ddepth < 0.0) start += hitCoord;
}
```
where hitCoord should be replaced by the temporary start vector.

Issue #1526, opened 2019-12-22
Make use of Refraction BSDF


trsh:
This will helps as to simulate stuff like thick glass, ice, heat distortion, etc.

PR #2537, opened 2022-08-02
Microshadowing ssr

e2002e:


e2002e:
Concerning microshadowing and ssrs, when a point light is present in the scene, the sampleLight() function is used, which takes extra parameter in this case but that were not passed.

e2002e:
Same than for the water_ssr branch, I've deleted changes on top of commited changes, that's why original files appear on the changes list.

MoritzBrueckner:
> that's why original files appear on the changes list.

Not only the files appear, but also the original changes, so this PR still includes changes to the water etc. Depending on which PR is merged first this doesn't matter, but otherwise this will lead to PRs with no changes if this one is merged first.

e2002e:
Ok, I did not manage the commit list as well as with the water pass pull request, but it's readable.

e2002e:
It is fixed for me. The last Issue i'd like to deal with is about tesselation active and a translucent material with no textures.

e2002e:
I don't get it why sample light is called inside make_mesh and make_clusters when there is a deferred_light_frag.glsl that seems to do the work.
I tried deleting the statements from the code and the result is that there are not crash and the materials still get the correct lighting.
I'm closing this.

MoritzBrueckner:
> I don't get it why sample light is called inside make_mesh and make_clusters when there is a deferred_light_frag.glsl that seems to do the work.

Because there is also a forward render path for example that makes use of this (in `make_mesh.make_forward_base()`).

e2002e:
Ok, but the translucents materials are calling make_forward_base() even though the deferred renderer is used.
Should it not check for rp_renderer before writing sampleLight() to the shader ?

e2002e:
So I understand translucent is treated apart from the other shaders when using deferred renderer.
I don't get it why the written shader has precedence over the deffered_light_frag.glsl, I did not really dive into Iron code.
I picked the normal from the deferred_light, since the default one cannot be used for microshadowing, 
 ```
```
Not setting it this way for microshadowing in forward render does a strange thing.

e2002e:
> > I don't get it why sample light is called inside make_mesh and make_clusters when there is a deferred_light_frag.glsl that seems to do the work.
> 
> Because there is also a forward render path for example that makes use of this (in `make_mesh.make_forward_base()`).

So why using make_forward_base for translucent materials when in deferred mode ? Could check for renderer and use make deferred ?

MoritzBrueckner:
> I don't get it why the written shader has precedence over the deffered_light_frag.glsl

> So why using make_forward_base for translucent materials when in deferred mode ? Could check for renderer and use make deferred ?

No, because translucent materials are forward rendered even when deferred rendering is used for all other materials. If it wouldn't be that way, objects behind translucent materials would not be rendered correctly, since an translucent shader would override the fragements/pixels of objects behind it on the gbuffer.

In forward rendering, each object calculates the lighting in its own shader, whereas in deferred rendering all objects draw their normals, colors, etc. to a gbuffer that is then used by the deferred shader to calculate the lights (so the deferred shader is usually only executed once). This reduces the computational complexity by a lot.

e2002e:
Ok, got it !

MoritzBrueckner:
> Also I'd like to raise the matter about IOR, which seems uneffective.

Can you please elaborate? If you are talking about the IOR input of one of the shader nodes, then I think it's simply not supported right now (see https://github.com/armory3d/armory/issues/1526).

e2002e:
> > Also I'd like to raise the matter about IOR, which seems uneffective.
> 
> Can you please elaborate? If you are talking about the IOR input of one of the shader nodes, then I think it's simply not supported right now (see #1526).

No, I think the principled shader's IOR field is not take into account when alpha is set to not 1.

MoritzBrueckner:
Sorry for my late answer, the linked issue is specifically about the refraction shader node, but in general there currently is no refraction support independent of the shader node used. So yes, I think only Fresnel IOR is supported right now.

e2002e:
> Sorry for my late answer, the linked issue is specifically about the refraction shader node, but in general there currently is no refraction support independent of the shader node used. So yes, I think only Fresnel IOR is supported right now.

Ok, I'm assigning myself to it.

e2002e:
Ok my issue is with the function getFloat32 from Float32Arrays,
It doesn't want to be called and states that it is not a function I guess the square bracket calles .get() which calles the master static class' getFloat32() and is wondering if not declaring these Arrays static are the issue.

e2002e:
About IOR, I am still trying to unearth voxelgi, and the refraction should come with it.
But maybe i'm lazy and postponing this part of work I auto assigned to.

e2002e:
I'm closing this.

MoritzBrueckner:
What did you want to fix with this PR? Maybe you can open an issue for it so that we don't forget about it. Also the glsl error from https://github.com/armory3d/armory/issues/2520 is still not fixed since you closed https://github.com/armory3d/armory/pull/2531. Should we reopen #2520 then?

PR #2539, opened 2022-08-02
Water pass ssr

e2002e:
Cleanup the commits list, you can compare.


MoritzBrueckner:
Nice! Does it make sense to merge the "Reflect" and "SSR" settings for the water or is it better to keep them separate? Btw, if you want to get rid of the cleanup commits you can squash them together, but it may be a bit complicated if you're using Git from the command line.

Also, it seems that you deleted some dot files and folders (`.gitattributes`, `.github` etc.) for some reason.

e2002e:
I don't think we should let the user the choice, given that the color is mix(refract, ssrColor, waterReflect), and then another mix with (color, envColor, waterReflect), it increases together. Still looking a bit strange, but I'll not come back to it tonight.

PR #2585, opened 2022-09-15
Draw to material image texture

QuantumCoderQC:
This PR extends the draw 2D nodes to be able to draw to material image textures. This would be useful to create speech bubbles and the like.

A small demo:
Animation

Shader and Logic setup:
image


luboslenco:
Awesome!

PR #2599, opened 2022-09-25
[windows-hl] Fix "After Publish" actions and add "Open in Visual Studio" operator

MoritzBrueckner:
What started as a simple bug fix became a rewrite of the entire interaction with Visual Studio:

- With the switch from kincmake to kmake, `vswhere.exe` is no longer included in the SDK, which means that every query for installed VS versions would fail. Now, Armory tries to use `vswhere.exe` from the VS installer, but it will fail for old VS versions. We could either include the vswhere binary again or simply ignore this since you need an ancient version of VS to trigger this issue.

- Moved the VS related code to a separate module to reduce redundancies and keep the codebase organized.

- Added a ton of checks for potential errors and added adequate error messages for them. Also fixed a few issues that could happen on non-Windows OSs.

- More intuitive UI that notifies the user in case a selected VS version could not be found:
  Improved UI

- Added an operator to open the exported project in Visual Studio even if the user forgot to select an "After Publish" action:
  Open in VS
  This operation will figure out the version of VS that the .sln file requires, and if it isn't installed it will try other installed versions down to the minimum required version (if specified in the .sln). It doesn't depend on the VS version the user has selected, because that could differ from the actually exported project.


luboslenco:
Wow!

PR #2601, opened 2022-09-26
Debug Console: show active camera in outliner and hide "Set Active Camera" button for active cameras

MoritzBrueckner:
The active camera is now highlighted in the outliner, making it easier to find:

Screenshot

Also, the "Set Active Camera" button is no longer visible for already active cameras.

---

@ luboslenco (or @ anyone who knows Zui better than me): is it possible to add scrollbars not only to windows, but to areas inside of windows? The outliner can get pretty long in complex scenes which then pushes the object properties out of view. Also, is there a way to check in Zui whether the mouse currently hovers some UI elements/windows? Currently, you can click "through" the debug console which in some cases is a bit annoying.

I would like to implement both these things, but my Zui knowledge is rusty and I can't seem to find a good solution.

### DIFF
--- a/Sources/armory/trait/internal/DebugConsole.hx
+++ b/Sources/armory/trait/internal/DebugConsole.hx
@@ -309,7 +309,7 @@ class DebugConsole extends Trait {
 					ui.indent();
 
 					if (selectedObject != null) {
-						if (Std.isOfType(selectedObject, iron.object.CameraObject)) {
+						if (Std.isOfType(selectedObject, iron.object.CameraObject) && selectedObject != iron.Scene.active.camera) {
 							ui.row([1/2, 1/2]);
 						}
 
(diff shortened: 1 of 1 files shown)


luboslenco:
> is it possible to add scrollbars not only to windows, but to areas inside of windows

I think it would need to be split into multiple windows, but the windows can be stacked onto each other to appear as one. There is also a way to exclude the top-most window content from scrolling via `ui.beginSticky()` and `ui.endSticky()`.

> is there a way to check in Zui whether the mouse currently hovers some UI elements/windows

Is the problem that clicking inside the debug console also clicks stuff the game?

MoritzBrueckner:
Thanks :)

How does stacking windows work? Calling `ui.window()` always seems to end the previous window.

> Is the problem that clicking inside the debug console also clicks stuff the game?

Yes. I would like to implement an option that if enabled would not activate mouse node outputs if the user hovers over the debug console, and a function in the API that allows Haxe users to check whether the debug console is hovered so that they can implement the same functionality.

Issue #2476, opened 2022-04-07
Custom web browser launch options


tong:
Passing basic command line options to the web browser would make the html5 player look like the krom player and allow to set other, probably handy, parameters. A problem here is that different browsers do not have the same command line interface. While most users presumably are using chrome (or a fork) i think it's preferable to support whatever someone wants to use.

For this i have come up with a pretty hacky solution using a shell script as middlware between armory and the browser:
Shell script: https://gist.github.com/tong/6aa3c7268b18ec1eae1fb01a46233e1b
Armory changes: https://github.com/tong/armory/blob/b03e2fc52bd2aea40648bac28e3c74caa7fdff33/blender/arm/make.py#L552

Usage:
- Set environment variable `BROWSER` to the path of the shell script: `export BROWSER=armory-browser` (the python webbrowser module is using this if set)
- Optionally set the path to the browser to use `export ARMORY_BROWSER=chromium` (otherwise xdg default)
- Optionally set extra flags to pass to the browser: `export ARMORY_BROWSER_EXTRA_OPTIONS=--disable-extensions,--disable-gl-extensions`
- Start blender, hit play

Blender will set an environment variable with all options it wants to pass to the browser.
Currently: `ARMORY_BROWSER_OPTIONS=width=1920,height=1080,fullscreen=0,vsync=0,devtools=0`
This variable is then used by the script to start the configured browser with settings applied (if availble for this browser), to the advantage that one can use another browser and/or options in another shell at the same time.

The flags implemented here (for chrome) are:
- [`--app`](https://peter.sh/experiments/chromium-command-line-switches/#app) Create a new chromeless window
- [`--window-size`](https://peter.sh/experiments/chromium-command-line-switches/#window-size) Set the window size to project size
- [`--start-fullscreen`](https://peter.sh/experiments/chromium-command-line-switches/#start-fullscreen) 
- [`--disable-gpu-vsync`](https://peter.sh/experiments/chromium-command-line-switches/#disable-gpu-vsync)
- [`--auto-open-devtools-for-tabs`](https://peter.sh/experiments/chromium-command-line-switches/#auto-open-devtools-for-tabs)

(List of chrome flags: https://peter.sh/experiments/chromium-command-line-switches/, [chrome_switches.cc](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/chrome/common/chrome_switches.cc))

While it works, i thought i drop this here in hope someone comes up with a more simple solution.

tong:
[...]

PR #2602, opened 2022-09-27
Start browser using proc_play

tong:
Use `state.proc_play` instead of `webbrowser.open` to start the html5 player.  
This allows to stop the browser from withing blender (Stop button) and prevents opening multiple browser windows, like on krom.
Not sure why this wasn't the default behavior, but i can't see any drawbacks.

### DIFF
--- a/blender/arm/make.py
+++ b/blender/arm/make.py
@@ -545,14 +545,15 @@ def build_success():
     wrd = bpy.data.worlds['Arm']
 
     if state.is_play:
+        cmd = []
         if wrd.arm_runtime == 'Browser':
             # Start server
             os.chdir(arm.utils.get_fp())
             prefs = arm.utils.get_arm_preferences()
             t = threading.Thread(name='localserver', target=arm.lib.server.run_tcp, args=(prefs.html5_server_port, prefs.html5_server_log), daemon=True)
             t.start()
-            html5_app_path = 'http://localhost:{}/{}/debug/html5'.format(prefs.html5_server_port, arm.utils.build_dir())
-            webbrowser.open(html5_app_path)
+            url = 'http://localhost:{}/{}/debug/html5/'.format(prefs.html5_server_port, arm.utils.build_dir())
+            cmd = [webbrowser.get().name, url]
         elif wrd.arm_runtime == 'Krom':
             if wrd.arm_live_patch:
                 live_patch.start()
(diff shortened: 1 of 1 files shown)


MoritzBrueckner:
Sorry that I just now got the time to test this after it is already merged, unfortunately it doesn't seem to work with Firefox on Windows:

```
  File "[...]/armory\blender\arm\handlers.py", line 143, in poll_threads
    raise e
  File "[...]/armory\blender\arm\handlers.py", line 136, in poll_threads
    callback()
  File "[...]/armory\blender\arm\make.py", line 488, in build_done
    build_success()
  File "[...]/armory\blender\arm\make.py", line 571, in build_success
    state.proc_play = run_proc(cmd, play_done)
  File "[...]/armory\blender\arm\make.py", line 78, in run_proc
    p = subprocess.Popen(cmd)
  File "[...]\Blender\Blender 3.3\3.3\python\lib\subprocess.py", line 966, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "[...]\Blender\Blender 3.3\3.3\python\lib\subprocess.py", line 1435, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
OSError: [WinError 87] Falscher Parameter
```

The traceback looks a bit weird (handlers.py:143 before handlers.py:136), I suppose that the first line is raising this very exception that only starts with the second line in the traceback.

`Falscher Parameter` translates to `wrong parameter`, the cmd list is `['', 'http://localhost:8040/build_ui_canvas/debug/html5/']`. So apparently `webbrowser.get().name` doesn't work like expected.

The type of `webbrowser.get()` is `WindowsDefault` in my case, which seems to be created whenever the browser is set up as Window's default browser. Unfortunately, it doesn't seem to get instanciated with a name because Python then delegates all the browser handling to the system and doesn't care about the specific browser.

Perhaps it would be a good compromise to check whether the returned name is empty, and if it is use `webbrowser.open()` like before?

tong:
does `start` work? (https://answers.microsoft.com/en-us/ie/forum/all/starting-default-browser-from-command-line/2e4275f1-5093-4723-9bf0-8070d476525d)

MoritzBrueckner:
Yes, but it will not have any advantage over the old approach since is is a shell command (you'd need to set `shell=True`) that will return immediately after invocation. There's also no difference regarding opening new windows, it never happend to me that Armory opened multiple browser windows with Firefox on Windows, instead the game was/is always opened in a new tab.

tong:
The goal is to be able to pass custom arguments to the browser launch command (#2476) which isn't possible when using the webbrowser module.

I can't see any difference in webbrowser.open and webbrowser.get so i am not sure why this fails. The command for the WindowsDefault browser is `webbrowser.open`. Any luck?

.. opened in a new tab. Exactly.
Makes already 2 :)

MoritzBrueckner:
Ah yes, then using `start` will be the better solution :)

It's not directly about `webbrowser.get()`, but about `webbrowser.get().name`. When using `open()`, Python is only interested in opening whatever browser the system is using, but if we want to start it from the command line we need its name, which Python simply doesn't know in this case.

PR #2605, opened 2022-09-29
Custom play launch commands

### LINKED ISSUE #2476: Custom web browser launch options

tong:
Adds the possiblity to use a custom player launch command by setting environment variables.
- `ARMORY_PLAY_HTML5`
- `ARMORY_PLAY_KROM`

The given string gets transformed with python templates using platform specific params:
- html5: `host, port, width, height, url, path, dir, browser`
- krom: `pid, audio, location, krom_path, path, resources, width, height`

If not set the default launch commands are used.

Html5 example:
```sh
export ARMORY_PLAY_HTML5='chrome --app=$url --new-window --window-size=$width,$height --auto-open-devtools-for-tabs'
```

The default launch commands:
```sh
export ARMORY_PLAY_HTML5='$BROWSER $url'
export ARMORY_PLAY_HTML5='$BROWSER http://$host:$port/$path'
export ARMORY_PLAY_HTML5='$BROWSER http://$host:$port/$dir/debug/html5'
```

The default launch command on krom:
```sh
export ARMORY_PLAY_KROM='$krom_path $path $resources --consolepid $pid'
```

This partly resolves #2476 but without any hacks.

Also resolves the issue introduced with the custom command if the browser can't get determined (#2602) by fallback to `webbrowser.open()`.

### DIFF
--- a/blender/arm/make.py
+++ b/blender/arm/make.py
@@ -556,12 +556,13 @@ def build_success():
             t = threading.Thread(name='localserver', target=arm.lib.server.run_tcp, args=(prefs.html5_server_port, prefs.html5_server_log), daemon=True)
             t.start()
             build_dir = arm.utils.build_dir()
-            url = 'http://{}:{}/{}/debug/html5/'.format(host, prefs.html5_server_port, build_dir)
+            path = '{}/debug/html5/'.format(build_dir)
+            url = 'http://{}:{}/{}'.format(host, prefs.html5_server_port, path)
+            browser = webbrowser.get().name
             if 'ARMORY_PLAY_HTML5' in os.environ:
-                str = Template(os.environ['ARMORY_PLAY_HTML5']).safe_substitute({'host': host, 'port': prefs.html5_server_port, 'width': width, 'height': height, 'url': url, 'dir': build_dir})
+                str = Template(os.environ['ARMORY_PLAY_HTML5']).safe_substitute({'host': host, 'port': prefs.html5_server_port, 'width': width, 'height': height, 'url': url, 'path': path, 'dir': build_dir, 'browser': browser})
                 cmd = re.split(' +', str)
             if len(cmd) == 0:
-                browser = webbrowser.get().name
                 if browser == '':
                     webbrowser.open(url)
                     return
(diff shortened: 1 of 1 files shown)


MoritzBrueckner:
Awsome! Is it intended that https://github.com/armory3d/armory/issues/2476 is automatically closed by this PR (since you wrote that it's only partly resolved)? Otherwise I can remove the auto-closing link.

tong:
I think it's ok to close the issue.  

What's not working is the handling of bool values since the template modules doesn't support it. Things like:
```
if fullscreen --fullscreen end
``` 
.. have to be hardcoded by the variable by simply adding `--fullscreen` or not.

MoritzBrueckner:
Hi, there seems to be another problem with `webbrowser.get().name`, apparently the `name` property doesn't always exist: https://forums.armory3d.org/t/not-possible-to-start-in-browser-or-krom-on-macosx/4991

tong:
Is a null check enough?
https://github.com/tong/armory/commit/4e51256288fa8d81a7657363c728201b97f47b7a

tong:
https://github.com/armory3d/armory/compare/main...tong:armory:fix-open-browser-call hopefully works. objections?

MoritzBrueckner:
Looks good! Thanks :)

Issue #1774, opened 2020-07-18
Emission shader problem


Simonrazer:
**Description**
The emission shader behaves very strange. Here is how a almost-white, Strength 1 shader looks like in
Cycles:
![cyc](https://user-images.githubusercontent.com/30187223/87849824-eef31800-c8eb-11ea-895f-e77b7f94a04b.png)
Eevee:
![eevee](https://user-images.githubusercontent.com/30187223/87849829-f4e8f900-c8eb-11ea-9f0f-43b9a17480aa.png)
aaaaand Armory:
![arm](https://user-images.githubusercontent.com/30187223/87849834-fb777080-c8eb-11ea-8f83-654051aa9cfc.PNG)
It is still totally shaded!

If I change the Strength to 5, color to red, and turn off shadows the result in Armory looks like this: 
![armRed](https://user-images.githubusercontent.com/30187223/87849854-42656600-c8ec-11ea-9a2b-3d41d2478c42.PNG)
when it should look like:
![eeveeRed](https://user-images.githubusercontent.com/30187223/87849859-498c7400-c8ec-11ea-9652-e18cedd6503a.png)

**Expected behavior**
The emission shader should behave just like it does in Blender, display its input without shading it in any way, with the "Strength" value acting as a sort of Gain to the appearance. This way when the Strength is set to 1 it can be used for baked materials, which should always ONLY show the color of its texture, not show extra shading created by armory. The solid shading mode is not versatile enough for this, since not all objects in a baked scene are static.

**System**
Blender: 2.82
Armory: latest git
OS: Win 10
Graphics card: Gtx 1060

[emission.zip](https://github.com/armory3d/armory/files/4941411/emission.zip)

MoritzBrueckner:
So I've searched a bit about this topic now and it looks like we need to store emission values in a buffer too. Currently it's just multiplied with the base color in the material's shader (so **before** the lighting is applied) which results in the above screenshots. I found this page from the unity documentation: https://docs.unity3d.com/2020.1/Documentation/Manual/RenderTech-DeferredShading.html which says that the lighting is added to the emission buffer which makes sense but it's not done in Armory.

The current buffer layout is the following:

```glsl
fragColor[0] = vec4(n.xy, roughness, packFloatInt16(metallic, matid));
fragColor[1] = vec4(basecol, packFloat2(occlusion, specular));
```
which equals to
|  | R8 | G8 | B8 | A8 |
| -- | -- | -- | -- | -- |
[...]

PR #2606, opened 2022-09-29
Fix socket indices of Principled BSDF node

MoritzBrueckner:
Apparently, Blender changed the Principled BSDF sockets a second time between 3.0-3.3 without a note in the release notes, at least I can't find it anywhere.

This PR fixes parsing normals and emission values. Emission is still pretty broken since we have no space left in the gbuffers to save any emission color (https://github.com/armory3d/armory/issues/1774), and I don't think that we can fix it without a 3rd render target. However, this would mean that either the 3rd render target would be bound all the time during mesh rendering, or we'd need to sort the meshes each frame to prevent to many context switches and maybe also bandwidth issues (I'm not sure what affects the bandwidth in what way, more specifically where it matters). @ luboslenco do you have an opinion on this? I think the current emission support is a problem, but I can also understand if you say that improving it would waste too much performance.

Here's a comparison between Blender and Armory, the cubes on the left use the Principled BSDF node, and the cubes on the right use the emission node. If you want I can add this file to the [armory_calibration](https://github.com/armory3d/armory_calibration) repo, but as you can see it doesn't completely work yet:

grafik

----

*Off-topic: what exactly is `rp_render_to_texture` supposed to do/solve? There's a bug with transparency if this flag is disabled, I found two possible fixes and my decision about what fix to choose depends on what this flag does.*

### DIFF
--- a/blender/arm/material/cycles_nodes/nodes_shader.py
+++ b/blender/arm/material/cycles_nodes/nodes_shader.py
@@ -49,7 +49,7 @@ def parse_addshader(node: bpy.types.ShaderNodeAddShader, out_socket: NodeSocket,
 
 def parse_bsdfprincipled(node: bpy.types.ShaderNodeBsdfPrincipled, out_socket: NodeSocket, state: ParserState) -> None:
     if state.parse_surface:
-        c.write_normal(node.inputs[20])
+        c.write_normal(node.inputs[22])
         state.out_basecol = c.parse_vector_input(node.inputs[0])
         # subsurface = c.parse_vector_input(node.inputs[1])
         # subsurface_radius = c.parse_vector_input(node.inputs[2])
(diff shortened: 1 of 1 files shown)


e2002e:
Hi, i just read that. I'm not familiar with GPU programming, yet having one more texture in the shader should cost nothing. I'm quite sure about that.

luboslenco:
> and I don't think that we can fix it without a 3rd render target

If there is no way around it we can go with the 3rd render target. Ideally it would toggle itself on/off if emission use is detected in a project? For a start let's just attach it during mesh rendering.

> If you want I can add this file to the [armory_calibration](https://github.com/armory3d/armory_calibration) repo

Yeah, that might come in handy.

> what exactly is rp_render_to_texture supposed to do/solve?

When `rp_render_to_texture` is disabled it should render directly to the screen framebuffer, instead of our own allocated texture / render target.

PR #2615, opened 2022-10-01
Fix building for Krom on Windows

MoritzBrueckner:
This fixes an exception when trying to build for Krom on Windows that was introduced in https://github.com/armory3d/armory/pull/2605/commits/435f155c8a01a606ea72fd93d63921042644da3d. Unfortunately, I didn't notice it in time for the 22.10 release. There were two different problems:

- `str` was used as a variable name, thus overwriting all uses of `str()`
- `pid` is an integer and was causing an exception on Windows because it was used in a `subprocess` command list that only accepts strings or byte objects

@luboslenco I would have pushed this to main directly, but since it's quite a major issue: could you please update the builds on itch\.io after this PR is merged? Thanks a lot, and sorry for the inconvenience :)

### DIFF
--- a/blender/arm/make.py
+++ b/blender/arm/make.py
@@ -560,8 +560,8 @@ def build_success():
             url = 'http://{}:{}/{}'.format(host, prefs.html5_server_port, path)
             browser = webbrowser.get().name
             if 'ARMORY_PLAY_HTML5' in os.environ:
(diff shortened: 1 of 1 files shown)


luboslenco:
Done, thanks for the fix! 👍

PR #2616, opened 2022-10-02
voxels gi

e2002e:
Hi, this is the unearthed voxel gi module.
I stick to the most basic setup: no compute shaders, and no voxels refraction.
The tests are a success, being able to light a room with openings from outside without leaks.
The correct setup can be hard to find, and being able to increase voxels resolution on the Z axis is a necessary option ; it allows for sticking to lower resolutions and still get the desired effect.
I also added a "1024" field for overall voxel res but have not been able to make it run.


e2002e:
Screenshot_20221002_162140

e2002e:
Screenshot_20221002_170136

e2002e:
Screenshot_20221002_173914

MoritzBrueckner:
Hi, this looks pretty awesome! I wanted to try your changes with the [Sponza](https://github.com/armory3d/armory_calibration) scene that's often used for testing GI, and without any changes I get the following error:

```
[...]\armory\Sources/armory/renderpath/RenderPathDeferred.hx:542: characters 24-32 : iron.RenderPath has no field voxelize
```

Looks like some define isn't set correctly so that `voxelize()` isn't compiled.

Unfortunately I don't know why Voxel GI was removed originally, so perhaps @ luboslenco can shed some light on this and maybe review these changes?

e2002e:
Hi Moritz, Some iron code must be also modified. You can clone mine or, if I remember well, just replace rp_voxelao with rp_gi, because the parts of the code regarding voxels are enclosed into preprocessor that use these.

e2002e:
Hi @MoritzBrueckner, I just tried with browser render, there are no voxels.
```
def voxel_support():
    # macos does not support opengl 4.5, needs metal
    return state.target != 'html5' and get_os() != 'mac'
```
When bypassing these test, on browser it yields:
Error compiling to GLSL ES 300: At least ESSL 3.10 required for shader image load store.
`

MoritzBrueckner:
> Some iron code must be also modified. 

Then please also open a PR for this, otherwise this one can't be merged.

Thanks for testing this on html5, as long as there are no errors everything looks fine :)

e2002e:
Gemarcht !

rpaladin:
Hi. Please stop spamming to get attention. Lubos will reply when he has time. Thanks.

e2002e:
> Hi. Please stop spamming to get attention. Lubos will reply when he has time. Thanks.

You'll forgive that I don't get it why a pull request opened a month ago is not being merged, this was hard work.
Seeing this makes it so that I stop working on refraction now (and to be honest, I'm just not able to achieve it).

e2002e:
I'm okay with the fact that there are discussions on other places, that I don't know of, that is why I'm saying nothing that is not related to technical aspects. You should do the same.

MoritzBrueckner:
@e2002e Hi, what's the matter with the repeated closing/re-opening of this PR and the deletion of comments? If you don't know yet whether this PR is ready, you can convert it to a draft PR in the sidebar. I don't want to wrongfully accuse you of doing this to get attention (I'm sure you know that it may take a while for Lubos to answer, especially since this is not a small change and I asked him for a small review above), but just in case that this is the reason, please stop it. It's not nice and it creates stress, and a lot of people get notified by opening/closing pull requests.

Armory is a free open source project after all, and – as unpleasant as it may sound – nobody is obliged to anything for this project, so please be patient with Lubos (who will surely decide on this PR sooner or later). We should appreciate that even with so many other projects he still continues to support Armory and that Armory has so many contributors working in their spare time.

e2002e:
There is incomprehension from me because I have set that quite some time ago now, and it's still not being merged.
To be sad but true, I truly closed this on the intend of simply giving up, not about pinging...
You should give me a bit of feedback, I don't know what's happening, also there are moments when I don't click the right button and will close and comment instead of just commenting.
I also though this was merged, did you guys merge it and then cancel ? That's not the point at all, honestly that sound like a really bad lesson, but now I think I might just have dreamed that.
I'm still working on refraction.

e2002e:
What is this shit here ? I did take several months to get this done.
There is no way I'm getting moralized about my own discouragement.

e2002e:
You'll forgive me being discouraged to the point of closing the topic, then realizing that it's a bit too much.
Don't take that as a way of of "looking for attention", seriously I don't care, that's me helping diligently here, not the opposite.
Thanks;

e2002e:
Hi there, I pulled main so that the commits are compatible. Now one thing is that emission (voxels) is not working. But looking in the code I can't find nowhere where emission was used in voxels anyway, even though it did work.

blender/arm/material/make_voxel.py
```diff
@@ -1,29 +1,251 @@
...
+    if rpdat.rp_gi == 'Voxel GI':
+        con = make_gi(context_id)
+    else:
+        con = make_ao(context_id)
 
     assets.vs_equal(con, assets.shader_cons['voxel_vert'])
-    assets.fs_equal(con, assets.shader_cons['voxel_frag'])
-    assets.gs_equal(con, assets.shader_cons['voxel_geom'])
+    assets.gs_equal(con, assets.shader_cons['voxel_frag'])
+    assets.fs_equal(con, assets.shader_cons['voxel_geom'])
```
MoritzBrueckner, line 19: Is this a typo?

MoritzBrueckner:
> Now one thing is that emission (voxels) is not working. But looking in the code I can't find nowhere where emission was used in voxels anyway, even though it did work.

Hm, does this happen with the deferred render path? I slightly changed the way emission is stored, perhaps that's the reason. Previously, emission was always stored as the basecolor if the material ID was 1. Now it's a bit more complicated, it's still stored in the basecolor if `matid == 1` (and the define `_EmissionShadeless` is set), but if the define `_EmissionShaded` is set there can be more emission values stored in a separate render target as described [here](https://github.com/armory3d/armory/blob/ce24457c9dc45c7aaecf96483c92c31915dc13bf/Shaders/custom_mat_presets/custom_mat_deferred.frag.glsl#L22), because there are cases when we need both base colors and emission colors for the same material. When I get more time I will explain all this in more detail in the wiki :)

blender/arm/material/make_voxel.py
e2002e: Oh thanks ! I would never have spotted that.

e2002e:
Hi, yes with deferred light there's not emission. I don't think that your changes are the reason for this not working. As I said there is no code using 'emission' (which is a float) in the voxel fragment shader. I think I must have changed something before pushing this...

PR #2617, opened 2022-10-02
Add option to set playback volume

QuantumCoderQC:
Adds an option to set volume of the playback sound. Thanks for the feature request by @ RichBern on Discord.

image

### DIFF
--- a/blender/arm/logicnode/sound/LN_play_sound.py
+++ b/blender/arm/logicnode/sound/LN_play_sound.py
@@ -94,9 +94,4 @@ class PlaySoundNode(ArmLogicTreeNode):
         if self.arm_version not in (0, 1):
             raise LookupError()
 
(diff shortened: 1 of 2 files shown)


blender/arm/logicnode/sound/LN_play_sound.py
```diff
@@ -85,3 +89,14 @@ def draw_buttons(self, context, layout):
         if not self.property3:
             row.enabled = False
         row.prop(self, 'property4')
+
+    def get_replacement_node(self, node_tree: bpy.types.NodeTree):
+        if self.arm_version not in (0, 1):
+            raise LookupError()
+
+        return NodeReplacement(
```
MoritzBrueckner, line 97: You can probably replace this with `return NodeReplacement.Identity(self)`

Issue #2618, opened 2022-10-03
Mix node and displacement vector


e2002e:
Using color from a mix node to a displacement vector yields error even if displace is set to vertex in renderpath.
```

ERROR: /home/i20/Game/build_boulders/compiled/Shaders/Rock_mesh.tese.glsl:46: 'n' : undeclared identifier 
ERROR: /home/i20/Game/build_boulders/compiled/Shaders/Rock_mesh.tese.glsl:46: '=' :  cannot convert from ' temp float' to ' temp 3-component vector of float'
ERROR: /home/i20/Game/build_boulders/compiled/Shaders/Rock_mesh.tese.glsl:46: '' : compilation terminated 
ERROR: 3 compilation errors.  No code generated.


SPIR-V is not generated for failed compile or link

Compiling shader 3 of 33 (Rock_mesh.tese.glsl) failed:
Shader compiler error.
(node:180405) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Shader compiler error.
(node:180405) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

```

e2002e:
fixed in #2619

PR #2620, opened 2022-10-03
Disp color mix

e2002e:
For #2618, without the changes for #2616.

### DIFF
--- a/blender/arm/material/cycles_nodes/nodes_input.py
+++ b/blender/arm/material/cycles_nodes/nodes_input.py
@@ -10,7 +10,6 @@ import arm.material.mat_state as mat_state
 from arm.material.parser_state import ParserState, ParserContext
 from arm.material.shader import floatstr, vec3str
 import arm.utils
-import arm.material.mat_utils as mat_utils
 
 if arm.is_reload(__name__):
     log = arm.reload_module(log)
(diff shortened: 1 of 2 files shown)


blender/arm/material/make_tess.py
```diff
@@ -24,6 +24,7 @@ def interpolate(tese, var, size, normalize=False, declare_out=False):
 
     if normalize:
         s += '{0}{1} = normalize({1}_0 + {1}_1 + {1}_2);\n'.format(prep, var)
+        s += 'vec3 n = normalize({0});\n'.format(var)
```
luboslenco, line 27: Do we need normalize() here?

e2002e: Well no, but in this case n is useless and should be replaced with wposition.
`vec3 TextureCoordinate_Normal_res = n;`
I'm looking into that.

MoritzBrueckner: The value of `TextureCoordinate_Normal_res` is probably generated in [`nodes_input.parse_texcoord()`](https://github.com/armory3d/armory/blob/34bf363e5f8cfeccde760eefe44d010a2b9c5a4c/blender/arm/material/cycles_nodes/nodes_input.py#L293-L294) (called from within [`cycles.parse_vector()`](https://github.com/armory3d/armory/blob/34bf363e5f8cfeccde760eefe44d010a2b9c5a4c/blender/arm/material/cycles.py#L305)), and the name is generated [here](https://github.com/armory3d/armory/blob/34bf363e5f8cfeccde760eefe44d010a2b9c5a4c/blender/arm/material/cycles.py#L534).

blender/arm/material/cycles_nodes/nodes_input.py
```diff
@@ -103,7 +104,7 @@ def parse_attribute(node: bpy.types.ShaderNodeAttribute, out_socket: bpy.types.N
     return c.cast_value('0.0', from_type='float', to_type=out_type)
 
 
-def parse_rgb(node: bpy.types.ShaderNodeRGB, out_socket: bpy.types.NodeSocket, state: ParserState) -> vec3str:
+def parse_rgb(bnode: bpy.types.ShaderNodeRGB, out_socket: bpy.types.NodeSocket, state: ParserState) -> vec3str:
```
MoritzBrueckner, line 107: Looks like a typo here, in the function you still use `node` everywhere

e2002e: fixed, don't know how this still worked

PR #2625, opened 2022-10-05
Tese compiled.inc

e2002e:
When using tesselation and wave texture (it might not be the only one causing the issue), PI is used but compiled.inc is not included.
This commit included it unconditionally when writing tese.glsl.


MoritzBrueckner:
Looks like you accidentally included the changes for https://github.com/armory3d/armory/pull/2620 as well (including the typo)

PR #2627, opened 2022-10-06
Screen space refraction

e2002e:
This adds an option in renderpath to enable refraction.
It is complex to make it a postprocess effect, and anyway this only concerns translucent objects, so what it does is writing the desired functions and code to the translucent.frag.glsl shader of the object.
This starts with a copy of the ssr_pass, for now I just replaced the reflect() function with refract(). Parts of the code has to be removed or modified in either make_transluc.py or make_mesh.py still. But since I can't test it i'm opening the pull request already to have answers on the issue:
```
float getDeltaDepth(const vec3 hit) {
	depth = textureLod(gbufferD, getProjectedCoord(hit), 0.0).r * 2.0 - 1.0;
	vec3 viewPos = getPosView(viewRay, depth, cameraProj);
	return viewPos.z - hit.z;
}
```
calls:
```
vec3 getPosView(const vec3 viewRay, const float depth, const vec2 cameraProj) {
	float linearDepth = cameraProj.y / (cameraProj.x - depth);
	// float linearDepth = cameraProj.y / ((depth * 0.5 + 0.5) - cameraProj.x);
	return viewRay * linearDepth;
}
```
But when lunching Krom, 
```
'getPosView' : no matching overloaded function found 
'=' :  cannot convert from ' const float' to ' temp 3-component vector of float' : compilation terminated
```


MoritzBrueckner:
Very cool, can't wait to test this :)

Are you sure that `getPosView()` is included in the .glsl file where the error occurs?

It looks like you accidentally included some other unrelated changes with this PR, could you please remove them? Also (regarding those changes), please don't just copy code from other projects without giving credits, not only is it a bit unfair for the authors of the respective projects, but in this case it even seems to violate their license. Since these files don't seem to be related to this PR I guess it's just for testing stuff and learning how it works, but keep that in mind ;)

e2002e:
> you define float depth on the module/file scope, if you make it local to the function (since it's only used in the function anyways), it will compile again. 

I wasn't able to compile it even this way.

> It looks like you accidentally included some other unrelated changes with this PR, could you please remove them?

Yes I'll do that.

> don't just copy code from other projects without giving credits

No worries, there are functions from the ssr_shader pass !

e2002e (commented):
This does remove the commit that was related to an attempt with ddgi.
Also fixes two typos.

e2002e:
Turns out the gbuffer file wasn't included properly (due to a typo).
I think the refraction data is there is refractCol but I don't know how to use it.

e2002e:
I'm closing this, got really enthusiastic here.
Thanks for testing anyway.

e2002e:
@MoritzBrueckner Guten tag !
Can I by any chance find out the opacity value of the fragment via the gbuffers ?
Reason is I had success with refraction when doing a post process pass, but every material was refracting (without background colors yet).

e2002e:
Honestly I'm not going to be able to accomplish this. I'm quite new to opengl and the trigonometric aspect of these things is rebuking me. I build my knowledge from a basis that is a bit pyramidal in the meaning that I'm okay with the technical aspect of computers, which gives me some kind of feeling of authority over work to be done, but I have to recognize that this mac gyver style has it's limits here. I'm going to make a few more pull requests on bugs I noticed, leaving the work of including refraction to someone mastering this field better than me.

e2002e:
I realized being a little dramatic about that. I'm taking a break for vacations this week end and will keep this task in mind.

e2002e:
I think that everything is put together. I.E there is a render pass with the translucent object, and another one after it so that it can refract itself. I can't test it though because in the make_shaders.py, when trying to write the shaders line 140, con2 doesn't seem to have been initialized. I you have an idea as about why please tell me.

e2002e:
I got an issue.
I need to copy the framebuffer before rendering the translucent objects, then draw them, and then draw the refraction.
But the forward objects don't get put in the tex or buf sampler, they won't show up in reflection, and cannot be used to refract themselves.

e2002e:
That's still not it, when using a transparent sphere it goes from the translucent data to the supposedly refracted one, which just adds white. I don't get it since rays not hitting anything are supposed to return black (0.0).

What I did is using two translucent shaders, and doing a copy of the rendered texture before rendering the first original translucent pass. In this way it's possible to get the background color.

e2002e:
I have written code to create another mesh shader 'translucent_2' when using refraction ; this allows to still use the irradiance, but when adding uniform in make_mesh.py:
```
frag.add_uniform('sampler2D tex')
frag.add_uniform('sampler2D gbufferD')

frag.add_uniform('mat4 P', link='_projectionMatrix')
frag.add_uniform('vec2 cameraProj', link='_cameraPlaneProj')
```
the terminal output tells:
```
Uniform cameraProj not found.
Uniform P not found.
```

e2002e:
I tried the following:
adding a 'refraction' pass to the material, in which I write opacity and ior value in gbuffer2, 
the material is deferred and is not rendered alonside the other meshes, because we need the scene without the object.
Once the scene is rendered we can save the screen and then draw the refractive mesh. From there we have the scene with all objects and shading, AND the scene without the refractives. We can apply a shader to the object twice, refraction plus refraction of itself (though there is issue with refracting refraction). Anyway nothing works, uniforms are not found, this is the real problem.

MoritzBrueckner:
> ```
> Uniform P not found.
> Uniform cameraProj not found.
> Uniform V3 not found.
> ```

If I'm not mistaken, these warnings happen when Kha/Kinc cannot find the referenced uniform in the shader, so please make sure that they actually exist in the generated shader and that they are used, otherwise they may be optimized away by the shader compiler (in which case this warning isn't that important though).

e2002e:
Hi @MoritzBrueckner. I have good news, the refraction is working !
Now there are two main issues:
Sunlight produces a red result without refraction, and the background texture seen through the mesh doesn't react well.
Ich wurde fur das ein bischen mer gearbeiten.
Capture d’écran du 2022-11-03 02-35-48

e2002e:
Capture d’écran du 2022-11-03 17-52-27
There still are problems, some parts of the mesh won't refract anything and be all black at certain angles.

e2002e:
Capture d’écran du 2022-11-04 14-45-35


Capture d’écran du 2022-11-04 14-45-50

e2002e:
Okay @MoritzBrueckner,
Here is what I'm coming to:
I can't get matrices on a custom shader, so I'm customizing the translucent_resolve.
Another issue is the normals data: I have to use a deferred material, skip discarding of the fragment when refraction is on, so that the normals get written in gbuffer0. The meshes are still drawn in the 'translucent' pass so they avoid being treated by the deferred_light shader. 
Now this looks correct from any angle, but the effect will appear depending on the distance between the camera and the object.
Too close or too far -> no refraction.
Capture d’écran du 2022-11-06 16-47-11

e2002e:
Hi @MoritzBrueckner,
I made some real progress working refraction.
The issue I'm facing is when ior is above 1, black borders appear on the mesh, I think that these fragments are supposed to refract the mesh, not the background. I'm going to deal with that.

Capture d’écran du 2022-11-08 18-46-57
Capture d’écran du 2022-11-08 18-47-28

e2002e:
Ok that is set.
Capture d’écran du 2022-11-09 00-47-08

Some artifacts are present but playing with the settings will probably fix that.
I need advise on why the reflections are producing this effect:
Capture d’écran du 2022-11-09 00-50-23

e2002e:
How do I get the coordinate of a Fragment ?
I need to test the hit according to the object, if it's in front then don't draw.

PR #2638, opened 2022-10-27
Fully support emission colors in all supported shader nodes

MoritzBrueckner:
This PR implements what was discussed in https://github.com/armory3d/armory/pull/2606#issuecomment-1266620114 and fixes https://github.com/armory3d/armory/issues/1774.

It took me a bit longer than expected since I was also working on improving the Armory PBR node but there still are some unresolved issues concerning automatic compatibility updates from the old PBR node so that I decided to remove the Armory PBR changes again and postpone them to hopefully next month. This way I have more time to test everything, I don't want to break anything shortly before a release.

**Major changes**:
- If there is both base color and emission color in a material, it will use a dedicated render target to store the emission color. If the node setup has emission but doesn't require a base color (e.g. there are only emission nodes or mix nodes mixing them), no additional render target is used and the emission color is stored in place of the base color, just like before this PR. On the forward RP, only slight adjustments were required.

  Because of this, the emissive component of materials now looks exactly like in Eevee.

- Internally, the shader parser makes use of "emission types": a material can either be non-emissive, or it has shaded emission (= the emission color is stored in the emission render target and is simply added on top of everything), or it has shadeless emisson (= the emission color is used instead of the base color, and the base color is treated as being black to not interact with any lights). Previously, emissive materials were always "shadeless" in Armory.

- To ensure correct gbuffer indices, there are now some new defines in `compiled.glsl`: `GBUF_IDX_0`, `GBUF_IDX_1`, `GBUF_IDX_2`, `GBUF_IDX_EMISSION` and `GBUF_SIZE`. Even though the indices 0 and 1 always exist and the rendertarget 2 is also always at index 2 if it exists, it allows us to be more flexible in the future and the user doesn't have to add a bunch of \#ifdefs to their custom shaders to get the correct indices.

- The mix node now clamps the mix factor to [0, 1], like Blender does it.

**Emission in the Eevee viewport:**
emission_viewport

**Emission in Armory:**
emission_armory

Since the emission calibration file also contains more stuff for testing the PBR node, I will open the PR for it once I've implemented the improved Armory PBR node (it will get an additional "Emission Color" socket and a socket to control the shading influence) as mentioned above.

### DIFF
--- a/Shaders/custom_mat_presets/custom_mat_deferred.frag.glsl
+++ b/Shaders/custom_mat_presets/custom_mat_deferred.frag.glsl
@@ -1,15 +1,30 @@
 #version 450
 
+#include "compiled.inc"
+
 // Include functions for gbuffer operations (packFloat2() etc.)
 #include "../std/gbuffer.glsl"
 
 // World-space normal from the vertex shader stage
 in vec3 wnormal;
 
-// Gbuffer output. Deferred rendering uses the following layout:
-// [0]: normal x      normal y      roughness     metallic/matID
-// [1]: base color r  base color g  base color b  occlusion/specular
-out vec4 fragColor[2];
+/*
+    The G-Buffer output. Deferred rendering uses the following render target layout:
+
+    | Index             | Needs #define   || R            | G            | B               | A                  |
+    +===================+=================++==============+==============+=================+====================+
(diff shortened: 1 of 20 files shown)


luboslenco:
Amazing!

PR #2646, opened 2022-11-02
Fixed GetDateTimeNode

Onek8:
Sorry, "All" Option would return different timestamps in the case of operating with the seconds/millisecond values. In/out action resolved this issue


Onek8:
image

luboslenco:
Thanks!

PR #2649, opened 2022-11-06
Document the event API

MoritzBrueckner:
In correspondence to the new wiki article: https://github.com/armory3d/armory/wiki/Events. Both the API docs and the wiki article use quite technical language, but I think they should be exact and leave little room for interpretation.

I noticed that `Event.get(name)` may return `[]` or `null` for events without listeners depending on how the API was used before. From an outside perspective this is pretty "random" behaviour, so I guess this should be changed?

### DIFF
--- a/Sources/armory/system/Event.hx
+++ b/Sources/armory/system/Event.hx
@@ -1,18 +1,44 @@
 package armory.system;
 
+/**
+	Detailed documentation of the event system:
+	[Armory Wiki: Events](https://github.com/armory3d/armory/wiki/events).
+**/
 class Event {
 
 	static var events = new Map<String, Array<TEvent>>();
 
+	/**
+		Send an event with the given name to all corresponding listeners. This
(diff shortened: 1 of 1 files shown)


QuantumCoderQC:
> I noticed that `Event.get(name)` may return `[]` or `null` for events without listeners depending on how the API was used before. From an outside perspective this is pretty "random" behaviour, so I guess this should be changed?

Yes, keeping it uniform is preferable. Perhaps returning `[]` is better? since `null` could mean that the event does not exist(regardless of weather it has listeners)

MoritzBrueckner:
Thanks, yeah I also think so, however in my opinion `null` would be better here. In Armory, events only really "exist" if there are listeners for them, you can send every possible event independent of whether anything is registered for it. So as long as there is no listener for an event, the listener's map's `get()` return value is naturally `null`, and I think it's better to just go with that and also remove the map's entry if the last listener was removed for an event to not keep unnecessary memory around.

Issue #1413, opened 2019-09-27
save the nimbus! - transparency


onelsonic:
So in blender you have two ways to achieve a correct 24bits transparency :

1. Cycle Renderer: You will use directly the alpha channel from a the Principled BSDF material.

2. Eevee Renderer: You will use the Principled BSDF and a Transparent BSDF and mix them together with a Mix Shader

Neither of these works in Armory:
are there any other way to achieve correct transparency other than haxe coding the pixel color with calculations?

here is my file :
https://github.com/onelsonic/armory_tests/blob/master/nimbus_24bits.7z

What I am trying to achieve, a correct transparency : 
![nimbus](https://user-images.githubusercontent.com/53947594/65795781-58424880-e16b-11e9-80b8-dc05c02f1f96.png)

Sandy10000:
Couldn't solve the problem of noise in the center, but other than that  could fix it by changing the settings.

3D Viewport : Since it is in edit mode, change to object mode.
Shader Editor : Use the settings for cycles.
Material tab : Uncheck Alpha Test as in the image.
![3](https://user-images.githubusercontent.com/48373918/66750728-2994d380-eec8-11e9-8173-042f4dc5f1cf.jpg)
![2](https://user-images.githubusercontent.com/48373918/66750733-2d285a80-eec8-11e9-8ac5-ffa8323c7c89.jpg)

Sandy10000:
Open the image "nimbus_24bits_transparent.png" used for transparency in the RenderDoc Texture Viewer, select alpha and set the black point to 1, and the same shape as noise was displayed.
![1](https://user-images.githubusercontent.com/48373918/67578020-93409780-f77c-11e9-9e45-27476dda1c5b.png)

Sandy10000:
The bug has not been fixed and is a workaround, but adding a Math node and subtracting 0.0002 from the alpha value eliminates the noise.
![1](https://user-images.githubusercontent.com/48373918/67634542-d9266880-f900-11e9-93ae-a6be88a7bc62.png)
![2](https://user-images.githubusercontent.com/48373918/67634544-daf02c00-f900-11e9-8a40-233d9206ef54.png)

onelsonic:
Hi @Sandy10000, nice workaround. Maybe we can fix it using this workaround in the code of Armory.
This works with the Cycle settings ok.
I will have a look, great finding 👍

Sandy10000:
I don't know where to fix it so far, so if you can fix it.

[...]

PR #2652, opened 2022-11-09
Ss refraction

e2002e:
Refraction's good.
Capture d’écran du 2022-11-09 18-14-36


MoritzBrueckner (changes requested):
That's looking great!

I've added a few review comments related to merging in the emission changes that you mentioned in your first version of the post, hopefully I caught everything (I'm not 100% sure).

Do you have an example file that you could share? I cannot seem to get it to work, the refraction option is enabled in the render path settings but I cannot see any refraction with the Glass BSDF in the game using the deferred renderer (it's just regular translucency).

When trying it with the forward RP I get the following error:
```
Trace: TypeError: Cannot read property 'getContext' of null
    at <anonymous>:11593:21
    at <anonymous>:3986:5
    at <anonymous>:13108:5
    at Function.getSceneRaw (<anonymous>:3997:4)
    at Function.parse (<anonymous>:13104:18)
    at Function.getShader (<anonymous>:3979:24)
    at iron_RenderPath.loadShader (<anonymous>:11592:18)
    at Function.initTranslucency (<anonymous>:1108:30)
    at Function.init (<anonymous>:1663:25)
    at get (<anonymous>:2832:39)
```
This looks like either the `translucent_resolve` or the `copy_pass` pass isn't correctly included.

blender/arm/material/cycles_nodes/nodes_shader.py
```diff
@@ -33,13 +33,13 @@ def parse_mixshader(node: bpy.types.ShaderNodeMixShader, out_socket: NodeSocket,
     fac_inv_var = c.node_name(node.name) + '_fac_inv'
     state.curshader.write('{0}float {1} = clamp({2}, 0.0, 1.0);'.format(prefix, fac_var, fac))
     state.curshader.write('{0}float {1} = 1.0 - {2};'.format(prefix, fac_inv_var, fac_var))
+    bc1, rough1, met1, occ1, spec1, opac1, rior1, emi1 = c.parse_shader_input(node.inputs[1])
+    bc2, rough2, met2, occ2, spec2, opac2, rior2, emi2 = c.parse_shader_input(node.inputs[2])
```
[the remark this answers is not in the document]
e2002e: I think I got that back to what it was originaly

e2002e:
Hi, thanks for the review !

> Do you have an example file that you could share? I cannot seem to get it to work, the refraction option is enabled in the render path settings but I cannot see any refraction with the Glass BSDF in the game using the deferred renderer (it's just regular translucency).
> 

Do you know blenderkit ? It's an addon offering many free models: https://www.blenderkit.com/
I think you need to increase the search field to some value that fill the distance between the camera and what's behind the refractive objects.

> When trying it with the forward RP I get the following error:
> 
> ```
> Trace: TypeError: Cannot read property 'getContext' of null
>     at <anonymous>:11593:21
>     at <anonymous>:3986:5
>     at <anonymous>:13108:5
>     at Function.getSceneRaw (<anonymous>:3997:4)
>     at Function.parse (<anonymous>:13104:18)
>     at Function.getShader (<anonymous>:3979:24)
>     at iron_RenderPath.loadShader (<anonymous>:11592:18)
>     at Function.initTranslucency (<anonymous>:1108:30)
>     at Function.init (<anonymous>:1663:25)
>     at get (<anonymous>:2832:39)
> ```
> 
> This looks like either the `translucent_resolve` or the `copy_pass` pass isn't correctly included.

Yes, copy pass is added under these conditions:
```
    if rpdat.rp_renderer == 'Deferred':
        assets.add_shader_pass('copy_pass')

    if rpdat.rp_render_to_texture:
        assets.add_khafile_def('rp_render_to_texture')

        if rpdat.rp_renderer == 'Forward' and not rpdat.rp_compositornodes:
            assets.add_shader_pass('copy_pass')
```

It looks like render_to_texture is not active even when post process is enabled with forward renderer, forcing adding the copy_pass shader doesn't help, result is red.

e2002e:
Okay that's sorted. Only thing is compositor needs to be disabled.

e2002e:
Don't pull that yet ! I have issue with non convex meshes

e2002e:
Fixed the artifacts, but there is an issue related to depth. Render shows objects that are between the refractive meshes and the camera.
Capture d’écran du 2022-11-15 21-36-42

e2002e:
Where is gbufferD created ?
I need to know that because i need an alternate depth buffer and the right format.

e2002e:
I got the objects that are between the camera and the refractive objects being shown on the latter.
I need someone to check on the code. Going to work on my own personal projects until then.
Bis näschte mal.

blender/arm/material/cycles_nodes/nodes_shader.py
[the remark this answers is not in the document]
MoritzBrueckner: The updated code still looks a bit wrong ([here](https://github.com/armory3d/armory/pull/2652/files#diff-bece3344f1d7948f8d43e6809a9cb9ae1589dbab73c96d67ef796d20ea68835aR37-L42) and [here](https://github.com/armory3d/armory/pull/2652/files#diff-bece3344f1d7948f8d43e6809a9cb9ae1589dbab73c96d67ef796d20ea68835aR58-L63)), the input shaders need to be parsed after we set the emission type to `NO_EMISSION` (to be potentially overridden by the input shader).

e2002e:
I'm almost finished here, only thing is that the shader should avoid rendering objects that are between the camera and the refractive objects. Instead of testing DeltaDepth < 0.0 it should be < fragmentPos but this doesn't work

blender/arm/material/cycles_nodes/nodes_shader.py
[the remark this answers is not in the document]
e2002e: Can you specify which files you'r refering to ? These links give me the whole list of commits from the start.

MoritzBrueckner:
Got a new error after checking out the latest commit of your branch, it happens with both forward and deferred RPs using a Glass BSDF node:

```
Shader flags: _Irr _Rad _Deferred _ShadowMap _CSM _Legacy _Sun _Brdf _SMAA _SSAO _SSRefraction _EmissionShadeless _EmissionShaded _Deferred _ShadowMap _CSM _Legacy _SMAA _SSAO _SSRefraction
Compositor flags: _CToneFilmic
Khafile flags: arm_deferred arm_csm arm_legacy rp_hdr rp_renderer=Deferred rp_shadowmap rp_shadowmap_cascade=1024 rp_shadowmap_cube=512 rp_background=World rp_render_to_texture rp_compositornodes rp_antialiasing=SMAA rp_supersampling=1 rp_ssgi=SSAO rp_bloom rp_ssrefr rp_ssrefr_half rp_translucency rp_gbuffer_emission
  File "[...]\armory\blender\arm\props_ui.py", line 1186, in invoke
    return self.execute(context)
  File "[...]\armory\blender\arm\props_ui.py", line 1205, in execute
    make.play()
  File "[...]\armory\blender\arm\make.py", line 519, in play
    build(target=runtime_to_target(), is_play=True)
  File "[...]\armory\blender\arm\make.py", line 439, in build
    export_data(fp, sdk_path)
  File "[...]\armory\blender\arm\make.py", line 248, in export_data
    for s in assets.shader_passes_assets[ref]:
KeyError: 'ssrefr_pass'
```
---

I'm also wondering whether it would work precision-wise to pack the IOR/opacity values from the new dedicated refraction render target in the G-Buffer into a single channel? Then it might work to get rid of the additional render target and instead use either the currently unused alpha channel in `GBUF_IDX_2` or `GBUF_IDX_EMISSION` - this way there would be one less render target used in case one of the other two targets are required as well.

Perhaps it would also make sense to implement a generally more intelligent G-Buffer channel packing that is defined by the exporter based on what is actually used in the game, but this is probably something for the future and it might have some disadvantages I don't know about...

e2002e:
> wondering whether it would work precision-wise to pack the IOR/opacity values from the new dedicated refraction render target in the G-Buffer into a single channel?

If i get what you mean this is already the case.
frag.write('fragColor[GBUF_IDX_REFRACTION] = vec4(packFloat2(rior, opacity));')

> Then it might work to get rid of the additional render target and instead use either the currently unused alpha channel in GBUF_IDX_2 or GBUF_IDX_EMISSION
It's possible but I think that it'll make the code more complex for gaining a texture, gpus have gigas of memory, a render target will use a few megabytes...

MoritzBrueckner:
> If i get what you mean this is already the case.
> frag.write('fragColor[GBUF_IDX_REFRACTION] = vec4(packFloat2(rior, opacity));')

Ah sorry, I didn't see the `packFloat2()`. Then I think it would work to store the packed value for example in the currently unused alpha channel of the emission render target (and perhaps rename it and its index to make it clear that it's not exclusively for emission). This way there is only one additional render target if the user is using both emission and screen space refraction, instead of the two targets that are currently used for it. In case only one of the options is used, there is still only one target used so nothing would change in that case. This saves a bit of memory and probably also texture reads.

> It's possible but I think that it'll make the code more complex for gaining a texture

Yeah it would certainly make the code more complex and it's out of scope for this PR, I just wanted to mention this idea to make it clear that there might be an even better solution possible. But whether it _really_ is a better solution would need to be carefully profiled...

e2002e:
I don't get some parts of the code, like what's the difference between viewRay and eyeLook, I get it the getPosView is used to retrieve the fragment's position at the camera, getPos (which I need to use) gets the fragment's position in world space.

The difference between reflection and refraction are that we need the texture to be saved before rendering the refractive objects, and also the depth buffer, this is the point where I'm at loss, There seems to be loss of data when I use the copy from before rendering meshes. Why can't I find the declaration for gbufferD anywhere ?

e2002e:
Capture d’écran du 2022-11-18 01-32-36
This shows the issue, with the plateau being hit on the way to where the refraction vector points to.
One thing I really don't get is why the refraction is correct without using two depth buffers.

e2002e:
I suspect that my copy of depth is not set right,
Do you know where gbufferD is declared ? Das is vischtisch

e2002e:
I need to know where gbufferD is created.
I searched the folder iron and armory for a clue, kein gluck.

e2002e:
Can I get some clarification on what getPosView is doing ?
I Think

e2002e:
I need to know the way to retrieve fragment's position, as I understand, getPosView returns the fragment at camera position.
I tried to use getPos but no good result.

e2002e:
Capture d’écran du 2022-11-23 18-50-50

e2002e:
Capture d’écran du 2022-11-24 15-57-49
Capture d’écran du 2022-11-24 15-58-18
Issue here, refraction depends on the steps.

e2002e:
Did you check that ? I used normalize in the appropriate way so that steps doesn't influence the effects, i.e with small steps there was about no refraction, but with big steps, refraction on tight angles would not show.
Modified the ssr and water reflect function in the same fashion.

e2002e:
@MoritzBrueckner I can't find the requested change. It seems that i set that and everything's good here. Awaiting for review.

blender/arm/material/cycles_nodes/nodes_shader.py
```diff
@@ -34,12 +34,12 @@ def parse_mixshader(node: bpy.types.ShaderNodeMixShader, out_socket: NodeSocket,
     state.curshader.write('{0}float {1} = clamp({2}, 0.0, 1.0);'.format(prefix, fac_var, fac))
     state.curshader.write('{0}float {1} = 1.0 - {2};'.format(prefix, fac_inv_var, fac_var))
 
+    bc1, rough1, met1, occ1, spec1, opac1, rior1, emi1 = c.parse_shader_input(node.inputs[1])
     mat_state.emission_type = mat_state.EmissionType.NO_EMISSION
-    bc1, rough1, met1, occ1, spec1, opac1, emi1 = c.parse_shader_input(node.inputs[1])
```
MoritzBrueckner, line 38: This is still wrong (also a few lines below and in `parse_addshader`), the shader input parsing needs to happen _after_ the emission type is temporarily set to `NO_EMISSION`.

blender/arm/material/parser_state.py
```diff
@@ -79,9 +80,11 @@ def reset_outs(self):
         self.out_occlusion = '1.0'
         self.out_specular = '1.0'
         self.out_opacity = '1.0'
+        self.out_rior = '1.450'
         self.out_emission_col = 'vec3(0.0)'
 
-    def get_outs(self) -> Tuple[vec3str, floatstr, floatstr, floatstr, floatstr, floatstr, vec3str]:
+    def get_outs(self) -> Tuple[vec3str, floatstr, floatstr, floatstr, floatstr, floatstr, floatstr, floatstr, vec3str]:
```
MoritzBrueckner, line 86: Looks like one `floatstr` too much here

blender/arm/write_data.py
```diff
@@ -569,12 +570,19 @@ def write_compiledglsl(defs, make_variants):
             f.write('#define GBUF_IDX_1 1\n')
 
             idx_emission = 2
+            idx_refraction = 2
             if '_gbuffer2' in wrd.world_defs:
                 f.write('#define GBUF_IDX_2 2\n')
                 idx_emission += 1
+                idx_emission += 1
```
MoritzBrueckner, line 577: Should this be `idx_refraction`?

blender/arm/material/cycles_nodes/nodes_shader.py
e2002e: Fixed that, it caused a smearing of the emissive light.

blender/arm/write_data.py
e2002e: To be frank, I did not manage any kind of mixing with basecolor when refraction is used.
It does refract, more or less, and that's all. No color is taken into account. This is because the refractive objects must not contribute to the render, but actually I can manage that easily.

e2002e:
I have a setup so that the refraction shader should refract also the refractive object itself, and it should be shaded.
Issue is it won't compile.

ERROR: /home/i20/Documents/build_untitled/compiled/Shaders/ssrefr_pass.frag.glsl:127: 'mix' : no matching overloaded function found 
ERROR: /home/i20/Documents/build_untitled/compiled/Shaders/ssrefr_pass.frag.glsl:127: 'assign' :  cannot convert from ' const float' to ' temp 3-component vector of float'

e2002e:
Thanks for the review, there are no performances issues with the module. I need to work on the refraction so that it does draw the shaded models, keeps the background, and draws the refraction according to these both passes.

e2002e:
I got an issue with the copy_pass.
```
TypeError: Cannot read property 'setPipeline' of null
    at iron_RenderPath.drawShader (<anonymous>:3901:17)
```
The context is shaders_data/copy_pass/copy_pass

I changed make_renderpath.py to export the shader to the assets, and do load it in renderpath.

e2002e:
Can you tell me what's up with the requested changes ? I don't find any unresolved requests, and about my issue I'm stuck with the copy_pass's context being null.

e2002e:
I need to know the difference between discarding a fragment and setting is to 0.0 and returning.
While doing reflection, it simply doesn't draw the shader. Now in refraction pass, it would render it black.
I have a refraction with shading, but the background - for which the depth is one - is rendered in a glitchy way.

e2002e:
Hallo, sehest du meine kougelns ?

Capture d’écran du 2022-12-05 21-11-54

The issue I got is with the background, I'm discarding the fragments when depth == 1, so the shader shoudn't impact on it. Yet it does.

e2002e:
I'm done, these are shaded balls with refraction.
Background's appearing, but I had to set the color explicitly, neither discarding nor setting it to black would do,
I don't get it why that would work (and it does) in reflection shader. Anyway, this is alles klar;

e2002e:
@MoritzBrueckner, what's up with the requested changes ? I marked them as solved.

e2002e:
Meine geburtag ist in sieben tags.
Ich wunche eine review and merge wenn möglish.

e2002e:
I'm having issues on flat surfaces, even with a displacement.
Capture d’écran du 2022-12-12 09-35-41
It refracts some parts of the mesh and leaves others apart.
Works good on convex meshes, don't know why.

MoritzBrueckner:
> Thanks for the review, there are no performances issues with the module.

My question wasn't really whether it works well for you, but whether you have measured the difference in performance this PR introduces.

> I need to know the difference between discarding a fragment and setting is to 0.0 and returning.
Discarding stops rendering that fragment, nothing in the output render target(s) for that fragment is changed (at least unless you already wrote to the output RT before that, I'm not sure about this case). Setting an output value to 0.0 and return will set the fragment in the RT to 0.

Sources/armory/renderpath/RenderPathDeferred.hx
```diff
@@ -838,11 +846,85 @@ class RenderPathDeferred {
...
+				#if rp_material_solid
+				path.drawShader("shader_datas/deferred_light_solid/deferred_light");
+				#elseif rp_material_mobile
+				path.drawShader("shader_datas/deferred_light_mobile/deferred_light");
+				#else
+				voxelao_pass ?
+					path.drawShader("shader_datas/deferred_light/deferred_light_VoxelAOvar") :
+					path.drawShader("shader_datas/deferred_light/deferred_light");
+				#end
+
```
MoritzBrueckner, line 918: Do I understand this code correctly that you basically render the scene twice if there are refractive materials?

Shaders/deferred_light.frag.glsl
```diff
@@ -0,0 +1,533 @@
+#version 450
```
MoritzBrueckner, line 1: Is this file used for anything? deferred_light.frag.glsl normally is at `Shaders/deferred_light/`

Shaders/ssr_pass/ssr_pass.frag.glsl
```diff
@@ -92,14 +92,14 @@ void main() {
 	n = normalize(n);
 
 	vec3 viewNormal = V3 * n;
-	vec3 viewPos = getPosView(viewRay, d, cameraProj);
+	vec3 viewPos = normalize(getPosView(viewRay, d, cameraProj));
```
MoritzBrueckner, line 95: Regarding all the `normalize()` changes from https://github.com/armory3d/armory/pull/2652/commits/f5e85a500c0485bec953fd2be63e6d7aa315e5de, how does it happen that the value isn't normalized in the first place? Shouldn't a reflection of the already normalized `viewPos` be normalized again? Was the normal vector not normalized or were there rounding issues that could result in values slightly larger than 1?

Shaders/ssrefr_pass/ssrefr_pass.frag.glsl
```diff
@@ -0,0 +1,126 @@
...
+	float roughness = g0.z;
+	vec4 gr = textureLod(gbuffer_refraction, texCoord, 0.0);
+	float ior = unpackFloat2(gr.r).x;
+	float opac = unpackFloat2(gr.r).y;
+
+	depth = textureLod(gbufferD, texCoord, 0.0).r * 2.0 - 1.0;
+
+	if (depth == 1.0) { fragColor = textureLod(tex, texCoord, 0.0); return; }
+	else if (ior == 1.0 || opac == 1.0 || roughness == 1.0) { 
+		discard; return;
```
MoritzBrueckner, line 89: I don't think you need the additional `return` here

Shaders/translucent_resolve/translucent_resolve.frag.glsl
```diff
@@ -2,24 +2,24 @@
 #version 450
 
 #include "compiled.inc"
-
-uniform sampler2D gbuffer0; // accum
-uniform sampler2D gbuffer1; // revealage
+#include "std/math.glsl"
+#include "std/gbuffer.glsl"
```
MoritzBrueckner, line 6: What's the reason for including these?

Shaders/water_pass/water_pass.json
```diff
@@ -61,11 +61,6 @@
 					"name": "P",
 					"link": "_projectionMatrix",
 					"ifdef": ["_SSR"]
-				},
-				{
-					"name": "V3",
-					"link": "_viewMatrix3",
-					"ifdef": ["_SSR"]
```
MoritzBrueckner, line 68: `V3` is still used in `water_pass.frag.glsl`, I don't think this should be removed.

Sources/armory/renderpath/Inc.hx
```diff
@@ -383,16 +381,18 @@ class Inc {
...
+		
+	    path.loadShader("shader_datas/translucent_resolve/translucent_resolve");
 	}
 
 	public static function drawTranslucency(target: String) {
 		path.setTarget("accum");
 		path.clearTarget(0xff000000);
 		path.setTarget("revealage");
 		path.clearTarget(0xffffffff);
+		path.clearTarget(0xffffffff);
```
MoritzBrueckner, line 393: Duplicate line

Sources/armory/renderpath/RenderPathDeferred.hx
```diff
@@ -783,6 +835,103 @@ class RenderPathDeferred {
...
+				path.bindTarget("tex", "tex1");
+				path.bindTarget("_main", "gbufferD");
+				path.bindTarget("gbufferD1", "gbufferD1");
+				path.bindTarget("gbuffer0", "gbuffer0");
+				path.bindTarget("gbuffer_refraction", "gbuffer_refraction");
+				path.drawShader("shader_datas/ssrefr_pass/ssrefr_pass");
+
+				#if (!kha_opengl)
+				path.setDepthFrom("tex", "gbuffer0");
+				#end
```
MoritzBrueckner, line 930: Do I understand this code correctly that you basically render the scene twice if there are refractive materials?

blender/arm/material/cycles.py
```diff
@@ -130,7 +130,8 @@ def parse_material_output(node: bpy.types.Node, custom_particle_node: bpy.types.
                     wrd.world_defs += '_EmissionShaded'
 
         if parse_opacity:
-            curshader.write('opacity = {0} - 0.0002;'.format(out_opacity))
+            curshader.write('opacity = {0};'.format(out_opacity))
```
MoritzBrueckner, line 133: What's the reason for this change? Looks like the - 0.0002 fixed https://github.com/armory3d/armory/issues/1413 (https://github.com/armory3d/armory/commit/35274c86734c04a62c6cfa4836b00d9412c84f09) although it is probably more of a workaround. There is a known issue with the - 0.0002 (see https://discord.com/channels/486771218599510021/1047945916411297872) that still needs to be resolved, but this change doesn't seem related to refraction or is it?

blender/arm/material/make_inst.py
```diff
@@ -14,7 +14,7 @@ def inst_pos(con, vert):
         vert.write('    croty * srotz * srotx + sroty * crotx, -crotz * srotx, -sroty * srotz * srotx + croty * crotx')
         vert.write(');')
         vert.write('spos.xyz = mirot * spos.xyz;')
-        if((con.data['name'] == 'mesh' or 'translucent') and vert.contains('wnormal')):
+        if con.data['name'] == 'mesh' or con.data['name'] == 'translucent' and vert.contains('wnormal'):
```
MoritzBrueckner, line 17: If I'm not mistaken `and` binds stronger than `or`, so as long as you don't really want to change the functionality of this if statement you should bring back the parentheses. The old `or 'translucent'` part was of course wrong, good catch :)

blender/arm/material/make_transcluc.py
```diff
@@ -0,0 +1,51 @@
+import bpy
```
MoritzBrueckner, line 1: Similar problem as with `Shaders/deferred_light.frag.glsl`, the filename contains a typo and thus git adds a completely new and unused file.

MoritzBrueckner (changes requested):

Shaders/ssr_pass/ssr_pass.frag.glsl
MoritzBrueckner: Oops, sorry, this is the wrong location for this comment. I am referring to the other `normalize()` calls introduced in https://github.com/armory3d/armory/pull/2652/commits/f5e85a500c0485bec953fd2be63e6d7aa315e5de

e2002e: Yes I don't get it, yet this is the only solution for refraction, so I applied it to reflection too.

Shaders/water_pass/water_pass.json
e2002e: OK

Sources/armory/renderpath/RenderPathDeferred.hx
e2002e: No, same than for overlay, the deferred shading only processes the newly drawn meshes

Shaders/ssr_pass/ssr_pass.frag.glsl
MoritzBrueckner: All right, then I assume it's due to floating point errors. For things like these a graphics debugger like RenderDoc is really helpful btw :)

Shaders/ssrefr_pass/ssrefr_pass.frag.glsl
e2002e: yep

blender/arm/material/cycles.py
e2002e: These kind of setup are called voodoo variable, they are called like so because they pop out at some point where no one but the author can understand it.
I have a correct transparency with my setup, it should say so.

blender/arm/material/make_inst.py
e2002e: got it

blender/arm/make_renderpath.py
```diff
@@ -178,10 +178,8 @@ def build():
             assets.add(assets_path + 'clouds_map.png')
             assets.add_embedded_data('clouds_map.png')
 
-    if rpdat.rp_renderer == 'Deferred':
-        assets.add_shader_pass('copy_pass')
-    else:
-        assets.add_khafile_def('rp_forward')
+    assets.add_shader_pass('copy_pass')
+    assets.add_khafile_def('rp_forward')
```
MoritzBrueckner, line 182: Aren't you adding the `rp_forward` define here even if deferred is used? This doesn't look correct.

blender/arm/material/make_mesh.py
```diff
@@ -509,7 +509,7 @@ def make_forward(con_mesh):
     wrd = bpy.data.worlds['Arm']
     rpdat = arm.utils.get_rp()
     blend = mat_state.material.arm_blending
-    parse_opacity = blend or mat_utils.is_transluc(mat_state.material)
+    parse_opacity = True#blend or mat_utils.is_transluc(mat_state.material)
```
MoritzBrueckner, line 512: What's the reason for this?

blender/arm/material/make_mesh.py
```diff
@@ -533,14 +533,14 @@ def make_forward(con_mesh):
...
-            frag.add_out('vec4 fragColor[2]')
+            frag.add_out('vec4 fragColor[3]')
             frag.add_include('std/gbuffer.glsl')
             frag.write('n /= (abs(n.x) + abs(n.y) + abs(n.z));')
             frag.write('n.xy = n.z >= 0.0 ? n.xy : octahedronWrap(n.xy);')
             frag.write('fragColor[0] = vec4(direct + indirect, packFloat2(occlusion, specular));')
             frag.write('fragColor[1] = vec4(n.xy, roughness, metallic);')
         else:
-            frag.add_out('vec4 fragColor[1]')
+            frag.add_out('vec4 fragColor[3]')
```
MoritzBrueckner, line 543: Doesn't this also affect all non-refractive materials?

blender/arm/make_renderpath.py
e2002e: True, I'm debugging this ATM.
I'm issue with forward, even while all refraction's related clauses are deactivated in the renderpath:

Trace: TypeError: Cannot read property 'isCubeMap' of undefined
    at iron_RenderPath.setTarget (<anonymous>:4668:21)

blender/arm/material/make_mesh.py
e2002e: I will have to deal with this in the same fashion than you did with the emissive

e2002e:
There is a problem with forward renderer, no matter if every clause related to refraction is taken out from RenderPathForward.hx, same issue appears, a black screen with this message:
```

Trace: TypeError: Cannot read property 'isCubeMap' of undefined
    at iron_RenderPath.setTarget (<anonymous>:4668:21)
```

e2002e (commented):
line 839-840,
We set the refractive meshes to render on buffers, they are not cleared beforehand so maybe it's rendering the scene twice.

e2002e:
Gutten tag !
Concerning the re-render of the scene, clearing targets before drawing refractive meshes should avoid other meshes from being passed through the deferred render path.
I think it's right for deferred renderer, now for the forward I don't really know why transparent materials are not rendered. I'm just guessing this is an issue with the buffer used to get the colors to refract.

e2002e:
I'm wrapping up:
Refraction activated will create a new render pass, that creates a shader using make_mesh.py and add opacity and rior (1/ior) data. This shader is used in draw meshes to set the buffers that will be used in deferred_light.frag.glsl. This way we get colors and refraction. Limitations are with self refraction and refractions of refractive objects (screen space).

MoritzBrueckner:
> line 839-840,
We set the refractive meshes to render on buffers, they are not cleared beforehand so maybe it's rendering the scene twice.

Could you please elaborate? Which file are you referring to? If possible please share a permalink to the lines so that further commits don't affect the discussion.

> The problem is with forward render path. Refractive mesh don't display.

If you haven't done so already you should give RenderDoc a try. It makes it pretty easy to see why something isn't rendering. On my end (Windows Krom D3D11) refractive meshes are visible on the forward RP, but they are rendered upside down (also on deferred btw):

refraction
(this should be a cube inside a plane)

When I run the Blender default scene (no refraction enabled) in deferred I get this:

grafik

I'm pretty short in time over the next days (christmas, family stuff) but I will try to do another review soon.

e2002e:
Hi, these post processing shaders work only on the textures provided to them. Once being processed, it's send to the GPU. If afterward we clear these textures (gbuffers), and call the iron functions that draw shaders, it will not process the old data.
These act as subsequent drawing of the final image, adding elements by elements.

e2002e:
I'm not a windows user, haven't got access to renderdoc (shade...)
Strange renders ! Did you try to use with a smooth shaded sphere ?

e2002e:
I inverted the viewPos, refraction is simillar that in eevee, yet the borders of the mesh are not displayed, probably because of the fact that these fragments are supposed to show the interior faces of the same mesh.
it should be drawn with inverted normals before hand.
correctedViewPos

e2002e:
Here we are !
* clamp(refracted.z, 0.0, 1.0) must be removed from the intensity factor's computation.

e2002e:
@MoritzBrueckner you've got to set the requested changes to resolved.
Ich wunche ihnen ein gut neue yahre !!!!

e2002e:
I do mean flipped normals. I'll try that next week.

e2002e:
refrRough

e2002e:
@MoritzBrueckner please mark requested changes as resolved if you think they are.
Also honestly I'm at the end of my werk with you. I'll have to just leave things as is. Voxel gi is güt but not this right here.

MoritzBrueckner:
> @MoritzBrueckner please mark requested changes as resolved if you think they are.

Hi, I didn't have time so far for another review, and GitHub doesn't automatically revoke the "changes requested" status if all individual review comments are resolved. So I would need to add another review for which I sadly don't have enough time at the moment. Maybe Lubos has, I don't know.

However, be aware that it takes a lot of time to read through the code and to test it locally and I don't want to do this if large parts of it get changed anyways shortly after doing the review. You mentioned forward doesn't work yet and based on your commit messages(\*) it doesn't look like this has changed, so I think you should mark this PR as a draft in the PR sidebar (or tell me to do so if you don't know how) and then you can mark it as done once it is ready :) Then someone can review your changes and if everything looks good there will be more confidence that this PR can be safely merged. Currently it's difficult for anyone else too see what the state of this PR is.

\* Also, please try to name your commits in a meaningful way (regarding "o" or "gimp" for example). They are part of public version history and nobody in the future will understand what you wanted to say with "o". If the commits themselves are just "save points" for you, consider to squash them together to form a new commit.

e2002e:
Hi @MoritzBrueckner, I did not get anything about what you said, though I understand the part about commit messages.
Thing i'd like to set is the forward renderer, but I will not.
You have all my blessings for completing this and also integrating the gi module into armory.
Aufviedersehen.

MoritzBrueckner:
I meant that it doesn't really make sense to review changes (and especially to approve them) if it's not clear whether you consider the PR to be finished or not. Reviewing changes takes time, and even if it's good to catch issues early on (and it's ok for me to do a review now and then) it is not necessarily worth reading through all the code if you plan to change the code anyways due to the PR being work-in-progress.

> Thing i'd like to set is the forward renderer, but I will not.
You have all my blessings for completing this and also integrating the gi module into armory.
Aufviedersehen.

I will convert this PR into a draft PR then, you can always undo that change and mark it as "done" once it is ready :)

e2002e:
Hi, thanks for answering.
I need to know why the deferred renderer doesn't draw anything with the attached object's fragment's shader. While in the forward it does.

14 janv. 2023 23:56:45 Moritz Brückner ***@***.***>:

MoritzBrueckner:
> I need to know why the deferred renderer doesn't draw anything with the attached object's fragment's shader. While in the forward it does.

You should give RenderDoc a try :)

e2002e:
I'm not going to do anything more concerning that, reason is: I cannot !
Das war eine güt arbeit mit inhen.

e2002e:
Here are my questionings : 
When using deferred render path, I'm NOT discarding opacity values in the shader that has to be passed to the lighting shader.
This makes the mesh transparent, and possible to render refraction with colors. I do not get it at all why there is actually transparency while the last line of deferred_light.frag is `fragColor.a = 1.0`.

This is for the part where it works, now when using forward renderer, there's no transparency. Using deferred lighting too, but opacity stays at 1.0 no matter what.

Give me a hand here.

e2002e:
Hi, I'm trying to use differentiated buffers for refractive objects.
So before rendering refraction, setTargetMeshes is made into setTargetMeshesRefr which sets gbuffer0 and gbuffer1 as targets.
Issue is there is no rendering.
I'm pushing this and you can have a look. Please try it with deferred and see if that works anyway.

e2002e:
I'm closing this since no interest. GI is yours I won't close it.

Issue #2650, opened 2022-11-07
If scene world color value is higher than 1.0; Armory breaks


rpaladin:
![image](https://user-images.githubusercontent.com/69180012/200224990-1fcb0693-ee4a-4044-9812-ff6c3dbdcb4b.png)

```
Trace: TypeError: Cannot read property 'length' of undefined
    at Function.getCameraRawByName (<anonymous>:6303:20)
    at <anonymous>:5845:29
    at Function.getSceneRaw (<anonymous>:6225:4)
    at Function.parse (<anonymous>:5844:18)
    at Function.getCamera (<anonymous>:6088:24)
    at iron_Scene.createObject (<anonymous>:5043:19)
    at traverseObjects (<anonymous>:4953:14)
    at <anonymous>:4965:6
    at iron_Scene.loadEmbeddedData (<anonymous>:5342:4)
    at <anonymous>:4934:11
```
Error occurs in both Krom and Browser players.

<hr />

Suspected culprit: https://github.com/armory3d/armory/blob/main/blender/arm/utils.py#L623-L624
Possibly a bits issue? There is quite a long print of decimals for `1.01`.
![image](https://user-images.githubusercontent.com/69180012/200227484-e0a2336d-f8a2-403b-b429-13d06a50ff2d.png)
Also possibly related: https://haxe.org/manual/types-overflow.html

PR #2656, opened 2022-11-10
Fix .arm reading issue when world color exceeds [0, 1] range

MoritzBrueckner:
This PR fixes https://github.com/armory3d/armory/issues/2650.

`arm.utils.color_to_int()` is meant to convert a 4-component color into a 32 bit integer value (for direct use as `kha.Color`\*) but it could happen that the resulting value was bigger than that. As a result, the integer would be exported as unsigned 64 bit in the .arm file and because the reader in Iron [currently does not implement this (`0xcf` prefix)](https://github.com/armory3d/iron/blob/9f0d936045214fbc507dfb8442a8b6f1abc82998/Sources/iron/system/ArmPack.hx#L46-L66) the parser returned a wrong result.

Haxe still doesn't seem to have a UInt64 type on all targets, and even the Int64 type is simulated on some platforms IIRC. Armory doesn't seem to use them at the moment, but should we still read two 32 bit integers in case of 64 bit numbers and output a garbage value just to make sure that the parser can continue correctly (+ output a warning)? It's then easier in the future to find any similar issues. I'm also not sure btw if the uint32 case is correctly implemented (it's using `readInt32()`), but I don't know of any issues with that.

_\* this is why we can clamp the function's input value, it's only used for background clearing on the forward RP. Converting the return value into a 32 bit int would introduce overflow situations._

### DIFF
--- a/blender/arm/utils.py
+++ b/blender/arm/utils.py
@@ -617,10 +617,15 @@ def update_trait_collections():
                 col = bpy.data.collections['Trait|' + t.name]
             col.objects.link(o)
 
+
 def to_hex(val):
     return '#%02x%02x%02x%02x' % (int(val[3] * 255), int(val[0] * 255), int(val[1] * 255), int(val[2] * 255))
 
-def color_to_int(val):
+
+def color_to_int(val) -> int:
+    # Clamp values, otherwise the return value might not fit in 32 bit
+    # (and later cause problems, e.g. in the .arm file reader)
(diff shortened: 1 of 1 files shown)


luboslenco:
Thanks!

> Armory doesn't seem to use them at the moment

If it's not used at the moment maybe we can raise a flag directly during export at https://github.com/armory3d/armory/blob/main/blender/arm/lib/armpack.py#L51.

Issue #2668, opened 2022-11-24
Bizarre camera reflection bug 


Armory-Community-Channel:
**A weird bug in the materials make a reflection visible on basic armory PBR or BDSF material. Even tho the are no other objects in the scene.**

*(drag & drop the zipped .blend file here)*

https://user-images.githubusercontent.com/116981508/203827267-3ec730cd-c7e1-49a1-9730-6ed8a7dceb3c.mp4

[bug.zip](https://github.com/armory3d/armory/files/10086048/bug.zip)

LVutner:
Good one. 

Possible fixes:
- Set sampler wrap mode to CLAMP (Best solution, we don't want repeated BRDF table)
- (Temporary fix) Change texture sampling method (This fixed weird reflection on my end)

*From **textureLod(senvmapBrdf, vec2(roughness, 1.0 - dotNV), 0.0).xy**
to
**texelFetch(senvmapBrdf, ivec2(vec2(roughness, 1.0 - dotNV) * 256.0), 0).xy;**

LVutner:
![image](https://user-images.githubusercontent.com/22204253/204074815-46d9ed15-8953-4371-87ac-355ec5c97841.png)

Screenshot showing the temporary fix (left) and old sampling (right)

LVutner:
I'll make a tiny pull request later, regarding this (and other) issue

MoritzBrueckner:
> * (Real fix) Set sampler wrap mode to CLAMP

In theory you could implement this [here](https://github.com/armory3d/iron/blob/827d77dcec12bd70ee05032199300f2ae21194e0/Sources/iron/object/Uniforms.hx#L99) specifically for the the BRDF texture similar to the hard-coded uniform names a few lines below that, but it probably makes more sense to instead get rid of hard-coded values and allow for configuring texture parameters in the shader pass .json similar to https://github.com/armory3d/iron/pull/118.

Armory-Community-Channel:
Bug fixed with SDK 2022.12

PR #2672, opened 2022-11-26
* FIX: Temporary fix(es) for BRDF LUT issues

LVutner:
A small fixes for BRDF LUT issues.

- Better texture sampling (Temp. fix for #2668)
- Correct texture coordinates (Fix for overbrightened edges of the meshes)


luboslenco:
Thanks!

PR #2673, opened 2022-11-26
Fix for Armory baker

Naxela:
The native Armory baker (non-lightmapper) had been throwing some errors lately related to some 3.0+ changes. Now it should work again:

image


luboslenco:
Thanks!

PR #2674, opened 2022-11-27
Voxels final

e2002e:
This is the voxels gi unearthed with emission working as due ! I had to create a new branch (once again) because started the rebase from the first pull request, and copied the whole armory folder into my fork.


e2002e:
There are a lot of difference with main, which are not being merge even though git tells me that it is.
I'm waiting for a review

e2002e:
I still did not get how emission is dealt with when using voxels, when merging with the new code that manages emission with an extra buffer, there is none.

PR #2691, opened 2022-12-05
Fix array node indexes

rpaladin:
In the past with https://github.com/armory3d/armory/commit/cf086d770ee6af9de0fa15c50876bc5f62198b0a.

I foolishly assumed `self.get_id_str()` was the same as `str(id(self))`, which broke some of the array nodes. I revert those certain line changes of mine in this PR. Sorry I didn't catch this before, I swear when I had tested them before, all of them worked, but I must have not looked hard enough: all blame goes to me.

Special thanks to @t3du for reporting these issues on Discord.


t3du:
sorry @rpaladin it seems i made a mistake with the review... could you please check the array add node in terms of the remove button?

thanks for fixing the switch node also,  do you think you could also change the output case number to match with the input value number? i thinks this is a little confusing
image

Issue #2694, opened 2022-12-10
Emission color doesn't output/display correctly if overlay setting enabled


rpaladin:
Possibly related to: https://github.com/armory3d/armory/pull/2638

Unrelated, is it normal that materials marked as overlay aren't affected by world shading or the renderpath?

![image](https://user-images.githubusercontent.com/69180012/227363501-c0cb0d81-600d-483f-84a0-2b5dc56f2d76.png)

e2002e:
The overlay meshes are drawn at the end of the render path's fucntions, they are not being shaded.
The emission material doesn't hold diffuse color so is rendered black, while the other shaders you tested do render with their base color, but don't emit nor get shaded.
I am going to work on that today.

PR #2706, opened 2022-12-14
shading and overlay

e2002e:
Overlay takes shading into account, so the emission is working as the issue #2694 mentioned.
screen
screen

### DIFF
--- a/Sources/armory/renderpath/RenderPathDeferred.hx
+++ b/Sources/armory/renderpath/RenderPathDeferred.hx
@@ -6,7 +6,6 @@ import iron.Scene;
 class RenderPathDeferred {
 
 	#if (rp_renderer == "Deferred")
-
 	static var path: RenderPath;
 
 	#if rp_voxels
@@ -53,7 +52,14 @@ class RenderPathDeferred {
 		#if rp_voxels
 		{
 			Inc.initGI();
+			#if arm_voxelgi_temporal
+			{
+				Inc.initGI("voxelsB");
+			}
+			#end
+			#if (rp_gi == "Voxel AO")
 			path.loadShader("shader_datas/deferred_light/deferred_light_VoxelAOvar");
+			#end
 		}
 		#end
 
@@ -69,9 +75,7 @@ class RenderPathDeferred {
 			t.scale = Inc.getSuperSampling();
 			t.depth_buffer = "main";
 			path.createRenderTarget(t);
-		}
 
-		{
 			var t = new RenderTargetRaw();
 			t.name = "tex";
 			t.width = 0;
@@ -81,9 +85,7 @@ class RenderPathDeferred {
 			t.scale = Inc.getSuperSampling();
 			t.depth_buffer = "main";
 			path.createRenderTarget(t);
-		}
 
-		{
 			var t = new RenderTargetRaw();
 			t.name = "buf";
 			t.width = 0;
@@ -92,9 +94,7 @@ class RenderPathDeferred {
 			t.format = Inc.getHdrFormat();
 			t.scale = Inc.getSuperSampling();
 			path.createRenderTarget(t);
-		}
 
-		{
 			var t = new RenderTargetRaw();
 			t.name = "gbuffer1";
 			t.width = 0;
@@ -115,9 +115,7 @@ class RenderPathDeferred {
 			t.format = "RGBA64";
 			t.scale = Inc.getSuperSampling();
 			path.createRenderTarget(t);
-		}
 
-		{
 			var t = new RenderTargetRaw();
 			t.name = "taa";
 			t.width = 0;
@@ -196,8 +194,7 @@ class RenderPathDeferred {
 			t.scale *= 0.5;
 			#end
 			path.createRenderTarget(t);
-		}
-		{
+		
 			var t = new RenderTargetRaw();
 			t.name = "singleb";
 			t.width = 0;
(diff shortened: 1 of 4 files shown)


e2002e:
Maybe we should add the overlay pass before the compositor to include the meshes in compositor pass ? If there is no reason to not do it, tell me.

MoritzBrueckner:
Hi, it's a good idea to also implement shading for overlay meshes, I probably just would have added the emission color and kept the shadeless output for now :D

However, I'm not sure whether we should re-run the entire deferred render step here, perhaps it makes more sense to render overlay meshes in a forward fashion. Assuming that overlay meshes probably take only a small part of the screen, it might waste too much GPU bandwidth to use the gbuffer, but this is something that needs to be profiled.

As for the reason that overlay meshes are rendered after the compositor pass (and other post processing effects) I would assume it's because of their potential use in UIs, but then there should still be some kind of tone mapping for example (and sometimes one maybe wants to include them in post processing?). Maybe we need to split the compositor pass up... @luboslenco Do you know/remember the purpose of the overlay option? Also, is it by purpose that overlay meshes were shadeless before this PR, for UI e.g.?
It makes sense of course to draw the overlay meshes after transparent meshes (and thus after deferred lighting) in order to really overlay them.

Did you test this on the forward RP? The code looks like it only works on deferred and might even break on forward rendering.

blender/arm/material/make_shader.py
```diff
@@ -97,7 +97,7 @@ def build(material: Material, mat_users: Dict[Material, List[Object]], mat_armus
             con = make_transluc.make(rp)
 
         elif rp == 'overlay':
-            con = make_overlay.make(rp)
+            con = make_mesh.make(rp, rpasses)
```
MoritzBrueckner, line 100: Why did you change this here? Also, isn't make_overlay.py completely obsolete then since it's no longer called? What's the reason for the change in make_overlay.py then?

blender/arm/material/make_overlay.py
```diff
@@ -30,7 +30,7 @@ def make(context_id):
     arm_discard = mat.arm_discard
     is_transluc = mat_utils.is_transluc(mat)
     parse_opacity = (blend and is_transluc) or arm_discard
-    make_mesh.make_base(con_overlay, parse_opacity=parse_opacity)
+    make_mesh.make_deferred(con_overlay, parse_opacity=parse_opacity)
```
MoritzBrueckner, line 33: I don't think this will work on forward RP.

Sources/armory/renderpath/RenderPathDeferred.hx
```diff
@@ -626,6 +646,12 @@ class RenderPathDeferred {
 		}
 		#end
 
+		#if ((rp_voxelgi_shadows) || (rp_voxelgi_refraction))
+		{
+			path.bindTarget(voxels, "voxels");
+		}
+		#end
+
```
MoritzBrueckner, line 654: Are these changes really part of the change of this PR or are they accidentally committed because of your Voxel GI PR?

Sources/armory/renderpath/RenderPathDeferred.hx
```diff
@@ -53,7 +52,14 @@ class RenderPathDeferred {
...
 		{
 			Inc.initGI();
+			#if arm_voxelgi_temporal
+			{
+				Inc.initGI("voxelsB");
+			}
+			#end
+			#if (rp_gi == "Voxel AO")
 			path.loadShader("shader_datas/deferred_light/deferred_light_VoxelAOvar");
+			#end
```
MoritzBrueckner, line 62: Same question here, will this even work without your Voxel GI PR?

Sources/armory/renderpath/RenderPathDeferred.hx
MoritzBrueckner: Looks like GitHub didn't like my rather large selection range, this not only refers to the 4 lines that are shown but also to the voxel-related lines above that.

blender/arm/material/make_shader.py
e2002e: shader needs to fill datas for the deferred shader

blender/arm/material/make_overlay.py
e2002e: Ok I'll find a workaround

Sources/armory/renderpath/RenderPathDeferred.hx
e2002e: No only AO

Sources/armory/renderpath/RenderPathDeferred.hx
e2002e: yes

e2002e:
> I'm not sure whether we should re-run the entire deferred render step here

The deferred shader, If I'm not mistaken, will only render on the bind texture, which is set to buf, and only contains the newly drawn overlay meshes.

MoritzBrueckner:
Sorry, I think I was a bit imprecise here, I didn't mean that it re-renders the entire scene again, but that you do another deferred lighting pass. I fear that this could be detrimental for performance because overlay meshes are likely to only fill small parts of the screen most of the time. There would be many unused fragments in the buffers that maybe waste memory or cause cache misses. I'm not a GPU expert so I might be wrong here, but I think this is something you should keep in mind (for translucent meshes we always use forward rendering for example, but this is obviously a slightly different case).

e2002e:
No, I was incorrect when saying that the target is what is rendered by the shader. I know the deferred shader deals with gbuffer0 and gbuffer1. These are overridden, when target is set to overlay. Kein Problems.

e2002e:
The issue you talk about is that when drawing overlay meshes via deferred shader, the rest of the scene can be present in the buffers. It is not because we clear the buffers beforehand.
I took into account the fact that overlay can serve as UI elements, so I'm going to add an option to shade these or not.

PR #2719, opened 2022-12-20
Implement rotate render target node

QuantumCoderQC:
New node to rotate the 2D render target.

image


t3du:
good!, i tried to enable rotation for the draw nodes when i implemented them, but this did not work because they kept rotating endlessly, so i remove that functionality

Issue #1166, opened 2019-02-13
[FR] EEVEE bloom parameters match armory bloom ?


mistajuliax:
Here you can see settings are different and doesn't match each other. It would be great to have same settings and result share by armory and eevee to preview render result expected in armory.
I don't know if it possible and wich techniques uses armory and eevee but as they both uses post process may be it is something doable ... 
![bloom 01](https://user-images.githubusercontent.com/7542691/52712003-9fd8e300-2f93-11e9-83c6-f812b9ac3289.jpg)
Another issue is the view distance ... here you can sse nothing with armory default bloom parameters unlesss you are very near from object.
![bloom 02](https://user-images.githubusercontent.com/7542691/52712185-1f66b200-2f94-11e9-826f-65a6e0a1dc58.jpg)
![bloom 03](https://user-images.githubusercontent.com/7542691/52712234-43c28e80-2f94-11e9-966c-1b4ae0973240.jpg)
[laser_beam_test.01.zip](https://github.com/armory3d/armory/files/2860358/laser_beam_test.01.zip)

*(drag & drop zipped .blend file here)*

kenthinson:
This would be great to have them match. Also it would be nice to simply remove the armory bloom settings and just read the eevee settings during compile time. That way we wouldn't have to change them in two different places.

Issue #2612, opened 2022-10-01
[$100 Bounty] High-Quality Bloom


Naxela:
A working implementation of high-quality bloom will release the bounty of $100 on Bountysource.

Armory3D already have a working and fast bloom solution which works well for lower-end devices, but for high-end visuals it starts to show it's limitations, namely in terms of anti-flicker and smoothness. An implementation which suppresses flickering and provides a softer threshold should be in focus.

Implementation material (Inspiration):
- https://github.com/keijiro/KinoBloom
- Blender Eevee Bloom: https://github.com/blender/blender/blob/594f47ecd2d5367ca936cf6fc6ec8168c2b360d0/source/blender/draw/engines/eevee/shaders/effect_bloom_frag.glsl 

Bountysource Link:
https://app.bountysource.com/issues/111898138-100-bounty-high-quality-bloom

gitgudyyao:
Here is a high-quality bloom implementation in GLSL that aims to suppress flickering and provide a softer threshold:

// uniform variables
uniform sampler2D tex; // source texture
uniform float threshold; // bloom threshold
uniform int blurPasses; // number of blur passes
uniform float blurSize; // blur kernel size
uniform vec2 texelSize; // size of a texel in the source texture

// varyings
varying vec2 vTexCoord; // texcoord of the fragment

// blur function
vec4 blur(sampler2D tex, vec2 texCoord, vec2 offset) {
    return 0.25 * (
        texture(tex, texCoord + offset) +
        texture(tex, texCoord - offset) +
        texture(tex, texCoord + vec2(offset.y, -offset.x)) +
        texture(tex, texCoord + vec2(-offset.y, offset.x))
    );
}

// main function
void main() {
    // sample the source texture
    vec4 color = texture(tex, vTexCoord);

    // calculate the bloom intensity
    float intensity = max(color.r, max(color.g, color.b));

    // apply blur if the intensity exceeds the threshold
    if (intensity > threshold) {
        vec2 offset = vec2(blurSize, 0.0) * texelSize;
        color = blur(tex, vTexCoord, offset);
        for (int i = 0; i < blurPasses - 1; i++) {
            color += blur(tex, vTexCoord, offset);
        }
        color /= float(blurPasses);
    }

    // output the final color
    gl_FragColor = color;
}

This implementation assumes that the following input is provided as a vertex attribute:

vTexCoord: the texture coordinate of the vertex

And that the following uniforms are provided:

tex: a texture containing the source image
[...]

PR #2725, opened 2022-12-24
High-Quality Bloom

MoritzBrueckner:
Requires https://github.com/armory3d/iron/pull/181.

A little christmas present, although it's just because I finally got time in the last days to finish this :)

This PR replaces the bloom effect with a higher quality implementation similar to the one used in Blender, Unity, etc. The main idea is to down- and upsample a series of mipmaps while filtering to approximate gaussian blur with relatively simple sampling patterns.

Bloom
[3D model "Christmas Ball" from Ivanix88](https://sketchfab.com/3d-models/christmas-ball-36868988285f4b60b3e389e1aa1bab3f) (CC BY 4.0), rendered in Armory

Since the algorithm closely resembles the one Blender is using\*, we can now use Blender's bloom settings (https://github.com/armory3d/armory/issues/1166#issuecomment-513069472) apart from tint color and manual clamping of bloom values. Implementing those two settings would be pretty simple, but I think they're not used often so for now I ignored them in favor of performance.

\* Actually it is quite difficult _not_ to match Blender's implementation because pretty much every modern bloom implementation seems to work like that, and Blender's bloom is just a slightly modified copy of the commonly used [KinoBloom](https://github.com/keijiro/KinoBloom).

bloom_settings

## Quality

Compared to Blender, Armory users now have three quality settings to choose from:
- **Low Quality**: uses a 4x4 box blur for downsampling and upsampling. This is what Blender is using for _downsampling_.
- **Medium Quality**: uses the "dual filter" (similar to "Kawase blur") described in ["Bandwidth-Efficient Rendering" by Marius Bjørge (Siggraph 2015)](https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-20-66/siggraph2015\_2D00\_mmg\_2D00\_marius\_2D00_slides.pdf) for downsampling and upsampling.
- **High Quality**: uses a 13-tap filter described in ["NEXT GENERATION POST PROCESSING IN CALL OF DUTY: ADVANCED WARFARE" by Jorge Jimenez (Siggraph 2014)](https://advances.realtimerendering.com/s2014/#_NEXT_GENERATION_POST) for downsampling, and a 9-tap (3x3) tent filter for upsampling. The 3x3 tent filter is also used by Blender for _upsampling_.

In most cases, the difference in quality is only barely visible, although there exist certain edge cases where it makes a difference (e.g. very small very bright areas).

**Old Bloom**
bloom_quality_old

**Low Quality**
bloom_quality_low

**Medium Quality**
bloom_quality_medium

**High Quality**
bloom_quality_high
(The high quality option is a little "blockier" than medium quality, but smoother)

In addition, I added an `Anti-Flicker Filter` option that if enabled applies [Brian Karis' weighted average](http://graphicrants.blogspot.com/2013/12/tone-mapping.html) filter to reduce fireflies (very bright pixels) that result in flickering due to working with discrete pixels instead of infinite resolution.

## Performance
Setup:
- Krom on Windows (D3D11)
- 1920x1080px (full-screen)
- GTX 750 Ti
- Performance measured with Intel GPA

Duration of the _full_ bloom pass:
| Old Bloom | Low Quality | Medium Quality | High Quality |
|---|---|---|---|
| ~1.41 ms   | ~ 0.64 ms  | ~ 0.77 ms | ~ 1.15 ms |

So the new bloom is not only nicer to look at, it's also faster even with the highest quality. The anti-flicker filter didn't make a measurable difference on my machine, although it should be slightly slower in theory.

### Remarks and further work
- I didn't make use of compute shaders because [Kha currently doesn't support passing render targets to compute shaders](https://github.com/Kode/Kha/issues/1406). The Kinc implementations are just empty at the moment (at least for D3D11). I tried to implement it, but I got crashes in the graphics driver so I rather quickly gave up for now. Once Kinc/Kha have better support for compute shaders we should make use of that to get rid of some more μs. There is a neat implementation of one-pass mip-chain downsampling by AMD ([source](https://gpuopen.com/fidelityfx-spd/)), whose principle we could maybe adapt to bloom downsampling.
- I added two general-purpose classes for downsampling and upsampling. This might be a little too over-engineered, but I wanted to remove some complexity from the render path code and other blur usages in Armory might profit from that. Please let me know if I should remove those classes again.
- The radius parameter needs to be available from Haxe, so I added a post processing data type to Iron's scene format. This is a bit unfortunate (because the radius could theoretically differ between scenes even though it's a render path setting and bloom is only implemented in Armory), but the only current alternative would be to get rid of the arm_config setting and always enable that to use the configuration file instead. I'm not sure whether we want that.
- The implementation of the 13 tap downsampling filter doesn't follow a [morton curve](https://en.wikipedia.org/wiki/Z-order_curve#Texture_mapping) for sampling values which in theory should make better usage of the cache I think. I couldn't measure a difference on my machine and my current GPU doesn't seem to have cache hardware counters so I decided for simpler to read code for now.
- Most other bloom implementations calculate the threshold before the first downsampling pass, but since that would require another context switch/shader pass I moved it after the first downsampling operation. I don't see a big visual difference, so I think it's acceptable.
- Some implementations don't use thresholds at all (e.g. [this one](https://www.froyok.fr/blog/2021-12-ue4-custom-bloom/)) because it is more physically accurate, but since Blender is using thresholds I'm sticking to that for now. The linked implementation also uses different texture addressing modes for upsampling/downsampling to get slightly better behaviour at the screen edges, but since this can only be changed in Iron at the moment this would require Iron to access Armory internals which obviously would break other applications like Armorpaint.
- There seems to be an issue with normals in Armory that sometimes result in NaN values. These will [propagate through the mipmaps](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@7.5/manual/Post-Processing-Propagating-NaNs.html) and cause screen flickering in those cases. This is _not_ a bloom issue, but it only really becomes visible through the bloom effect. We should try to fix it as soon as possible since it seems to happen quite often...

Closes https://github.com/armory3d/armory/issues/1166, and perhaps also https://github.com/armory3d/armory/issues/2612 (@Naxela, it's up to you whether you are satisfied with this implementation).

Merry christmas to everyone, and thanks @luboslenco and all other contributors for your hard work and support!


Naxela:
Hi @MoritzBrueckner, this looks great! - I haven't been able to test it out yet, but feel free to open up a claim on Bountysource related to #2612 🙂

luboslenco:
Wohoo!

rpaladin:
Hi @luboslenco. Super glad this PR got merged, however, I don't believe it's included in the new release? We were planning to announce/display it in this month's release notes as well (Christmas themed and all), and replace the notes splash with MoritzBrueckner's from above . If it's not too much trouble, would a re-release be okay?

PR #2726, opened 2022-12-24
Anchor and rotation option for draw nodes + New draw ellipse node

QuantumCoderQC:
1. Adds anchor point option for `Draw Rect` and `Draw Image` nodes. Option to rotate around the anchor point.
2. New draw ellipse node.

image


t3du:
cool now i know the trick to make the angle work thank QC!

PR #2741, opened 2022-12-31
Workspace config for code styles

rpaladin:
1. This PR helps Armory contributors uphold the code styles described in: https://github.com/armory3d/armory/wiki/Code-Style
2. I will add a small section to the mentioned wiki page if this PR is merged, to indicate that in order for the config to be recognized the armsdk folder must be stated as a VSCode `Workspace`.

<hr />

A few questions for you @ Lubos

1. If it's not too much trouble, and to avoid numerous PRs, would you mind opening/saving all of the armsdk files to update their indentations? There will be a _**extremely**_ large diff, so it might be better if you're the one doing it anyhow to avoid any potential mishaps. **NOTE:** You don't have to open/import all of the files in VSCode by-hand _(that would be insane)_. Instead, you could uncollapse all of the folders/files in your armsdk (if it's set as a `Workspace`), and select/drag all of them into VSCode, then use the `Save All` command. You could also automate this process using the terminal if you'd prefer instead. Or, if you'd like, I (or someone else) can do it.

2. We were also wondering if you wanted a mandated checkstyle process, for example like with [Iron's checkstyle](https://github.com/armory3d/iron/blob/main/checkstyle.json). However, this may cause unintentional diff changes. What do you think would be a good approach to teach contributors?

<hr />

Special thanks to @MoritzBrueckner for giving feedback / review.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+	"editor.detectIndentation": false,
+	"editor.tabSize": 4,
+	"editor.insertSpaces": false
+	"[python]": {
+		"editor.insertSpaces": true
+	},
+}


luboslenco:
1. Not a vscode user currently, if it's not too much trouble please go ahead, no prob with multiple PRs. 🙂
2. No need to mandate, we can re-format from time to time for now.

tong:
@rpaladin how about adding this as [editorconfig](https://editorconfig.org/) file instead?
Most editors have support for it.

rpaladin:
I'm not apposed to that idea, however, VSCode for example, does not support it natively: <https://github.com/microsoft/vscode/issues/56742> and requires the installing of an external plugin for the editorconfig support: <https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig>

PR #2744, opened 2023-01-01
Case Index node: fix accessing Vec4.equals() when DCE is on + cleanup

MoritzBrueckner:
Same fix as https://github.com/armory3d/armory/pull/2740 again, this time for https://github.com/armory3d/armory/pull/2743. While I was on it I also cleaned up the node's sources a bit (whitespace, unused variables/functions etc.) and took the liberty of changing the documentation as well as some socket names to hopefully make its usage a little clearer.

I didn't implement a replacement function for my changes by purpose (in fact, I even removed it since the node's version is still 1 so it is never called) because the node was only implemented a few days ago and isn't yet part of any SDK release.

### DIFF
--- a/Sources/armory/logicnode/CaseIndexNode.hx
+++ b/Sources/armory/logicnode/CaseIndexNode.hx
@@ -12,7 +12,7 @@ class CaseIndexNode extends LogicNode {
 		var value = inputs[0].get();
 
 		for (index in 0...inputs.length - 1) {
-			if (Std.isOfType(value, Vec4) ? value.equals(inputs[index + 1].get()) : (value == inputs[index + 1].get())) {
+			if (Std.isOfType(value, Vec4) ? (value: Vec4).equals(inputs[index + 1].get()) : (value == inputs[index + 1].get())) {
 				return index;
 			}
 		}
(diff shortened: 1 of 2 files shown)


t3du:
thanks the py file always struggles me, regarding the haxe file i thought it worked, i dont know what is dce on but it seems the comparison should be done in a new way then, ok.

MoritzBrueckner:
> i dont know what is dce on

DCE stands for [dead code elimination](https://haxe.org/manual/cr-dce.html) which is an optimization by the Haxe profiler to remove unused code. Armory only enables it for published builds (and only if the user hasn't disabled it).

In Armory's logic nodes, the return type of `inputs[n].get()` is typed as `Dynamic`, so the Haxe compiler doesn't know at compile time which type the value we try to compare has, and as such doesn't necessarily mark the vector methods as used. As a result, they get removed if they aren't used anywhere else, and at runtime when the dynamic type is resolved the required functions don't exist anymore. By using `(value: Vec4)` ([type check syntax](https://haxe.org/manual/expression-type-check.html)), the compiler is instructed to type `value` as `Vec4` and as such it knows that the used vector method must not be removed.

t3du:
wow thanks for the explanation, lot to learn yet

PR #2745, opened 2023-01-02
Updated tilesheet UI to support Sparrow atlas files.

MasterEric:
The current implementation of Tilesheets in Armory is very limited. It only provides support for grids of square sprites, and playing a range of frames from the grid in order. This is highly inefficient, as it does not allow the spritesheet to be optimized or frames to be reused.

This pull request implements user interface support for the Sparrow sprite format, which associates an XML file with the spritesheet, which looks like so:

```
<TextureAtlas imagePath="atlas.png">
 <SubTexture name="idle001" x="0" y="0" width="30" height="30"/>;
 <SubTexture name="idle002" x="30" y="0" width="65" height="78"/>;
 <SubTexture name="walk001" x="30" y="80" width="65" height="78"/>;
 <SubTexture name="walk002" x="30" y="80" width="65" height="78"/>;
 ...
</TextureAtlas>;
```

These spritesheets can be generated by programs such as TexturePacker or Adobe Animate.

Developers can then specify frames by prefix, for example the Walk animation will use the frames with the name `walk`, in order.

Care was taken to ensure only relevant options to the current atlas format display to the user, so the user is not prompted for a file when in Grid mode.

There is also room to add additional file formats as users request them.

## TODO

- [ ] Add additional options for Actions.
  - [ ] Make frame rate action-specific
  - [ ] Add the ability to specify an array of frame indices to use rather than a range.
  - [ ] Add the ability to horizontally or vertically flip the action.

## Note

This pull request is associated with armory3d/iron#182, and will not function as intended without it.

### DIFF
--- a/blender/arm/exporter.py
+++ b/blender/arm/exporter.py
@@ -14,6 +14,7 @@ http://creativecommons.org/licenses/by-sa/3.0/deed.en_US
 from enum import Enum, unique
 import math
 import os
+import pathlib
 import time
 from typing import Any, Dict, List, Tuple, Union, Optional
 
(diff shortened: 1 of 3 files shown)


EliteMasterEric:
As discussed in DMs, I have not been able to maintain this PR, but I grant the repository's maintainers full license over this code, such that they may take over development (so this work doesn't remain abandoned).

PR #2752, opened 2023-01-06
logic node: Get Object Collection

t3du:
image


t3du:
the first logic retrieved collections from blender but did not considered collections changes by armory nodes... i updated the logic to be consistent with the other collection nodes.

MoritzBrueckner:
Shouldn't this node return an array since an object can technically be part of multiple collections?

t3du:
Well then this at least return the first appearance. I could later explore this further, thanks for the comment.

PR #2753, opened 2023-01-07
logic node: Remove Trait from Object

t3du:
this node allows to remove a trait from a specific object (simil to the add trait to object node).

image

### DIFF
--- a/Sources/armory/logicnode/RemoveTraitObjectNode.hx
+++ b/Sources/armory/logicnode/RemoveTraitObjectNode.hx
@@ -1,0 +1,27 @@
+package armory.logicnode;
+
+import iron.object.Object;
+
+class RemoveTraitObjectNode extends LogicNode {
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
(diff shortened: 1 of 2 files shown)


Sources/armory/logicnode/RemoveTraitObjectNode.hx
```diff
@@ -0,0 +1,27 @@
...
+
+		assert(Error, object != null, "Object should not be null");
+		assert(Error, traitName != null, "Trait name should not be null");
+
+		var cname = Type.resolveClass(Main.projectPackage + "." + traitName);
+		if (cname == null) cname = Type.resolveClass(Main.projectPackage + ".node." + traitName);
+		assert(Error, cname != null, 'No trait with the name "$traitName" found, make sure that the trait is exported!');
+		
+		if(object.getTrait(cname) != null)
+			object.removeTrait(object.getTrait(cname));
```
MoritzBrueckner, line 23: Maybe consider writing the return value of `object.getTrait(cname)` to a variable so that it doesn't have to be computed twice.

PR #2758, opened 2023-01-13
Add player buttons to topbar

lampysprites:
This adds a compact launcher on the top of the blender window. 

Compared to Armory Player, the advantage is not needing to tab to render properties. Compared to key shortcut, it can terminate build, and (personally speaking) gives more obvious immediate signs of something happening than "building..." message in the viewport.

There could be a discussion about things that go in there. Possibly, an icon for build warnings?

* also if blender has a built-in stop icon, please can someone show it using a big red circle because i can't seem to find anything better than plane icon?

Aroa662w7j

### DIFF
--- a/blender/arm/props_ui.py
+++ b/blender/arm/props_ui.py
@@ -1376,7 +1376,7 @@ class ARM_PT_TopbarPanel(bpy.types.Panel):
         if state.proc_play is None and state.proc_build is None:
             row.operator("arm.play", icon="PLAY", text="")
         else:
-            row.operator("arm.stop", icon="MESH_PLANE", text="")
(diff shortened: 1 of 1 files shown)


t3du:
here are some options:

SEQUENCE_COLOR_01
image

CANCEL
image

lampysprites:
yep well no clear winner but the first one is a bit better i think? i guess it's more of a reminder this way. thanks!

rpaladin:
The `Icon Viewer` add-on that ships with Blender is a great tool for quickly finding available icons for Blender.

image

luboslenco:
Thanks for contributing!

PR #2761, opened 2023-01-17
Remove Object Node: remove object from collection

t3du:
The remove object node, removes the object from its collection only when that collection is assigned in blender. But if the collection is assigned within armory, the object remains in the collection despite being removed. This added code removes the object from the armory collection to keep it consistent with the other collection nodes.

### DIFF
--- a/Sources/armory/logicnode/RemoveObjectNode.hx
+++ b/Sources/armory/logicnode/RemoveObjectNode.hx
@@ -26,6 +26,11 @@ class RemoveObjectNode extends LogicNode {
 				#end
 			}
 		}
+		
+		var raw = iron.Scene.active.raw;
+		for (g in raw.groups) 
+			iron.Scene.active.getGroup(g.name).remove(object);
+		
 		object.remove();
 		runOutput(0);
 	}


MoritzBrueckner:
Is this really something the `Remove Object` node should do? Collections in Armory currently don't work the way collections in Blender do, the former are just abstract groups that you can use in any way you want. I personally understand the `Remove Object` node as a node that removes an object from the current scene (i.e. removes it from the scene hiearchy which you can see in the debug console outliner), but I think it shouldn't automatically remove the object from all collections.

What's the actual problem you are trying to solve with this? Doesn't the `Remove Object from Collection` node already do what you want to achieve?

t3du:
Yeah i think it should. The remove object is removing the object from the blender collection. But if that object is added to the that same collection using logic nodes it doesnt remove it. So is not consistent. Is true that knowing this i could use the node to remove the object from the collection, but i think that would be a work around. At least thats my opinion.

MoritzBrueckner:
Hi, sorry for my late answer, I'm rather busy at the moment.

I now realized that there isn't only abstract group data, but Iron also has a map of group names to actual object _instances_ [here](https://github.com/armory3d/iron/blob/bf2a628cf50a9b961a7289daa6adfc9ba722eb0a/Sources/iron/Scene.hx#L71), from which you remove the object (correctly, otherwise there would be memory leaked). However, shouldn't your new code be part of `object.remove()` then to not only clean up the memory with logic nodes? I'm not yet completely sure what the map is used for (it looks like it's mostly used by logic nodes), but it feels like something is still wrong (not with your code, but in general with handling this map)... Ideally, Iron would have a bunch of tests for this and a clear specification of how groups/collections are supposed to work.

Also, could you please elaborate what you mean with "Blender collection" in this case? The scene hierarchy in Iron is not a collection, and I can't find any code in `object.remove()` that removes the object from any collection (apart from your now merged changes in the corresponding logic node).

t3du:
My first idea was to include the code as part of the object remove, but i couldnt figure it out how this function removed the object from the collection. So i put the code to do that in the logic node in order to keep the consistency. I detected this problem because i had a bug in my scene which was solved after this fix. With blender collection i mean the object hierarchy defined in blender. Which is the same object collection that the collection nodes retrieves as a starting point. Maybe as you said there is a bigger problem underneath, like in the past when we had a similar situation with the remove parent which you solved. This is the way i found to solve it according to my expertise, sure you can do it better since you have a better understanding of armory itself.

PR #2766, opened 2023-01-22
Logic Node: Loop Continue

t3du:
image


MoritzBrueckner:
Hi, looks like you accidentally committed the wrong .py file (LN_loop_break.py which already exists), it's also in the wrong directory.

t3du:
yes, thanks for the comments, ill fix it.

t3du:
done!

PR #2767, opened 2023-01-22
Ss refraction

e2002e:
Hi, this is working as previously. I.E forward renderer not working, deferred rendered 's working fine.
Now I need you, @MoritzBrueckner, to test it once more because the results you showed me are abnormal.

I'll need some help with the forward renderer, path.drawMeshes is drawing the fragColor instead of writting it in the buffer.

Also restored the translucent shader.


e2002e:
Z issue solved.

rpaladin:
Hi. Please stop directly mentioning users, as well as spamming opening/closing/reopening pulls - it's very annoying for _everyone_. I do not mean to be rude, but perhaps it would be best to not attempt to contribute to areas you are unfamiliar with.

Better to focus on other areas that are easier to learn or you are more familiar with. Like new logic nodes or new Haxe API.

PR #2772, opened 2023-01-28
Logic Node: Random String

t3du:
This node could be used to generate random strings for numeric codes or alpha numeric codes... of a certain length.  

image

image

### DIFF
--- a/Sources/armory/logicnode/RandomStringNode.hx
+++ b/Sources/armory/logicnode/RandomStringNode.hx
@@ -1,0 +1,20 @@
+package armory.logicnode;
+
+class RandomStringNode extends LogicNode {
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function get(from: Int): Dynamic {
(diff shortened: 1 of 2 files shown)


MoritzBrueckner:
Hi, sorry I'm a bit late to the party :D

This doesn't look like it works correctly, the input socket is a string but the node expects `Array<String>`:

https://github.com/armory3d/armory/blob/28d82200c56e97fa3119335aeafd4761c3743d2f/Sources/armory/logicnode/RandomStringNode.hx#L11

Maybe it works on Krom and HTML5 because both strings and arrays can use `[]` (which might be why you didn't notice it) but I believe it will crash on static targets.

Also, depending on the string length, using a [`StringBuf`](https://api.haxe.org/StringBuf.html) for the string concatenation should give you better performance here. Little benchmark: https://try.haxe.org/#d03B0Fda.

t3du:
Thanks for the comments as always. Ill look it up and fix it.

PR #2781, opened 2023-02-04
Logic Node: Array Compare

t3du:
image

### DIFF
--- a/Sources/armory/logicnode/ArrayCompareNode.hx
+++ b/Sources/armory/logicnode/ArrayCompareNode.hx
@@ -1,0 +1,16 @@
+package armory.logicnode;
+
+class ArrayCompareNode extends LogicNode {
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function get(from: Int): Dynamic {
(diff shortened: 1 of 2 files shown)


MoritzBrueckner:
Is there a specific reason to compare the arrays by their string representation? This looks pretty inefficient and is potentially also somewhat unreliable since if I'm not mistaken Haxe might shorten the string representations of certain objects (you can probably test this by using object arrays). Comparing the arrays element-wise should be a better solution I guess, you also might want to return early if the lengths don't match.

t3du:
MoritzBrueckner, thanks for your comments as always. There is not an actual way of comparing arrays with nodes. Both Gate and Compare does not work. So in order to find a way to do it, after several attempts, I end up solving the problem using strings. Why strings, because I want the compare to work with arrays of more than one dimension, object arrays, an all the usual cases, that i tested and the code complies accordingly.  I agree with you that there should be a better efficient way to do it. Until that happens, at least we got this functionality in the meantime. The beauty of all this contribution is everyone can collaborate on anyone code and improve it.

PR #2782, opened 2023-02-04
Fix ArrayContains and ArrayIndex Nodes

t3du:
Both nodes returned -1 for vector items or array items (the case when the input array is an array of more than one dimension), even though those items did indeed exist in the array. That's because the indexOf function of  the array class does not handle the comparison properly.

With this fix now the comparison is handled and the correct value is returned for both nodes.


MoritzBrueckner:
Same as https://github.com/armory3d/armory/pull/2781#issuecomment-1416851502, I don't think you should use string representations for comparison of values (also regarding https://github.com/armory3d/armory/pull/2783). Depending on the amount of redundant code introduced by value comparisons, Armory might profit from a method `compareDynamic()` or `equalsDynamic()` that could then be used by various nodes.

t3du:
Well, in the meantime is not working at all. While the efficient code arrives, this can comply the comparison.

PR #2783, opened 2023-02-04
Logic Nodes: Array Count and Array Distinct

t3du:
Array Count Node: returns a 2D Array with every item count
Array Distinct Node: returns  both Array list of distinct and duplicated items

image

### DIFF
--- a/Sources/armory/logicnode/ArrayCountNode.hx
+++ b/Sources/armory/logicnode/ArrayCountNode.hx
@@ -1,0 +1,33 @@
+package armory.logicnode;
+
+class ArrayCountNode extends LogicNode {
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function get(from: Int): Dynamic {
+		var ar: Array<Dynamic> = inputs[0].get();
+
+		var values: Array<Dynamic> = [];
+		var values_list: Array<Dynamic> = [];
+		var count: Array<Int> = [];
+		var val_count: Array<Dynamic> = [];
+		
+		for(item in ar){
+			if(values.indexOf(Std.string(item)) == -1){
+				values_list.push(item);
+				values.push(Std.string(item));
+				count.push(1);
+			}
+			else {
+				count[values.indexOf(Std.string(item))] += 1;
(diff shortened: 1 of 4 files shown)


MoritzBrueckner:
How do the use cases for these nodes look like? This looks like highly specific functionality to me, which makes probably more sense to implement in a project when required (either by using other logic nodes or Haxe). But it's up to Lubos to decide this :)

Also, the code of the `Array Count` node (which took some time to understand, I think the node needs a more detailed explanation) can probably be simplified to the following:

```haxe
var counts: Map<String, Int> = new Map();

for (item in ar) {
	var itemRepr = Std.string(item); // <-- This avoids converting the item multiple times to string
	var itemCount = counts[itemRepr];
	counts[itemRepr] = itemCount == null ? 1 : itemCount++;
}

// Convert to array
var out: Array<Array<Dynamic>> = new Array();
for (key => value in counts) {
	out.push([key, value]);
}

return out;
```
(untested, and assuming you want to keep the string representations of an array item as keys in the output)

t3du:
im adding these nodes because ive been using arrays a lot with collections, since collections are arrays.  The cases to count, sort, distinct or duplicated items are all in the same frequent bag. Sure, as you said, this can be achieves with building the logic with other nodes, but this simplifies its use. I agree with you that more explanation could be good for a better understanding. As always if you dont think that these nodes are a contribution to armory, please dont merge them. I found them useful but that is my opinion and nothing more.

luboslenco:
> This looks like highly specific functionality to me, which makes probably more sense to implement in a project when required

There used to be a separate logic_nodes repository with extra logic nodes. Maybe we can revisit that idea and move off some of the more specific nodes there, with an option to easily download these extra nodes.

rpaladin:
> There used to be

There still is? https://github.com/armory3d/logic_pack

t3du:
I also thought the logic_pack could be an option... im not familiar whether is a very known directory for newcomers.

Besides, I was also wondering if later we could consider to reorganize nodes like the actual blender version with subfolders for every folder:
image

PR #2789, opened 2023-02-08
Ss refraction

e2002e:
Changes in reflect / refract functions to get the correct result.
I did get refraction to work on forward renderpath.
Screenshot from 2023-02-08 02-22-45

Also changed the depth testing mechanism in ssr_pass because it would try to show reflection from behind a mesh, resulting in strange artifacts.

Last issue is with half resolution (which I merged for both reflection and refraction) ; when in deferred renderer, it'll skip the reflections, as for forward render it will not render anything at all, complaining about a missing texture.

I'll look into that.


e2002e:
I fixed the half res issue.
Sreen space refraction + reflection are working with both renderer.
Yet the refractive meshes are not producing reflection with half res + forward render.

Screenshot from 2023-02-08 10-52-25
Screenshot from 2023-02-08 10-52-16

e2002e:
This is what i'm talking about:

Original reflection module:
Screenshot from 2023-02-09 11-04-42

Modified to not show hits behind visible object:


Screenshot from 2023-02-09 11-08-54

e2002e:
Reflection + refraction are doing güt.
Screenshot from 2023-02-09 12-28-31

Settings are not the same for refraction though, if step is below a certain point it'll render the reverse effect. Also min step much be much below 0.5 (which is supposed to be the ideal setting).

PR #2796, opened 2023-02-12
New copy to bundled operator/button

rpaladin:
Special thanks to @ QuantumCoderQC for review.

For those who wonder, I wrapped the code into a function, so that it's easier to implement code for copy/repath sound data-blocks in the future as well. Currently, to my knowledge, imported sounds can only be exported from Blender IF embedded in videos. So the exporting will need to be done using raw Python, but that's for another day. 🙂

image

Copy Override - False (First Log - mage Does Not Exist / Second Log - Image Exists)

image

Copy Override - True (First Log - mage Does Not Exist / Second Log - Image Exists)

image

### DIFF
--- a/blender/arm/props_ui.py
+++ b/blender/arm/props_ui.py
@@ -2324,49 +2324,50 @@ class ARM_OT_CopyToBundled(bpy.types.Operator):
 
         # Blend - Images
         for asset in data:
+            # File is saved
             if asset.filepath_from_user() != '':
                 bundled_filepath = project_path + '/Bundled/' + asset.name
                 try:
                     # Exists -> Yes
                     if os.path.isfile(bundled_filepath):
-                        # Override - > Yes
+                        # Override -> Yes
                         if (wrd.arm_copy_override):
+                            # Valid file
                             if asset.has_data:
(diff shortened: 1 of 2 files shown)


blender/arm/props_ui.py
```diff
@@ -2303,6 +2308,70 @@ def draw(self, context):
...
+                                log.error(f'Asset name "{asset.name}" has no data to save or copy, skipping')
+                                continue
+                        # Override - > No
+                        else:
+                            # Syntax - Yellow
+                            print(log.colorize(f'Asset name "{asset.name}" already exists, skipping', 33), file=sys.stderr)
+                            continue
+                    # Exists -> No
+                    else:
+                        if os.path.isfile(bundled_filepath):
```
MoritzBrueckner, line 2351: What's the point of this if statement? The entire else block around that is only executed if `bundled_filepath` does not exist, so it looks like this would always evaluate to false (assuming that no other process creates the file in between, which would "fail" with this code anyways).

rpaladin: Not sure why I added that. Removed it, thanks.

PR #2802, opened 2023-02-20
PlayActionFrom: add boolean for Reverse Action

t3du:
Requires: [iron:187](https://github.com/armory3d/iron/pull/187)

This modification allows the node to play the action backwards for both Object and Bone Animations. The functionality clones the action as actionReverse and reverses the frames. In that way the animation class works as usual.

image

### DIFF
--- a/Sources/armory/logicnode/PlayActionFromNode.hx
+++ b/Sources/armory/logicnode/PlayActionFromNode.hx
@@ -4,6 +4,7 @@ import iron.object.Animation;
 import iron.object.Object;
 import iron.Scene;
 import kha.arrays.Float32Array;
+import iron.object.ObjectAnimation;
 
 class PlayActionFromNode extends LogicNode {
 
@@ -91,7 +92,7 @@ class PlayActionFromNode extends LogicNode {
 				var oaction = null;
 				var tracks = [];
 
-				var oactions = animation.getOactions();
+				var oactions = cast(animation, ObjectAnimation).oactions;
 
 				for (a in oactions)
 					if (a.objects[0].name == actionR) isnew = false;
--- a/Sources/armory/logicnode/PlayActionFromNode.hx
+++ b/Sources/armory/logicnode/PlayActionFromNode.hx
@@ -3,6 +3,7 @@ package armory.logicnode;
 import iron.object.Animation;
 import iron.object.Object;
 import iron.Scene;
+import kha.arrays.Float32Array;
 
 class PlayActionFromNode extends LogicNode {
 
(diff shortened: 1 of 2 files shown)


QuantumCoderQC:
Hi,

Thanks for the PR. I personally think that reversing the track every time an animation is played is an overkill. Especially when the animation system already supports reverse animation playback. So this PR still does not solve the actual problem of why an error occurs when playing animation in reverse.

The error occurs here:
(https://github.com/armory3d/iron/blob/79551fe6e9c720d219dce6fcc6293ef04fbd616c/Sources/iron/object/Animation.hx#L108)

After a quick check. I just found out the `Play Action From Node` does work in reversing the animation of an armature. But only when the `Blend` is set to 0. So this setup works for armatures:

image

So, the issue seems to be that blending does not work when reversing. Hope this helps 😄

t3du:
The reverse action logic is do it only once. So if you trigger the play action node more than once that logic is not reversing the animation anymore, its using the actionReverse.

An that works for object animation too?

Because if so then we dont need this. Although this works with the blending.

Anyway in case we only want the part of the object animation i would need to make a small change.

t3du:
The reverse action logic is done only once. So if you trigger the play action node more than once that logic is not reversing the animation anymore, its using the actionReverse.

An that works for object animation too?

Because if so then we dont need this. Although this works with the blending.

Anyway in case we only want the part of the object animation i would need to make a small change.

t3du:
thanks QC it seems what you mentioned only works for that case: BoneAnimation without blending.

So in that way we could:

1. not merge the PR, this means not reversing for ObjectAnimation or BoneAnimation with blending.
2. modify the PR v1, this means to only add reversing for ObjectAnimation.
3. modify the PR v2, this means to add reversing for ObjectAnimation and BoneAnimation with blending.
4. merge the PR, this means the reverse option (although reverse is possible for BoneAnimation without blending).

What do you think?

QuantumCoderQC:
Thank you for the answer. In my opinion, this is an issue with iron and not armory. So, I think we must first find the solution to the root of the problem: why iron fails to reverse animations when it is supposed to. Then the issue is resolved automatically in armory as well.

While we can get away with this "workaround", in the future, if someone manages to fix the iron side of things, and we update the node accordingly, the behavior of the node could be different.

Of course, these are just my opinions.

t3du:
I completely agree. I dont know if someone is planning to fix the real issue, that considering in the future your new animation system could be implemented and this all will be obsolete. This is the way i could handle to manage the reverse situation in a complete way. In the meantime as a workaround I think is useful and clearer to the user because the reverse option is explicit.

Meanwhile lubos decides, im just gonna make a small change considering that now oactions is a public variable.

luboslenco:
Oki merging as a workaround, until the iron part gets fixed.

PR #2804, opened 2023-02-22
Gi

e2002e:
All you need for global illumination on average hardware.
Couldn't get the compute voxels to work (would be great when just relighting and not revoxelizing).
I added a feature that is simple for reflection - fineReflection - which is a reflection pass with aperture divided by two on top of the first raymach.
Screenshot from 2023-02-21 22-17-15
Screenshot from 2023-02-21 22-17-36
See with and without it.
Last but not the lightest issue, forward render doesn't handle roughness very well, ich weiss nictch warum.
I'm going to work on transparent shadows next. Take care.


e2002e:
I'm closing this for now. This will get this aspect once done, and there still will be the transparent shadows to deal with.
Screenshot from 2023-03-05 20-24-22

PR #2836, opened 2023-03-24
Node group improvements

QuantumCoderQC:
This PR adds many QOL improvements to the node groups.

1. Better UI for group nodes, indicating the number of users, fake user and a button to create a copy of current group:
image
Also removed the `New Group` button when there is an existing node group in use. It was sometimes accidentally clicked making it annoying.

2. Add shortcuts to the node group operations:
image

3. `Add call group node` operator now follows cursor location and adds the node there:
LNG Add

4. More improvements with the overall behavior of sockets when connecting/ disconnecting.

### DIFF
--- a/blender/arm/keymap.py
+++ b/blender/arm/keymap.py
@@ -28,6 +28,24 @@ def register():
     km.keymap_items.new("tlm.clean_lightmaps", type='F7', value='PRESS')
     arm_keymaps.append(km)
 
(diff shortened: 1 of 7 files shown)


e2002e:
@QuantumCoderQC
Hi, can you take a look at my pull request for refraction ?
There is an issue with armory group shader, it must take ior as input, but I did not find the code related to it.

rpaladin:
Please stop cross-posting in other peoples issues/pulls. It's rude.

PR #2850, opened 2023-04-11
Make Bump node work with arbitrary height inputs

MoritzBrueckner:
Previously, Armory ignored the "Height" input of the Bump node and instead only supported a predefined set of nodes as inputs for bump mapping. The first occurance of those nodes controlled how the bump output looked like, and if multiple of these nodes were combined (e.g. with a mix node), only one of them would be taken into account (that is, the output of the mix node is ignored in the example).

### Previous approach

My changes add support for arbitrary height inputs by changing how bump-related shader code is generated. Previously, the bump node would take the GLSL code that any of the supported input nodes generated and then inject coordinate offsets into that code in order to sample the values four times around the current texture coordinates to calculate a new normal vector. The offset was hard-coded and caused a blurring of the output (see images below) due to not matching the offset of a single fragment, and taking four additional samples was not really necessary (it only needs two additional samples to get good results).

Screenshot from Blender (the example material uses two wave pattern textures mixed together):
Blender screenshot

Screenshot from Armory, with PR changes:
Armory screenshot (with PR changes)

Screenshot from Armory, before this PR (only one wave texture taken into account for bump, also note the blurred look):
Armory screnshot (without PR changes)

Please ignore the seams in above pictures, they result from using badly tiled textures.

### New approach

The new approach works by evaluating the subtree connected to the "Height" input three times in what I called – for lack of a better name – "parser passes". There exist three passes: `REGULAR` (generate code to get the regular value of the tree) and `DX_SCREEN_SPACE`/`DY_SCREEN_SPACE` that instruct the nodes to output GLSL code that calculates the result of the subtree if the input coordinates are offset by a single fragment (= screen-space offset). By doing so, we dynamically adapt the distance between multiple samples to that of a fragment.

If the parser is in one of the `DX_SCREEN_SPACE`/`DY_SCREEN_SPACE` passes, it decides for each node whether the node's output might differ with respect to different fragments, and if that is the case, create new variables in the GLSL code with suffixes `_dx` and `_dy` and run the node's parsing function again. If the node generates code representing input values that vary between fragments (controlled via `ParserState.dxdy_varying_input_value`), Armory will compute screen-space offsets of the result and store them in the dx/dy variants. If a node does not need to be re-evaluated in the current shader pass, it will resort to the already parsed code of the `REGULAR` pass. So if you mix a bump map with a constant value (no matter how complex calculating this value is), the constant value will only be calculated once per fragment.

Once the passes are parsed, the bump node then manually calculates the screen-space derivatives for the height based on the pass outputs, and combines them with the screen-space derivatives of the fragment's world position to calculate new tangent/bitangent vectors that then are used to calculate the new normal vector. One could skip all this hassle and directly use `dFdx/dFdy` functionality for the height input, however, this leads to pretty pixelated outputs as even with the `dFdxFine/dFdyFine` variants (not available on all targets), the derivatives are calculated over a group of 1x2 pixels (also see https://github.com/armory3d/armory/pull/1952#issuecomment-716389033). So instead, we only take the screen-space derivatives of (usually) low-frequency values like the position or texture coordinates and calculate the height derivatives ourselves, this results in much better quality and this is also what Blender seems to do (see https://github.com/blender/blender/commit/ffd5e1e6acd296a187e7af016f9d7f8a9f209f87).

Generally, the output now pretty much looks like in Blender, although we use slightly simpler maths. Blender [does a bit more work to calculate the normals](https://github.com/blender/blender/blob/dd0cd1df1d1a5668ded98f4eca51021cbbccf97c/source/blender/gpu/shaders/material/gpu_shader_material_bump.glsl#L35-L43), and I don't really understand why it's needed (especially the determinant). Similar code can be found in [this paper](https://mmikk.github.io/papers3d/mm_sfgrad_bump.pdf), which seems to originate from the ShaderX5 article "Normal Mapping Without Precomputed Tangents" that is discussed [here](http://www.thetenthplanet.de/archives/1180). If anyone knows whether the current solution does not work correctly in some cases please let me know, but I think we're fine with the current approach and if there are any edge cases we will eventually find them :)

### Performance

In the best case, and presumably also the average case (e.g. `UV -> Image Texture -> Bump`), the new method is a little faster since we only need three instead of five total samples per fragment, but if the subtree that is connected to the "Height" input is *much more complex*, shader performance can become a problem. This should align with common sense that shader trees should be kept as simple as possible, and I will document this in the wiki once (and if) this PR gets merged.

---

The "passes" technique introduced in this PR might also help to solve another issue in the future ([problem description on Discord](https://discord.com/channels/486771218599510021/487613529990365184/999041327301922938) and related [Discord thread](https://discord.com/channels/486771218599510021/999349203765571614)) which happens if Armory is not able to infer tangent/bitangent vectors.

Thanks to Discord user @ Shakespeare for reporting this issue!

### DIFF
--- a/blender/arm/material/arm_nodes/shader_data_node.py
+++ b/blender/arm/material/arm_nodes/shader_data_node.py
@@ -1,13 +1,16 @@
 from bpy.props import *
-from bpy.types import Node
+from bpy.types import Node, NodeSocket
 
 import arm
 from arm.material.arm_nodes.arm_nodes import add_node
+from arm.material.parser_state import ParserState
 from arm.material.shader import Shader
 
 if arm.is_reload(__name__):
     arm.material.arm_nodes.arm_nodes = arm.reload_module(arm.material.arm_nodes.arm_nodes)
     from arm.material.arm_nodes.arm_nodes import add_node
+    arm.material.parser_state = arm.reload_module(arm.material.parser_state)
+    from arm.material.parser_state import ParserState
     arm.material.shader = arm.reload_module(arm.material.shader)
     from arm.material.shader import Shader
 else:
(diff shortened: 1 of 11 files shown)


luboslenco:
Amazing!

MoritzBrueckner:
Looks like I forgot to push a commit to this PR (just code cleanup), pushed it directly to main now: https://github.com/armory3d/armory/commit/6fc961015872a4c2341295206e629f5696d5afd1

PR #2854, opened 2023-04-17
Refraction

e2002e:
This is screen space refraction, working in both deferred and forward renderer.
Only bug is that when ssr (reflections) are set to half res, the refractive objects don't produce reflections, I'll look into that next week.
Apart from that there is no refraction of refractive objects, being a single post processing pass.

Screenshot from 2023-04-17 10-46-08
Screenshot from 2023-04-17 10-46-46
Screenshot from 2023-04-17 10-49-03

PS: I noticed that the arguments supposed to be passed to sampleLight() when SSRS is active are not present, the program will crash (had fixed that about a year ago though).
I'll do a pull request regarding that.

EDIT:
The reason why I renamed rpdat.rpvoxelao to rpdat.rp_voxels is because I'm sill working on the voxel module and will need the variable named as so. It's still a boolean, but will be a string for parsing beetween voxels AO/GI/Off


e2002e:
I'm unable to fix the helf res issue by myself, is someone can take a look that would be great.
Here is a picture of the issue it addresses:

Screenshot from 2023-04-19 16-03-37

By doing another depth test:

Screenshot from 2023-04-19 16-05-13

PR #2861, opened 2023-04-22
Refraction

e2002e:
This is a pull reques adding refraction, screen space effect.
Only issue is with ssr set to half res, in which case refractive objects are not reflected.
I will investigate this when I'm done with the clipmaps for the voxelgi module.


blender/arm/props_renderpath.py
```diff
@@ -280,6 +282,7 @@ class ArmRPListItem(bpy.types.PropertyGroup):
            description="A name for this item",
            default="Desktop")
 
+    rp_driver_list: CollectionProperty(type=bpy.types.PropertyGroup)
```
MoritzBrueckner, line 285: What's the reason for moving `rp_driver_list` to here? It was a world property before and is now a renderpath property (due to moving it), but other code still expects this to be a world property. I think you need to move this back to its original place.

Sources/armory/renderpath/RenderPathDeferred.hx
```diff
@@ -730,6 +751,97 @@ class RenderPathDeferred {
...
+
+				#if rp_material_solid
+				path.drawShader("shader_datas/deferred_light_solid/deferred_light");
+				#elseif rp_material_mobile
+				path.drawShader("shader_datas/deferred_light_mobile/deferred_light");
+				#else
+				voxelao_pass ?
+					path.drawShader("shader_datas/deferred_light/deferred_light_VoxelAOvar") :
+					path.drawShader("shader_datas/deferred_light/deferred_light");
+				#end
```
MoritzBrueckner, line 831: Is it really required to re-run the entire deferred lighting pass? This looks slow

MoritzBrueckner (commented):
The code looks mostly good to me now, although I can't really judge the algorithm (I've never implemented screen space refraction myself). The only thing I wonder about in terms of the algorithm is that the deferred_light pass is run twice now (see my review comment about this).

I'm also a bit confused about using multiple buffers for forward rendering (the `GBUF_IDX_` thing now also seems to be applied to forward rendering), but I guess it's ok.

I tried the refraction branch with a simple test file and I got the following error when exporting (both forward and deferred have this problem):
```
ERROR: [...]\build_RefractionTest\compiled\Shaders\Material_001_translucent.frag.glsl:57: 'sampleLight' : no matching overloaded function found
ERROR: [...]\build_RefractionTest\compiled\Shaders\Material_001_translucent.frag.glsl:57: '' : compilation terminated
```

Test file: [RefractionTest.zip](https://github.com/armory3d/armory/files/11304860/RefractionTest.zip)

Sources/armory/renderpath/RenderPathDeferred.hx
e2002e: Hi, thanks for reviewing.
When doing refraction, we need to render the scene without the refractive objects, having done so, we render the refractive meshes using the data from the first pass.
The reason why we are doing another deferred pass is because we want the refractive objects to be shaded.

I think it's ok to do so since the buffers are filled only with the data from the refractive objects on the second pass.

I'm going to fix the various errors this morning.

e2002e:
> The code looks mostly good to me now, although I can't really judge the algorithm (I've never implemented screen space refraction myself). The only thing I wonder about in terms of the algorithm is that the deferred_light pass is run twice now (see my review comment about this).
> 

The second deferred pass is run with buffers containing datas for refractive materials only, so that they don't only refract but also get shaded.

> I tried the refraction branch with a simple test file and I got the following error when exporting (both forward and deferred have this problem):
> ```
> ERROR: [...]\build_RefractionTest\compiled\Shaders\Material_001_translucent.frag.glsl:57: 'sampleLight' : no matching overloaded function found...

I had mingled with this function, because when using micros shadowing, ssrs, or voxel shadows, the return from sampleLight should apply once for adding light values, and another time for modifiying overall shade (just like the sun does).

Sources/armory/renderpath/RenderPathDeferred.hx
e2002e: > Is it really required to re-run the entire deferred lighting pass? This looks slow

Yes, objects must get shaded. Nonetheless, buffers only countaint data for the refractive objects. I don't know what the deferred shader does in this case but if there is a performance impact, we should add a mask and discard the fragment.

blender/arm/props_renderpath.py
e2002e: I don't know, my fork got messy due to me working on the main branch by mistake. I have deleted in from github and forked it again twice in the last 6 months.
Now this pull request is an orphan branch rebased on main (which is untouched).

e2002e:
Hi, the algorithm is not different from reflection, it even is based on the same file. There just are slight differences : in the shader, the viewpos is inverted, and there is a procedure to respect in order for the shader to work properly, that is render the scene before (without refractive objects) and then add the laters to the scene using the shaders and the appropriately filled gbuffers and saved background colors and depth.

The whole deferred procedure is redone, yet only fragments containing refractive objects are rendered for this second pass.

Im working on the voxels module, with clipmaping, will come up with a pull request on next week.

Güte Sonnertag.

e2002e:
@MoritzBrueckner
>I'm also a bit confused about using multiple buffers for forward rendering (the GBUF_IDX_ thing now also seems to be applied to forward rendering), but I guess it's ok.

I realized that ... I'm reinstalling Lunbuntu on my laptop and will fix that in the afternoon.

PR #2862, opened 2023-04-23
Fix performance issue when creating zoo collection for correct modifier export

MoritzBrueckner:
This PR fixes a performance issue that would happen in projects with scenes that have thousands of objects. Armory adds all mesh and empty objects to a temporary collection to make them visible to the depsgraph in order to apply modifiers (see https://github.com/armory3d/armory/pull/1498), but the code had at least quadratic complexity due to—for each object—calling `export_coll.all_objects.keys()` and iterating through the result (list). By using a _set_ of object names that we manually update, the complexity should be linear now.

The following profiler screenshot shows the impact of the fixed issue in a file with ~3800 objects (I had to force a single-threaded build to show everything in the profiler results):
Profiler screenshot

I guess the current approach of using a Zoo collection is still somewhat slow in complex scenes due to linking many objects, but I currently have no idea how to further improve this.

Thanks to Discord user @ Shakespeare for reporting this issue and for providing a test file :)

### DIFF
--- a/blender/arm/make.py
+++ b/blender/arm/make.py
@@ -165,13 +165,15 @@ def export_data(fp, sdk_path):
     # have a "zoo" collection in the current scene
     export_coll = bpy.data.collections.new("export_coll")
     bpy.context.scene.collection.children.link(export_coll)
+    export_coll_names = set(export_coll.all_objects.keys())
     for scene in bpy.data.scenes:
         if scene == bpy.context.scene:
             continue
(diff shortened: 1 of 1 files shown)


fatihmtlm:
I can confirm the result. I used to get ~8s on one scene and ~46s on the other one.
Now I am getting ~7.5s and ~8.5 respectively. The times are from blender console.

-I'm Shakespeare from the discord server

PR #2867, opened 2023-04-28
Improve and fix handling of tree variables

MoritzBrueckner:
This PR fixes/improves two issues reported by @ QuantumCoderQC on Discord:

- There was an error after duplicating a node tree with tree variables: When duplicating a node tree, Blender copies the original tree variable for which no master node exists in the copy and then calls `ArmLogicVariableNodeMixin.copy()` for each node. Because there is no master node in the copied tree, the previous implementation of `copy()` then assumed that the variable doesn't exist and in turn created a new variable without any node assigned to it (the copied nodes were still correctly assigned to the original variable copied by Blender).

   Now, there is a check in place whether the variable already exists. If so, we simply add the copied node to the existing variable, just like before if a node was copied normally between trees.

- While discussing the first issue with @ QC, we came to the conclusion that we should explicitly handle tree variable naming conflicts when ungrouping logic node groups since implicit merging of variables might not be wanted by the user in all cases (even if conflicts are pretty rare). Now, when there are naming conflicts when ungrouping a node group, a popup is displayed to the user in which they can then choose if a conflicting variable in the group tree should automatically be renamed to `<GroupName>.<VariableName>` (incl. accounting for potential conflicts of the renamed variable) or whether the variable should be merged with its counterpart in the enclosing tree:

   Screenshot: variable conflict popup

Ideally there would be unit tests for some of the above things as there are a bunch of potential edge cases, I hope I found all of them while testing the changes. @ luboslenco In case I (or someone else) would like to add some Python unit tests for important or error-prone behaviour in the future, would it be ok if those were added to a folder in the armory repo? I know that there is an armory_ci repo, but to me this looks like a single project for build testing, and we should probably version unit tests along with the corresponding code.

### DIFF
--- a/blender/arm/logicnode/arm_node_group.py
+++ b/blender/arm/logicnode/arm_node_group.py
@@ -1,18 +1,16 @@
-# Some parts of this code is reused from project Sverchok. 
-# https://https://github.com/nortikin/sverchok/blob/master/core/node_group.py
+# Some parts of this code is reused from project Sverchok.
+# https://github.com/nortikin/sverchok/blob/master/core/node_group.py
 #
 # SPDX-License-Identifier: GPL3
 # License-Filename: LICENSE
+from functools import reduce
+from typing import List, Set, Dict
 
 import bpy
-from bpy.props import BoolProperty, EnumProperty
-from itertools import zip_longest, chain, cycle, islice
-from functools import reduce
+from bpy.props import *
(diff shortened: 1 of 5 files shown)


luboslenco:
Thanks!

> would it be ok if those were added to a folder in the armory repo?

Sounds good!

PR #2879, opened 2023-05-19
Ss refraction

e2002e:
Screen space refraction.
There is a bug that is the half option being on, refractive objects won't produce reflections.

Can see it at work in [this video](https://www.youtube.com/watch?v=5Vup1NjGD8o)

screen


blender/arm/make_renderpath.py
```diff
@@ -307,8 +308,13 @@ def build():
             assets.add_khafile_def('rp_ssr')
             assets.add_shader_pass('ssr_pass')
             assets.add_shader_pass('blur_adaptive_pass')
-            if rpdat.arm_ssr_half_res:
-                assets.add_khafile_def('rp_ssr_half')
```
MoritzBrueckner, line 311: Is this removed by purpose? It looks as if this define isn't set anywhere else so if half res is enabled in the Blender UI the define will not be set.

Maybe this is related to the problem with half res that you've mentioned?

blender/arm/material/cycles_nodes/nodes_shader.py
```diff
@@ -106,7 +106,8 @@ def parse_bsdfprincipled(node: bpy.types.ShaderNodeBsdfPrincipled, out_socket: N
         # clearcoar_normal = c.parse_vector_input(node.inputs[21])
         # tangent = c.parse_vector_input(node.inputs[22])
     if state.parse_opacity:
-        if len(node.inputs) >= 21:
+        state.out_opacity = c.parse_value_input(node.inputs[21])
+        if len(node.inputs) >= 21: #do we need to test that ?
```
MoritzBrueckner, line 110: I think the check is for compatibility with old node versions (Blender doesn't automatically update them as far as I know...)

MoritzBrueckner (commented):
Hi, sorry for taking so long to write you again. The code looks pretty solid now! I hope you'll find out what causes the forward and half res issues you've mentioned.

Maybe it's possible to integrate the `gbufferD1` render target into one of the currently unused channels of the `gbuffer_refraction` render target, this way we can probably use less VRAM. Unfortunately Kha doesn't provide a lot of texture formats, maybe `RGBA32` could work as well for `gbuffer_refraction` without loosing too much precision?

e2002e:
Hi, no worries Moritz. I actually just fixed a bug.. That is normalization
of the refract function. It would have scale issue otherwise. As for the
forward renderer, it's mandatory to use a texture, since the shader is
deferred just like reflection. It could be possible to make each object
refract independently, meaning writing the opacity and ior into the shader
attached to the object and render according to the distance. That procedure
would actually make refraction of refrative object possible. But that is
too much work für mich.

Le dim. 21 mai 2023 à 23:47, Moritz Brückner ***@***.***> a
écrit :

blender/arm/make_renderpath.py
e2002e: Ok that's fixed.

blender/arm/material/cycles_nodes/nodes_shader.py
e2002e: verstanden

e2002e:
> Hi, sorry for taking so long to write you again. The code looks pretty solid now! I hope you'll find out what causes the forward and half res issues you've mentioned.
> 
> Maybe it's possible to integrate the `gbufferD1` render target into one of the currently unused channels of the `gbuffer_refraction` render target, this way we can probably use less VRAM.

Yes possible, 32bits would be used for the buffers (if depth texture is really this size), and two short integers (opacity, ior) make the other bits.

Unfortunately Kha doesn't provide a lot of texture formats, maybe `RGBA32` could work as well for `gbuffer_refraction` without loosing too much precision?

I had issues with RGBA32, the refraction's textures need to be RGBA64. I experienced that issue for the voxels also, no colors were displayed in the foremost case. It seems there is an issue with kha's RGBA32 overall.

e2002e:
Hi there: @MoritzBrueckner @luboslenco @rpaladin (just pigging to get attention).
I decided to get to fix the forward renderer today, and to my surprise it was working ! I might have messed up my previous test files, or maybe drivers / OS issue.

You can see the shader at work in [this video](https://youtu.be/4FmyQnJfsSA)

e2002e:
I'm done with this, you should merge it. I'm trying to get the clipmaps working for the voxels module now.

e2002e:
@MoritzBrueckner. 
I just patched the reflections / refractions because I noticed that when rendering big scenes, reflections would be clamped to a very short distance regardless of the search option. The reason for that is that the reflect / refract functions return a normalized value which is divided by the step factor, and used in the rayCast function to iterate, but the loop being to small (maxSteps is set to 18) the rays will never hit objects which are far away.
The solution to this problem is to multiply maxSteps by the search variable, and since we'r at it, normalize it by the step size. So maxSteps = (1.0 / ssrRayStep * ssrSearchDist).

See it in action (also this screeshot is from the voxelgi fork).

Capture d’écran du 2023-07-13 21-38-27

PR #2883, opened 2023-05-31
Fix compilation of draw nodes if "Zui" setting is set to "Auto"

MoritzBrueckner:
This PR fixes compilation of the draw nodes that use the GraphicsExtension class if Zui export is set to "Auto". It is more of a last-minute workaround for SDK 23.06 than a proper fix, so we should revisit this in the future and either remove the "Auto" setting (simple solution but this would always export the default font, code-wise DCE will take care of not increasing the output size too much) or maybe implement an export callback for each node such that a node could register its need for Zui classes.

@ luboslenco do you have any preference regarding a proper solution?

### DIFF
--- a/Sources/armory/logicnode/DrawArcNode.hx
+++ b/Sources/armory/logicnode/DrawArcNode.hx
@@ -3,7 +3,9 @@ package armory.logicnode;
 import kha.Color;
 import armory.renderpath.RenderToTexture;
 
+#if arm_ui
 using zui.GraphicsExtension;
+#end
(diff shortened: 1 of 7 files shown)


luboslenco:
> @ luboslenco do you have any preference regarding a proper solution?

Would like to eventually switch armory to using [armorcore](https://github.com/armory3d/armsdk/tree/armorcore), which should make the overall maintenance much easier. It's a rather big change so will make sure to post more about it first.

PR #2895, opened 2023-06-20
Add support for Bullet physics debug drawing

MoritzBrueckner:
This PR adds debug drawing support for Bullet physics. It requires https://github.com/armory3d/haxebullet/pull/42 where I explained some more technical details.

Collider Wireframes

Contact Points

Ragdoll Constraints

The debug drawing settings can be found in the `Rigid Body World` panel of the scene properties:

grafik

On the Haxe side there are more debug flags available, however, these do not take effect in Armory since Bullet does not use them in the library core, instead they seem to be used in individual (and sometimes obsolete) Bullet examples only (urgh...). Some flags have dependencies between each other, the wireframe flag even leads to drawing the suspension of raycast vehicles for whatever reason (Bullet's codebase truly is something else... :D). In some cases I have put some explanations for the unused flags as comments to the code.

Tested on krom, html5 and windows-hl.

## Potential future work
- Implement proper clipping for drawing lines. Currently lines aren't rendered if any end is out of the far/near camera range.
- Investigate whether we can implement debug drawing for Oimo physics, if we still plan to support it. @luboslenco What was the reason for adding it in the first place? Does it have some features that Bullet hasn't?
- We could merge the line drawing with the debug console's line drawing, but then we need to decide whether we want to use the same kind of line drawing (2D vs. 3D) for both.
- Perhaps improve the drawing of the contact points, there are probably better ways to vizualize these than using diagonal crosses (+ normal vector).
- Expose the debug drawing flags in the debug console
- If someone knows a way to better align the checkboxes in the Blender UI and still have a good-looking margin to one of the panel sides (which does not exist if `layout.use_property_split == False`), please let me know :)

Thanks to all involved on the Discord server with providing feedback and ideas!

### DIFF
--- a/Sources/armory/trait/physics/bullet/DebugDrawHelper.hx
+++ b/Sources/armory/trait/physics/bullet/DebugDrawHelper.hx
@@ -1,0 +1,224 @@
+package armory.trait.physics.bullet;
+
+import bullet.Bt.Vector3;
+
+import kha.FastFloat;
+import kha.System;
+
+import iron.math.Vec4;
+
+#if arm_ui
+import armory.ui.Canvas;
+#end
+
+using StringTools;
+
+class DebugDrawHelper {
+	static inline var contactPointSizePx = 4;
+	static inline var contactPointNormalColor = 0xffffffff;
+	static inline var contactPointDrawLifetime = true;
+
+	final physicsWorld: PhysicsWorld;
+	final lines: Array<LineData> = [];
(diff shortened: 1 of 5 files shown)


luboslenco:
> What was the reason for adding it in the first place?

I think it was not about features but due to being written in Haxe and being more lightweight.

tong:
Awesome!

PR #2904, opened 2023-07-08
Add editorconfig

tong:
Adds an [editorconfig](https://editorconfig.org/) following #2741 to support various editors ([outofthebox](https://editorconfig.org/#pre-installed)), not only vscode.

Imho, if accepted, `armory/.vscode/settings.json` should be removed, as it's non-standard.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+tab_width = 4
+indent_style = tab
+
+[*.py]
+indent_size = 4
+tab_width = 4
+indent_style = space


rpaladin:
I'm not against the idea of supporting other text-editors, but I'm politely hesitant to the idea of removing the VSCode settings. I'm aware me saying this as the person who originally implemented them may sound a bit lame, but I would like to point out that VSCode is used by many and has included settings in multiple repositories including Kha, as well as many prominent frameworks/software companies: https://github.com/search?q=.vscode&type=code

tong:
The fact that vsode still does not support editorconfig without a third party plugin makes it look like microsoft is going after vendor lock-in instead of well established, open standards.
Supporting such tactics don't do any good in the long run and the 'prominent' framework/software excuse is exactly this at work.

PR #2905, opened 2023-07-10
Voxelgi

e2002e:
This pull request is rebranding the old voxel global illumination module from 19.02 (last version where the module is still present).
Changes are:
- All light types are taken into account for shadowmapping.
- Voxelization is clipmapped

Clipmaps means that regions size increase exponentially, making it possible to cover huge areas without performance impact since voxels size increase with distance from the viewer.
Ambiant Occlusion benefits from the clipmaps too.

I successfully tested it with a decent size landscape, the whole area is getting indirect light and there doesn't seem to be any miscomputations.
Note that I couldn't get a correct effect on a large flat glossy plane (supposed to represent a lake at the  foot of the mountains). Maybe this is due to reflections / refractions using only one cone, or because surface was completely flat with roughness set to 0.

I'll do more tests soon (when Kha or armory.py is fixed https://github.com/armory3d/armsdk/issues/65).

This is a link to the forum's thread I had opened: https://forums.armory3d.org/t/voxel-gi-with-clipmaps/5242

I'm also opening a pull request for iron.

Credits for the procedures goes mostly to WickedEngine from which I read a lot of code, and also (have to admit) to chatgpt who's willing (after a lot of questioning) to yield some relevant corrections.


e2002e:
Closing it, still have issues to deal with.

e2002e:
Still needs work, closing.

PR #2914, opened 2023-08-05
Ss refraction and fix for SSR

e2002e:
This pull request is implementing screen space refraction and fixing an issue with the reflection shader.

Capture d’écran du 2023-08-13 14-50-27

Here you can see various glass objects reflecting and refracting.
As you can see, the cube is appearing on the other side of the glass like in real life.

@MoritzBrueckner, @luboslenco.
Concerning the reflection fix, the issue is that when using a low step for accuracy, the raytracing function doesn't fit to the search distance because maxSteps are hardcoded.
The solution is this:`
#define maxSteps int(ceil(1.0 / ssrRayStep) * ssrSearchDist)
That way, we can reflect objects far away from the camera:

Capture d’écran du 2023-08-13 15-10-35

Concerning the refraction shader. We use it pretty much like the reflection's, but the procedure is a bit more complex:
We need to render the refractive objects independently from the rest of the scene, so saving the scene depth and color before rendering the refractive objects, first alone with all shading passes, and then do the refraction and mix both according to the opacity.
It's possible to get tinted refractive objects and they do reflect too because refraction's before reflection in the renderpath.
Yet the limitation is that it's not possible to refract a refractive object, because it's a single pass. Doing this procedure in forward render (attaching the refraction shader to each mesh) and rendering by distance from farther away to close would solve this problem.
Only issue is that when ssr half res is on, refractive objects do not reflect.


e2002e:
There is a bug with refraction that I can't deal with:
When ssr is set to half resolution, refractive objects don't reflect.

e2002e:
Will try to fix #2921 and reopen.

PR #2920, opened 2023-08-25
better param-bindings and upgrading to formula haxelib version 0.4.2

maitag:
Only the haxe-part at now but there is also some errorparsing-glitches at the python-side,
so functions like "cos" and "sin" etc. can not be used inside the math-expression:

blender/arm/logicnode/math/LN_math_expression.py

I am no py-expert so i am commented out all inside `def set_exp(self, value):` for testing.
(Formula-lib also gives good error-msg at runtime if there is something wrong into the math-expression or a missing variable etc.)

(^_^)

### DIFF
--- a/Sources/armory/logicnode/MathExpressionNode.hx
+++ b/Sources/armory/logicnode/MathExpressionNode.hx
@@ -1,7 +1,1178 @@
 package armory.logicnode;
+
 import haxe.io.Bytes;
 import haxe.io.BytesInput;
 import haxe.io.BytesOutput;
+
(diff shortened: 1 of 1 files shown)


maitag:
by the way, wouldn't it be better to put all inside `formula` folder into here:
Sources/armory/math
?

luboslenco:
Thanks!

> by the way, wouldn't it be better to put all inside formula folder into here:
Sources/armory/math ?

Probably good enough as it is for now.

PR #2922, opened 2023-08-26
voxel AO & GI + clipmaps.

e2002e:
Here it is !
This pull request is reimplementing the old voxel global illumination module (rebased from 19.02).
Here is a [link to the forum's post](https://forums.armory3d.org/t/voxel-gi-with-clipmaps/5242).

The main change is the use of clipmaps (roughly put, adaptive voxel size) so that we can process distant objects.

I came across a lot of articles and videos showing new voxelization techniques, I tried to read the official pdfs, but they are obscure and designed for scholars. So instead I read other people's code:
https://github.com/turanszkij/WickedEngine
https://github.com/compix/VoxelConeTracingGI

Eventually I'm coming with the most basic voxelization we can get, but it's fast and stable.

True reflections / refractions:
Capture d’écran du 2023-08-26 11-51-50

Distant voxelization:
Capture d’écran du 2023-08-26 11-36-34
(shitty framerate comes from the terrain being highly subdivided)

There is a major architectural change in the pipeline ; instead of computing the voxels position in the voxelizer and once again prior to passing the value to the conetracer, we write it in a buffer when voxelizing. That way it doesn't change value (because there is a delay) and we avoid artifacts and lod transition seams.

A last bug I couldn't deal with is that when using GI in forward render path, voxels coordinates have a wrong scale.
I'll be working on that again soon, and also implement a second bounce to be able to reflect / refract indirect light and finer illumination.

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -8,6 +8,9 @@
 #ifdef _Irr
 #include "std/shirr.glsl"
 #endif
+#ifdef _VoxelGI
+#include "std/conetrace.glsl"
+#endif
 #ifdef _VoxelAOvar
 #include "std/conetrace.glsl"
 #endif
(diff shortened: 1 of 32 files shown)


e2002e:
I removed the clipmap count as an option since the level increases naturally with the distance and we have not use for the variable.
Note about the bug and todo:
Forward render path and GI has wrong voxel's coordinates, and the bounces options doesn't yet work.
I need a way to get texture coordinates for the bounces shader, let me know if you can provide any insight.

e2002e:
closed as non working

PR #2923, opened 2023-08-26
Fix too bright environment and barely visible shadows if irradiance is off

MoritzBrueckner:
When irradiance was turned off (either globally or per material), the environment light that's added to the output fragment color was initialized with 1.0 which resulted in a fully-lit scene. This PR sets it to zero as it seems more natural. The default case with irradiance turned on is not touched by this PR.

Before:
Screenshot

After:
Screenshot

@e2002e I noticed that you also work with `envl` in your current voxel AO/GI work, please let me know in case this breaks anything for you :)

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -233,7 +233,7 @@ void main() {
 		if (g2.b < 0.5) {
 			envl = envl;
 		} else {
-			envl = vec3(1.0);
+			envl = vec3(0.0);
 		}
 	#endif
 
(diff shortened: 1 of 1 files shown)


e2002e:
Hi Moritz !
VoxelAO darkens the envl and then sets the fragColor to envl.
VoxelGI just adds envl to the fragColor.
So doesn't break anything.
I know this is not the place to discuss but if you could drop by my pull request to see my latest post let me know if any solutions.
Bye.

MoritzBrueckner:
> I know this is not the place to discuss but if you could drop by my pull request to see my latest post let me know if any solutions.

Yeah, sorry for always taking so long to look at your PRs. I'm a bit busy in the next days but I will try to answer as soon as possible :)

e2002e:
Hi. Kein problem.
Actually I have closed the pull request for now because there where issues I didn't realize. This is pretty much done and I just need to fix the seams between lod now. It's probably going to take a week for me to reopen.
Gute abend.

PR #2924, opened 2023-08-27
Fix for ssr search dist

e2002e:
This is a simple tweak to the ssr shader that allows to match the search distance.

In the rayCast loops on maxSteps which is an hardcoded, if the step size is too small we don't reflect far:
Capture d’écran du 2023-08-27 18-09-35

To match the searchDist we need to take the stepSize and search dist into account when setting the maxStep value:

#define maxSteps int(ceil(1.0 / ssrRayStep) * ssrSearchDist)

Capture d’écran du 2023-08-27 18-09-10


e2002e:
This is included in #2925

e2002e:
Closed as fixed by #2927.

e2002e:
I'm reopening this as you might want to work step by step.

e2002e:
@luboslenco Can you confirm the issue I'm showing here is present in the master branch ?

MoritzBrueckner:
Looks good to me, but what's the reason you replaced `const` by `#define`? Defines are literally replaced in the code by the preprocessor, so instead of

```glsl
for (int i = 0; i < maxSteps; i++) {
```
the compiler sees
```glsl
for (int i = 0; i < int(ceil(1.0 / ssrRayStep) * ssrSearchDist); i++) {
```

Have you checked whether the compiler is able to optimize that so that the defined expression is only evaluated once? Otherwise using another solution might be faster. If `const` doesn't work because the value is not perceived by the compiler as being constant, you could just calculate it in Python and [include the resulting value in `compiled.inc` like `ssrRayStep` and `ssrSearchDist`](https://github.com/armory3d/armory/blob/36ffc12ea5b7e38c8fe4511be39352bdb381cb17/blender/arm/write_data.py#L660-L662).

e2002e:
You'r right I didn't think about glsl optimizations (if there are any is this case). So here i rolled back the value type.

e2002e:
@MoritzBrueckner I realized I have missed the point about variable optimization.
Yes using a pragma will recompute the values for each fragment at each frame instead of having a constant stored.

e2002e:
@MoritzBrueckner, is that possible to get your mail contact ?
There are some points I need to discuss with someone who has more experience than me with armory3d.
I don't know if my email appears on my profile, if it doesn't please let me know. Or maybe we can open a discussion here ?

e2002e:
@luboslenco can you merge than ? It's reviewed by @MoritzBrueckner  and fixes a bad bug !

PR #2927, opened 2023-09-12
Screen space refraction

e2002e:
This pull request is implementing screen space refraction.
Limitation is that refractive objects don't refract.
There is a bug with half res and screen space reflections, that refractive objects won't reflect.

I included the fix for which I have been opening a pull request:  #2924


e2002e:
@MoritzBrueckner possible to review please ?

PR #2932, opened 2023-09-23
Voxelgi clipmaps

e2002e:
This pull request implements voxel gi with clipmaps !

clipmaps is a mechanism that voxelizes the scene with voxels that have exponential size depending on the distance to the camera.
This way we have grids that look like an hypercube.
It allows for voxelizing of distant objects with still detailed lighting at close distance.

The second bounce option is not working (yet)

Link to the forum's thread:
https://forums.armory3d.org/t/voxel-gi-with-clipmaps/5242/48

Bugs:
Temporal filter is not working with reflections / refractions, so I've limited the effect to diffuse light, I'm trying to debug that atm.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,8 @@
 {
 	"editor.detectIndentation": false,
 	"editor.tabSize": 4,
-	"editor.insertSpaces": false
+	"editor.insertSpaces": false,
 	"[python]": {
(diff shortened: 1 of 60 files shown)


e2002e:
I closed this because there is still work todo:
- Enable the second bounce
- Fix temporal filter for specular and refraction
- use emissive material

AO is fully working.

PR #2934, opened 2023-09-27
Fix node update routines for Gate and Array Float nodes

MoritzBrueckner:
Both nodes had wrong out-of-range checks for version numbers, the Gate node did update to an old version, and the Array Float node did not copy any input sockets.


rpaladin:
Sorry about that..

MoritzBrueckner:
No problem :) The node replacement system is pretty messy and sometimes difficult to understand anyways...

PR #2940, opened 2023-10-04
Recast Navigation System Overhaul

QuantumCoderQC:
## Overview:
This PR is a complete overhaul of the recast navigation system that solves a few issues and introduces many new features.

### Dynamic generation of Navmesh: 
The navigation mesh is generated at runtime unlike before, where it had to be exported as an `.obj` file and then created from from file. The new version creates the navigation mesh with object geometry info from iron. This also comes with an option to include immediate child meshes into the navmesh build.

### Crowd System:
Previously, the `NavAgents` would not have any awareness of the other agents. This caused agents to not collide with each other and intersect. The newly implemented crowds system solves this issue. Now the `NavCrowd` agents will be aware of the neighbors and try to remain separated while navigating to the goal. This does not remove the `NavAgent` trait. It is still usable for basic path finding without the crowds system.

### Temporary Obstacles:
The new module includes an option to create `tiled navmesh`. This enables for the possibility of adding and removing temporary obstacles with a cylindrical or cuboidal shape. The navigation `NavAgents` and `NavCrowd` agents obey the obstacles and dont pass through them. This opens new possibilities for a dynamic navmesh based games. The PR also includes a Logic Node to add cylindrical obstacles.

### Runtime Debug View:
Navigation mesh generated can be viewed at runtime if the `debug view` option is enabled for the navmesh.

### Support for Haxe HL targets:
This PR also includes support for navigation system on all HL targets. Only JS targets were supported before.

## Breaking Changes:
Since the new navigation system is a complete overhaul. there are a few breaking changes.

* The `NavMesh` and `NavAgent` bundled traits must be removed, re-added and re-configured if already being used in a game.
* There is no `Preview` of the built navmesh in Blender. Unlike before, the navmesh is generated at runtime. So this option of preview in Blender is removed. However, to make up for this, runtime debug view of navmesh is supported.

## Dependencies:
The new recast navigation system requires https://github.com/QuantumCoderQC/haxerecast instead of the currently used https://github.com/armory3d/haxerecast. These repositories do not share a common history.

## Demo:
Here is a small demo of the new recast system with temporary obstacles and navigation crowds system:
https://quantumcoderqc.github.io/RecastNavigationDemo/html5/index.html

Thanks to everyone on discord for their feedback and testing. Thanks to @ 1k8 for testing on Linux HL.

### DIFF
--- a/Sources/armory/logicnode/CrowdGoToLocationNode.hx
+++ b/Sources/armory/logicnode/CrowdGoToLocationNode.hx
@@ -1,0 +1,34 @@
+package armory.logicnode;
+
+#if arm_navigation
+import armory.trait.navigation.Navigation;
+#end
+
+import iron.object.Object;
+import iron.math.Vec4;
+
+class CrowdGoToLocationNode extends LogicNode {
+
+	var object: Object;
+	var location: Vec4;
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function run(from: Int) {
+		object = inputs[1].get();
+		location = inputs[2].get();
+
+		assert(Error, object != null, "The object input not be null");
+		assert(Error, location != null, "The location to navigate to must not be null");
+
+#if arm_navigation
+		assert(Error, Navigation.active.navMeshes.length > 0, "No Navigation Mesh Present");
+		var crowdAgent: armory.trait.NavCrowd = object.getTrait(armory.trait.NavCrowd);
+		assert(Error, crowdAgent != null, "Object does not have a NavCrowd trait");
+		crowdAgent.crowdAgentGoto(location);
+#end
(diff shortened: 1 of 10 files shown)


QuantumCoderQC:
Hi @luboslenco ,

Thank you for merging this PR. I don't know if you took note, but this PR also needs the new `haxerecast` library from the dependencies I mentioned above. Please replace https://github.com/armory3d/haxerecast with https://github.com/QuantumCoderQC/haxerecast.

QuantumCoderQC:
@luboslenco Sorry to bother you again. Could you also please update the haxerecast submodule at https://github.com/armory3d/armsdk please. 

Thanks you!

luboslenco:
Done! 👍

PR #2955, opened 2023-10-22
Wait For All Inputs Node

t3du:
This node waits for all inputs to trigger the output once.

image

### DIFF
--- a/Sources/armory/logicnode/WaitForNode.hx
+++ b/Sources/armory/logicnode/WaitForNode.hx
@@ -11,7 +11,7 @@ class WaitForNode extends LogicNode {
 	override function run(from: Int) {
 
 		if(!froms.contains(from)) froms.push(from);
-		if(inputs.length == froms.length) runOutput(0);
+		if(inputs.length == froms.length){ runOutput(0); froms = []; }
 
 	}
 
--- a/Sources/armory/logicnode/WaitForNode.hx
+++ b/Sources/armory/logicnode/WaitForNode.hx
@@ -1,0 +1,18 @@
+package armory.logicnode;
+
+class WaitForNode extends LogicNode {
+
+	var froms: Array<Int> = [];
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function run(from: Int) {
+
+		if(!froms.contains(from)) froms.push(from);
+		if(inputs.length == froms.length) runOutput(0);
+
+	}
+
+}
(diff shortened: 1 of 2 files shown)


QuantumCoderQC (changes requested):
Hi, currently there is no reset of the `froms` array. Which means this node, technically, works only once. After that, I think the output is triggered at any input. That brings the question if we need a reset input? If the node intends to work only once, I think there is not much of a point to it. In any case, the documentation must also be improved a bit. Things such as if the order of inputs matters or not and so on.

t3du:
Hi. Thanks for you comment I need to fix the reset option. I'm thinking in terms of sequential and parallel options. Using the example in the image, I know if I want to print 1, 2, 3 and 4 before printing 'wait for all' i could have done a sequential flow. But if a want to print in a parallel way 1, 2, 3, 4 and then print 'wait for all', i can't use a merge because it will be trigger more than once. So in that way this nodes only moves forward when all the inputs are executed. That is a basic example, but instead of several prints it could be different branches of code. And yes, The name and the documentation needs to be very clear about how it works.

MoritzBrueckner:
Regarding the documentation of the node, I think "when all inputs are received" is too imprecise, even in the new version of the documentation it sounds like the node only activates its output if all inputs are active at the same point in time.

What about the following (already including the `Reset` input):
```python
class WaitForNode(ArmLogicTreeNode):
    """
    Activate the output when any input is active _and_ all inputs apart from `Reset`
    have been activated at least once since the last reset of the node or the node's initialization.

    @input Reset: Reset the node to its initial state, requiring all other inputs
        to be activated at last once again in order to re-enable the node's output.
    """
```

The "when any input is active" part is important because the node currently re-activates the output if all inputs were already active and any input is activated again. If you change this so that the node can really only activate the output once (unless it's reset) this needs to be changed of course.

Technically, everything following ", requiring" in the reset input description is redundant information already stated in the node's description, but it's probably a bit clearer this way.

I left out the description for the other inputs and the output since there would be no new information to add. But feel free to include it if you want, just keep in mind that the syntax for the input/output descriptions is `@<input/output> <name>: <description>`. You don't have to follow the socket names though, you could also write something like `@input Input [0-n]: some description for all these inputs`.

----

I'm also wondering whether there is a better name for this node. In my opinion "inputs" should at least be plural (`Wait For Inputs`), but `Wait For All Inputs` or `Once All Active` or something like that might be clearer.

t3du:
Thanks for the comment. I'm gonna change the name to ` Wait For All Inputs` and update the documentation considering your words which sound more technically accurate. The reset is needed in the code in order for the node to work properly (as QC suggested with the array variable) but is not added as a node option. But now that you mention it,  it could also be added in the node as an additionally reset input.  Let me know.

PR #2960, opened 2023-11-01
Ss refraction

e2002e:
Screen space refraction.
I removed the min ray option as it wasn't used by the original shaders (since it's a binary search I set it to 0.5).
We can't refract refractive objects, also the ssr half res option isn't working with refraction.

I read the glsl docs, and it states that reflect / refract functions should take normalized values as parameters, so assuming it returns a normalized value.
So I'm sure it's safe to remove the normalize() that encloses these functions.


e2002e:
@MoritzBrueckner can you review this ?

Shaders/custom_mat_presets/custom_mat_deferred.frag.glsl
```diff
@@ -48,4 +48,8 @@ void main() {
 	#ifdef _EmissionShaded
 		fragColor[GBUF_IDX_EMISSION] = vec4(emissionCol, 0.0);
 	#endif
+	
+	#ifdef _SSRefraction
+		fragColor[GBUF_IDX_REFRACTION] = vec4(ior, opacity, 0.0, 0.0);
```
MoritzBrueckner, line 53: `ior` and `opacity` are probably undefined in this custom shader.

Shaders/ssr_pass/ssr_pass.frag.glsl
```diff
@@ -45,28 +44,27 @@ float getDeltaDepth(const vec3 hit) {
 
 vec4 binarySearch(vec3 dir) {
 	float ddepth;
-	vec3 start = hitCoord;
-	for (int i = 0; i < numBinarySearchSteps; i++) {
+	for (int i = 0; i < 7; i++) {
```
MoritzBrueckner, line 47: Should it be `i < maxSteps` instead of `i < 7`?

blender/arm/material/cycles_nodes/nodes_shader.py
```diff
@@ -170,10 +172,14 @@ def parse_holdout(node: bpy.types.ShaderNodeHoldout, out_socket: NodeSocket, sta
...
-    pass
-
+def parse_bsdfrefraction(node: bpy.types.ShaderNodeBsdfRefraction, out_socket: NodeSocket, state: ParserState) -> None: 
+    if state.parse_surface:
+        state.out_basecol = c.parse_vector_input(node.inputs[0])
+        c.write_normal(node.inputs[3])
+        state.out_roughness = c.parse_value_input(node.inputs[1])
+    if state.parse_opacity:
+        state.out_opacity = '0.0'
+        state.out_rior = c.parse_value_input(node.inputs[2])
```
MoritzBrueckner, line 182: There's this implementation for the Refraction BSDF node, but when using this, Armory prints `WARNING: Node tree "Refractive": material node type BSDF_REFRACTION not supported` to the console. The node is missing an entry in [node_meta.py](https://github.com/armory3d/armory/blob/0d2a0d5c31994e15c1ff52d4d10ea73fe571491d/blender/arm/material/node_meta.py#L65).

blender/arm/material/make_mesh.py
```diff
@@ -536,15 +542,25 @@ def make_forward(con_mesh):
                 frag.add_uniform('sampler2DShadow shadowMapSpot[4]', included=True)
 
     if not blend:
-        mrt = rpdat.rp_ssr  # mrt: multiple render targets
```
MoritzBrueckner, line 539: Why remove the comment that explains this abbreviation? I'm not sure how common it is, but I would imagine that the average reader doesn't know what "mrt" is supposed to say.

blender/arm/material/make_mesh.py
```diff
@@ -536,15 +542,25 @@ def make_forward(con_mesh):
...
-        if mrt:
+        mrt = 0
+        if rpdat.rp_ssr:
+            mrt = 2
+        if rpdat.rp_ss_refraction:
+            mrt += 1
+        if mrt != 0:
             # Store light gbuffer for post-processing
-            frag.add_out('vec4 fragColor[2]')
+            frag.add_out('vec4 fragColor[{mrt}]'.format(mrt = mrt))
```
MoritzBrueckner, line 552: You can simplify this by using `frag.add_out(f'vec4 fragColor[{mrt}]')` (note the `f` before the string literal)

MoritzBrueckner (commented):
Apart from a few very minor issues the code looks good to me :) Please note that I don't know much about screen space refraction, so I can't comment on whether the code makes much sense (performance-wise, for example).

However, I'm not really sure whether the code is working as expected: In a test scene with a Glass BSDF material, the object using this material simply becomes invisible at runtime, both in the deferred and in the forward RP.

Eeevee viewport:
Eevee screenshot

Armory output:
Armory screenshot

Test file:
[2960_pr_refraction.zip](https://github.com/armory3d/armory/files/13339760/2960_pr_refraction.zip)

Let me know if you need a RenderDoc capture of this file :)

Shaders/ssr_pass/ssr_pass.frag.glsl
e2002e: No, maxSteps are used for rayCast, binary search narrows down the output of this function and the original ssr shader sets a ssrMinRayStep to 7.

blender/arm/material/make_mesh.py
e2002e: Ok

blender/arm/material/cycles_nodes/nodes_shader.py
e2002e: Done !

Shaders/custom_mat_presets/custom_mat_deferred.frag.glsl
e2002e: I corrected that.

e2002e:
Hi @MoritzBrueckner, thanks for the review. I tested your file and it seems like you forgot to set the render path and also to turn on refraction in the post processing panel. Doing so shows me suzanne through the glass.

MoritzBrueckner:
Ah, oops, I didn't realize that :D Now it works on deferred, but on forward RP I get the following error:

```

**Comment by e2002e** [comment]
Ok, i'll fix this tomorrow.

**Comment by e2002e** [comment]
@MoritzBrueckner hi there !
I used an index to write to the correct sampler, just like in write_data.py so that it doesn't crash.
Nevertheless, it's not working in forward renderer yet. I suspect textures binding in forward render path are incorrect.

It does render refraction in forward rp only if ssr is activated.

**Comment by e2002e** [comment]
@MoritzBrueckner, fixed !
We have too bind the texture that is used with reflection even when only refraction's on.

**Comment by e2002e** [comment]
I was able to fix the raymarching by testing delta depth against depth, instead of with 0.0.
This enforces the refracted fragment is behind the refractive material, and allows for using binary search normally.
There is a downside to having only one refraction pass, it's that if refractive objects have two faces, and should normally invert the coordinates by adding of the effect, it can't "push" it as much as it should and so a refractive sphere with an high ior will not really show the world upside down. It works great with planes like lakes and windows though, for the rest I'd need to split the render into one pass per face's side.

**Comment by e2002e** [comment]
Hi there, concerning our talk on discord @MoritzBrueckner, yes please do a review.
Last time we fixed incoherences in the forward renderer but altogether it wasn't good: the raytracing was biased and there were normalization issues.
Now as you can see in the picture, the sphere inverts all rays while cylinder has this effect only horizontally, which is the expected behaviour.

File: blender/arm/material/make_mesh.py [hunk 6]
```diff
@@ -536,15 +542,28 @@ def make_forward(con_mesh):
...
             # Store light gbuffer for post-processing
-            frag.add_out('vec4 fragColor[2]')
+            frag.add_out(f'vec4 fragColor[{mrt}+1]')
             frag.add_include('std/gbuffer.glsl')
             frag.write('n /= (abs(n.x) + abs(n.y) + abs(n.z));')
             frag.write('n.xy = n.z >= 0.0 ? n.xy : octahedronWrap(n.xy);')
             frag.write('fragColor[0] = vec4(direct + indirect, packFloat2(occlusion, specular));')
-            frag.write('fragColor[1] = vec4(n.xy, roughness, metallic);')
+            if rpdat.rp_ssr:
+                frag.write('fragColor[1] = vec4(n.xy, roughness, metallic);')
```
MoritzBrueckner, line 559: Are you sure that this line should only be executed if `rpdat.rp_ssr` is true?

MoritzBrueckner (changes requested):

blender/arm/material/cycles_nodes/nodes_shader.py
```diff
@@ -35,11 +35,11 @@ def parse_mixshader(node: bpy.types.ShaderNodeMixShader, out_socket: NodeSocket,
     state.curshader.write('{0}float {1} = 1.0 - {2};'.format(prefix, fac_inv_var, fac_var))
 
     mat_state.emission_type = mat_state.EmissionType.NO_EMISSION
-    bc1, rough1, met1, occ1, spec1, opac1, emi1 = c.parse_shader_input(node.inputs[1])
+    bc1, rough1, met1, occ1, spec1, opac1, ior1, emi1 = c.parse_shader_input(node.inputs[0])
     ek1 = mat_state.emission_type
 
     mat_state.emission_type = mat_state.EmissionType.NO_EMISSION
-    bc2, rough2, met2, occ2, spec2, opac2, emi2 = c.parse_shader_input(node.inputs[2])
+    bc2, rough2, met2, occ2, spec2, opac2, ior2, emi2 = c.parse_shader_input(node.inputs[1])
```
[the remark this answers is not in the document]
e2002e: Ok, I rolled that back

blender/arm/material/make_mesh.py
e2002e: Yes true, refraction uses the roughness too.

e2002e:
@luboslenco can you merge this one ? It has been reviewed twice, is done and will not change anymore.
Only bug: ssr half res will not reflect refractive objects.

luboslenco:
Merged, thanks!

PR #2965, opened 2023-11-07
Show object properties in debug console

MoritzBrueckner:
Screenshot


rpaladin:
Finally something similar to the BGE debug properties profiler. 😍Nice work!

PR #2971, opened 2023-11-17
Armory Traits UI: Remove Traits and Print Scene Traits

t3du:
This is my first attempt to modify armory ui. I appreciate comments if you notice something is missing or it could be improved. I did my best with what I know.

I added an operator that removes all traits from active objects: 

image

Additionally I added to the operator **Print All Scenes Traits** the information of the Scene traits (@ rpaladin, I hope you find this ok). Before it only printed the object traits.

image

### DIFF
--- a/blender/arm/props_traits.py
+++ b/blender/arm/props_traits.py
@@ -739,6 +739,10 @@ class ARM_OT_RemoveTraitsFromActiveObjects(bpy.types.Operator):
     bl_idname = 'arm.remove_traits_from_active_objects'
     bl_description = 'Removes all traits from all selected objects'
 
+    @classmethod
+    def poll(cls, context):
+        return len(context.selected_objects) > 0
+
     def execute(self, context):
         for obj in bpy.context.selected_objects:
             obj.arm_traitlist.clear()
--- a/blender/arm/props_traits.py
+++ b/blender/arm/props_traits.py
@@ -734,6 +734,24 @@ class ARM_PT_SceneTraitPanel(bpy.types.Panel):
         obj = bpy.context.scene
         draw_traits_panel(self.layout, obj, is_object=False)
 
+class ARM_OT_RemoveTraitsFromActiveObjects(bpy.types.Operator):
+    bl_label = 'Remove Traits From Active Objects'
+    bl_idname = 'arm.remove_traits_from_active_objects'
+    bl_description = 'Removes all traits from all active objects'
+
+    def execute(self, context):
+        for obj in bpy.context.selected_objects:
+            lst = obj.arm_traitlist
+            while(len(lst) > 0):
+                lst.remove(0)
+            obj.arm_traitlist_index = 0
+
+        for obj in bpy.context.selected_objects:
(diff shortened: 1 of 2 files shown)


t3du:
I didn't try to do it for scenes, But instead for objects because usually there are a lot. The scene I think is more easy to handle manually.

blender/arm/props_traits.py
```diff
@@ -734,6 +734,21 @@ def draw(self, context):
...
+class ARM_OT_RemoveTraitsFromActiveObjects(bpy.types.Operator):
+    bl_label = 'Remove Traits From Active Objects'
+    bl_idname = 'arm.remove_traits_from_active_objects'
+    bl_description = 'Removes all traits from all active objects'
+
+    def execute(self, context):
+        for obj in bpy.context.selected_objects:
+            lst = obj.arm_traitlist
+            while(len(lst) > 0):
+                lst.remove(0)
```
MoritzBrueckner, line 746: You can just write `obj.arm_traitlist.clear()` if I'm not mistaken

MoritzBrueckner (commented):
A few remarks for discussion:

- What's the use case for these operators?
  - When do you need to remove all traits from selected objects?
  - When do you need to list all traits, but without knowing what objects they're on (this isn't printed at the moment)? Can we indent all traits within a scene and within objects and also print object names? That would at least be somewhat useful then. Ideally we could just display traits in the outliner, but that's probably not possible...
- Maybe `Remove Traits From Active Objects` should be renamed to `Remove Traits From Selected Objects`? The description of that operator also needs to be changed then.
- Maybe `Print All Scenes Traits` should be renamed to `Print All Traits` (since that's what it is)?

blender/arm/props_traits.py
t3du: better i will try that thanks!

t3du:
The listing already exists. I just added the scene traits. I consider both list and remove options useful. Sometimes one needs to revisit a file in order to continue developing in terms of logic assignment and that mapping helps to get a perspective. Other times the logics change and traits need to be remove from several objects in order to assign new ones or none.

Sure indent will make the output look better.

Sure I'll change all the names for simplicity.

Thanks for all your comments MoritzBrueckner.

blender/arm/props_traits.py
t3du: it did the trick!

t3du:
done!

rpaladin:
I had a few grammar issues with my initial post. What I meant to say is that because your new feature doesn't work for Scenes, it should be greyed-out in the Scene tab of the Properties Editor.  Notice how the `Copy Traits from Active Object` button is greyed-out in your screenshot. Your new feature should be greyed-out too if you're looking at the Scene tab. If I remember correctly from when I implemented the copy traits feature, I was able to grey-out the button by using a poll statement: https://github.com/armory3d/armory/blob/main/blender/arm/props_traits.py#L744-L746C87

t3du:
Wow thanks it works. thats what poll does:
image

Regarding the is_object I used the function that already existed that you created and added the operator.

PR #2972, opened 2023-11-22
Update WebSocket and replace Regex

Onek8:
Recent changes to PCRE need to be investigated to find out the cause of matched index outside bounds caused by the regex matches.

### DIFF
--- a/Sources/armory/network/WebSocket.hx
+++ b/Sources/armory/network/WebSocket.hx
@@ -190,7 +190,9 @@ class WebSocket extends WebSocketCommon {
 	}
 
 	inline private function parseUrl(url)
-	{
+	{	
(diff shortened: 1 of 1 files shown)


Sources/armory/network/WebSocket.hx
```diff
@@ -190,7 +190,9 @@ class WebSocket extends WebSocketCommon {
 	}
 
 	inline private function parseUrl(url)
-	{
+	{	
+		/** TO DO - FIND OUT WHAT IS BREAKING REGEX IN THE NEW PCRE2 FOR HL C
```
MoritzBrueckner, line 194: See https://github.com/Kode/Kha/pull/1478.

It's still a good idea to merge this I believe, depending on how long it takes for armory3d/Kha to be updated.

PR #2981, opened 2023-12-15
Play Sound Node: add option for sound/sound name

t3du:
By default the node will continue to be used as usual, but the option added will allow to specify a sound by a string name. This allows to parametrice the sounds with logic instead of having several sound nodes defined, very useful when having a lot of sounds. Additionally, sounds doesn't need to be imported to blender in order to be used by the sound node, as similar as the draw image node or the draw image sequence node, they just need to be in the Bundled folder.

image


t3du:
hi, i tested the node with existing play sound nodes and even the warning does appears it works fine. I didn't change the inputs so i think re connections are not needed. It seems that is what the update does? sorry i don't the update routine well enough, but it seems i forgot to change the line:

`self.arm_version not in (0, 2): `

PR #2984, opened 2023-12-20
play sound node: fix update routine version

t3du:
fix version


MoritzBrueckner:
Hi, this is still wrong because `(0, 2)` is a tuple, not a range. This means that now (with this PR merged), updating sound nodes with version 1 will cause a `LookupError` because 1 is not in `(0, 2)`. I fixed it in https://github.com/armory3d/armory/commit/a67b5f5222007bf527a02544da3b78c3d249bfa6, this way you don't need to open another PR :)

t3du:
thanks again! you're the best!

PR #2990, opened 2024-01-19
Voxelgi with clipmaps.

e2002e:
This pull request implements VoxelGI with clipmaps (working with AO too).
The image used to store the voxels is scaled on the y axis by the number of clipmaps, and each level of detail is stored at a different offset, so that we just have to compute the offset in the conetracer according to the distance to the camera, so the clipmap_center is a new parameter for tracing functions.

I removed the snapping mechanism because it was making seams and it already occurs in the conetracer.
We don't compute voxpos anymore, only the clipmap_center (eye + eyeLook) is passed to the conetracing function.

The dimension option, revoxelize, and 'around camera' where removed, since clipmapping depends on the distance to the camera. The new way of computing dimension is 'voxelSize * voxelgiResolution'. You'll see the new option size in the ray panel ; it determines the size of each voxels thus the precision of the voxelization.

Second bounce is working but it's not compatible with the temporal filter, I don't know why it's doing some splashes and smears. This is the only bug.

There is the forum's thread with some renders:
https://forums.armory3d.org/t/voxel-gi-with-clipmaps/5242/5


e2002e:
I'm closing (will rewrite for compute shaders)

PR #2996, opened 2024-02-06
Ssgi

e2002e:
This pull request introduces changes to the rtao shader and reimplements rtgi from version 19.02.
Changes are:
- no tweak to ssgiRayStep value
- base distance factor of 1.0
- maxSteps option is remove in favor of step & search dist which are used to compute the maxStep like in ssr.
- no changes are made to fragColor during rayCasting, we store the final value in a variable and will afterward assign to fragColor and divide / multiply according to the number of cones.
- rtgi uses another blur shader because the rtao / ssao one just work on one channel (I retrieved the old code's shader).
- rtgi uses it's own render target (RGBA32).

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -295,11 +295,11 @@ void main() {
 	fragColor.rgb = envl;
 
 #ifdef _SSAO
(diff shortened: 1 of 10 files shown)


e2002e:
I need to rework the rtgi.

PR #3001, opened 2024-02-14
Fix forward clusters

e2002e:
There was some multiplication vector that you forgot to add to forward make_clusters.py
This breaks the clusters in forward, this pull request fixes it.
I also made the spotclusters array a mat4 since that's what it it. Just adding a clause in shaders.py allows it.


luboslenco:
Thanks!

PR #3006, opened 2024-03-16
fix normalize issue.

e2002e:
The use of normalize() on the getPosView() function is making strange refractions on flat surfaces.
Instead, just use the negative return of getPosView()
Passing 1.0 / ior to refract() is also not good.

After changing this the render is correct.
Capture du 2024-03-25 14-05-37

I've set the opacity to 0.0 for glass, if user wan't a mix they can use the alpha value of principled.

@luboslenco this is fixing a broken feature that I introduced, and is compatible with the voxels pull request (voxelization of refractive objects). You should merge this one before

### DIFF
--- a/Shaders/ssrefr_pass/ssrefr_pass.frag.glsl
+++ b/Shaders/ssrefr_pass/ssrefr_pass.frag.glsl
@@ -81,11 +81,10 @@ void main() {
     vec3 n;
     n.z = 1.0 - abs(enc.x) - abs(enc.y);
     n.xy = n.z >= 0.0 ? enc.xy : octahedronWrap(enc.xy);
-	n = normalize(n);
 
     vec3 viewNormal = V3 * n;
     vec3 viewPos = getPosView(viewRay, d, cameraProj);
-    vec3 refracted = refract(-viewPos, viewNormal, ior);
+    vec3 refracted = refract(normalize(viewPos), viewNormal, 1.0 / ior);
     hitCoord = viewPos;
 
     vec3 dir = refracted * (1.0 - rand(texCoord) * ss_refractionJitter * roughness) * 2.0;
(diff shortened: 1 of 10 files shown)


e2002e:
@MoritzBrueckner I need your help again to review this fix.

Shaders/ssrefr_pass/ssrefr_pass.frag.glsl
```diff
@@ -102,27 +100,27 @@ void main() {
...
 
-#ifdef _CPostprocess
-    vec3 dir = refracted * (1.0 - rand(texCoord) * PPComp10.y * roughness) * 2.0;
-#else
+//#ifdef _CPostprocess TODO
+///    vec3 dir = refracted * (1.0 - rand(texCoord) * PPComp10.y * roughness) * 2.0;
+//#else
     vec3 dir = refracted * (1.0 - rand(texCoord) * ss_refractionJitter * roughness) * 2.0;
-#endif
+//#endif
```
MoritzBrueckner, line 111: What's the reason for all the `_CPostprocess` related TODOs here and in some other places in this file?

e2002e: The file is a modifies copy from ssr_pass.
Since I haven't implemented the C postprocess I commented out the related code.

e2002e:
@luboslenco it is important that you merge this as I introduced a broken functionality with the initial pull request for refraction.

e2002e:
I'm going to make a forward pass for refraction.

e2002e:
@MoritzBrueckner. Now refraction is only a forward pass, even when in deferred, still shading the refractive objects.

e2002e:
Still having strange results, will work on that this week.

e2002e:
@MoritzBrueckner @luboslenco 
I changed the way refract is done a little. Now it's like translucent pass, meaning that it's a forward pass, and instead of translucent resolve, the refraction shader is dispatched.
It did this because I noticed that shadowmaps are only generated for objects that have a "mesh" pass. This is still discarded but when we have transparent shadows I think this will be usefull.

PR #3008, opened 2024-03-21
Gi clipmaps

e2002e:
Hello, this pull request implements voxel gi with clipmaps.
I've been working on this from last year as you can see in the forum's thread:
https://forums.armory3d.org/t/voxel-gi-with-clipmaps/5242/116

Basically clipmaps are hypercubes, instead of having a fixed voxel size grid, we have several grids for which voxels double in size each time.
I copied code from WickedEngine and it was a bit complicated to understand everything, but the author answered my questions and everything is correctly setup, copyrights are included in the files where they should.

The extents options is gone as it depends on the voxel's size, the resolution, and the number of clipmaps. Adding an extra parameter is just a mess to calibrate and will not help the user.

The GI doesn't implement refractions.
The conetracer comes with a file containing precomputed cone directions that we use to write anisotropic data into the image on the x axis. The y axis is used for stacking clipmaps. That gives the images resolution a heavy impact on performances, and with my modest rtx3050, I must set it to 32 for GI, and 64 maximum for AO.

There are new compute shaders for stabilizing the grids and also generating datas (levels / precomputed datas for the conetracer / lighting).
These will not work with current's Krom's version and you should recompile it with the Kinc from my repo:
https://github.com/armory3d/Kinc/pull/4

Instead of tracing cones directly in the deferred_light shader, we use a compute shader to do so, and write the result to a sampler, that allows skipping tracing the same voxels and is a great optimization (from WickedEngine again).
Unfortunately, we can't do this with forward renderer.

One last thing: the voxels only work with opengl. Because (correct me if i'm wrong) kha automatically converts glsl shaders to hlsl for direct x11, but there is no shaders to convert for the voxelizer. They are generated by make_voxels.py so we need to write the hsls variant here. There is already an attempt to do so and it should be any for someone with a bit of experience and windows to finish that.
If really necessary i have an old desktop pc where i could install windows to do this.


e2002e:
I must fix a stroboscopic effect due to alternations of images.

PR #3009, opened 2024-03-25
voxels with levels of details.

e2002e:
Hello !
The AO module also benefits from the clipmaps, but is unstable and has a "dancing" effect.

- The voxelizer is pretty much the same thing appart that it writes to an image that is 6 times bigger on the x axis to store anisotropy.

 - A ping-ponging mechanism is used for stability between a shader computing the offset from the previous frame and the temporal shader that will do some extra boundary checks and radiance modification, using data from a simple lighting shader, the environment light, and an extra diffuse bounce. 
This temporal compute shader is also responsible for generating the different levels of details and the precomputed datas.

 - The conetracer has been replaced and doesn't compute cone direction from the normal, instead it has a list of precomputed directions.
 
 - The conetracing functions are called inside of compute shaders to avoid unecessary work and the final result from the voxelization is writen on 2D images here before being sampled. This procedure is not possible in forward render and we trace the cones directly.

- Specular conetracing uses a dither matrix to tamper with the fact that it only traces one cone.

 - The extents option has been removed and a new "clipmap count" and "size" option were added. The total extents is now (resolution * voxelSize) ** clipmap_count. The offset option was also removed since there is only once correct value for it.
 
- The voxels refraction is not (re)implemented.

- There is SDF (signed distance field) for speculars.

NOTE:
- Most of the new shaders come from WickedEngine and are copyrighted, my job was to implement them and keep the same procedure. It offers pretty much the same features (that i've listed above).

BUG:
- Environment light doesn't work with forward renderer.


Sources/armory/renderpath/RenderPathDeferred.hx
```diff
@@ -491,7 +518,7 @@ class RenderPathDeferred {
 		}
 		#end
 
-		#if (rp_ssr_half || rp_ssgi_half)
+		#if (rp_ssr_half || rp_ssgi_half || rp_voxels != "Off")
```
e2002e, line 521: I am using half depth for the call to getPos() in the resolve_diffuse/spec/ao shader simply because I don't know how to get the _main depth buffer from

blender/arm/make_renderpath.py
```diff
@@ -133,6 +153,7 @@ def add_world_defs():
 
     if '_Rad' in wrd.world_defs and '_Brdf' not in wrd.world_defs:
         wrd.world_defs += '_Brdf'
+        assets.add_khafile_def("arm_brdf")
```
e2002e, line 156: we need this to for environment lighting

blender/arm/make_world.py
```diff
@@ -82,6 +82,7 @@ def build():
                         if write_radiance:
                             # Set world def, everything else is handled by write_probes()
                             wrd.world_defs += '_Rad'
+                            assets.add_khafile_def("arm_radiance")
```
e2002e, line 85: same

blender/arm/make_world.py
```diff
@@ -184,6 +185,7 @@ def build_node_tree(world: bpy.types.World, frag: Shader, vert: Shader, con: Sha
         solid_mat = rpdat.arm_material_model == 'Solid'
         if rpdat.arm_irradiance and not solid_mat:
             world.world_defs += '_Irr'
+            assets.add_khafile_def("arm_irradiance")
```
e2002e, line 188: lastly

blender/arm/material/cycles_nodes/nodes_texture.py
```diff
@@ -507,6 +507,7 @@ def parse_tex_environment(node: bpy.types.ShaderNodeTexEnvironment, out_socket:
         # Append LDR define
         if disable_hdr:
             world.world_defs += '_EnvLDR'
+            assets.add_khafile_def("arm_envldr")
```
e2002e, line 510: lastly

blender/arm/material/make_mesh.py
```diff
@@ -551,18 +552,22 @@ def make_forward(con_mesh):
...
+            if rpdat.rp_ssr:
+                frag.write(f'fragColor[{index}] = vec4(n.xy, roughness, metallic);')
+                index += 1
             if rpdat.rp_ss_refraction:
                 if parse_opacity:
-                    frag.write(f'fragColor[2] = vec4(ior, opacity, 0.0, 0.0);')
+                    frag.write(f'fragColor[{index}] = vec4(ior, opacity, 0.0, 0.0);')
                 else:
-                    frag.write(f'fragColor[2] = vec4(1.0, 1.0, 0.0, 0.0);')
+                    frag.write(f'fragColor[{index}] = vec4(1.0, 1.0, 0.0, 0.0);')
```
e2002e, line 570: this fixes a bug with only refraction on

blender/arm/material/make_mesh.py
```diff
@@ -637,49 +642,63 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
...
+        frag.write('envl += prefilteredColor * (f0 * envBRDF.x + envBRDF.y);')
+    elif '_EnvCol' in wrd.world_defs:
+        frag.add_uniform('vec3 backgroundCol', link='_backgroundCol')
+        frag.write('envl += f0 * envBRDF.x + envBRDF.y);')
 
     frag.add_uniform('float envmapStrength', link='_envmapStrength')
-    frag.write('indirect *= envmapStrength;')
+    frag.write('envl *= envmapStrength * occlusion;')
 
     if '_VoxelAOvar' in wrd.world_defs:
```
e2002e, line 674: we need to dissociate environment light from indirect for GI

blender/arm/material/make_mesh.py
```diff
@@ -637,49 +642,63 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
...
-                frag.write('prefilteredColor = pow(prefilteredColor, vec3(2.2));')
-            frag.write('indirect += prefilteredColor * (f0 * envBRDF.x + envBRDF.y) * 1.5;')
-        elif '_EnvCol' in wrd.world_defs:
-            frag.add_uniform('vec3 backgroundCol', link='_backgroundCol')
-            frag.write('indirect += backgroundCol * f0;')
+            frag.write('envl /= PI;')
     else:
-        frag.write('vec3 indirect = albedo;')
-    frag.write('indirect *= occlusion;')
+        frag.write('vec3 envl = vec3(0.0);')
```
e2002e, line 649: we need to dissociate envl from indirect for GI

e2002e:
I have an Issue with windowed mode. Will investigate and reopen once fixed.

PR #3010, opened 2024-03-31
GI LOD

e2002e:
Hello !
The AO module also benefits from the clipmaps.

- The voxelizer is pretty much the same thing appart that it writes to an image that is 6 times bigger on the x axis to store anisotropy.

 - A ping-ponging mechanism is used for stability between a shader computing the offset from the previous frame and the temporal shader that will do some extra boundary checks and radiance modification, using data from a simple lighting shader, the environment light, and an extra diffuse bounce. 
This temporal compute shader is also responsible for generating the different levels of details and the precomputed datas.

 - The conetracer has been replaced and doesn't compute cone direction from the normal, instead it has a list of precomputed directions.
 
 - The conetracing functions are called inside of compute shaders to avoid unecessary work and the final result from the voxelization is writen on 2D images here before being sampled. This procedure is not possible in forward render and we trace the cones directly.

- Specular conetracing uses a dither matrix to tamper with the fact that it only traces one cone.

 - The extents option has been removed and a new "clipmap count" and "size" option were added. The total extents is now (resolution * voxelSize) ** clipmap_count. The offset option was also removed since there is only once correct value for it.
 
- The voxels refraction is not (re)implemented.

- There is SDF (signed distance field) for speculars and voxels shadows ; it's an optimization.

NOTE:
- Most of the new shaders come from WickedEngine and are copyrighted, my job was to implement them and keep the same procedure. It offers pretty much the same features (that i've listed above).

BUG:
- Environment light doesn't work with forward renderer.


e2002e:
There are bugs with windowed mode. I will reopen once fixed.

e2002e:
I'm reopening since the windowing issue seems fixed. The is one scene I got from blenderkit where I can't use windowed mode. It would produce strange displacements. The only issue I can think of is that there is a depth buffer issue when resizing the window in RenderPath.hx, but I already had it ignore images (they still should be resized). I'll leave a note in iron's pull request.
Every other scenes I tested is working whatever the settings are.

e2002e:
@luboslenco I thoroughly tested this today and GI works fine but for the  bug described.
AO is a bit unstabled and I'm not sure why, maybe because the "voxels" image is being writen without using atomic operations.
I'll try with R32 and imageAtomicMax function to see if that helps.

e2002e:
@luboslenco AO is stabilized !
It was indeed because the image used for voxelizing shaders needed to be compatible with image atomic operations, so I've ma it an r32.

PR #3012, opened 2024-04-11
Gi clipmaps

e2002e:
Hello !
The AO module also benefits from the clipmaps.

- The voxelizer is pretty much the same thing appart that it writes to an image that is 6 times bigger on the x axis to store anisotropy (data modified by normal weight), and 9 times bigger on the z axis to store the different channels (4 for rgba, 3 for emission rgb, and 2 for normals).

- A ping-ponging mechanism is used for stability between a shader computing the offset from the previous frame and the temporal shader that will do some extra boundary checks and radiance modification, using data from a simple lighting shader, the environment light, and a second diffuse bounce.
- This temporal compute shader is also responsible for generating the different levels of details stored on the images y axis and the precomputed datas (on x axis).

- The conetracer has been replaced and doesn't compute cone direction from the normal, instead it has a list of 16 precomputed directions.

- The conetracing functions are called inside of compute shaders to avoid unecessary work and the final result from the voxelization is writen on 2D images here before being sampled in deferred_light. This procedure is not possible in forward render and we trace the cones directly in the object's shader.

- The extents option has been removed and a new "clipmap count" and "size" option were added. The total extents is now (resolution * voxelSize) ^ clipmap_count. The offset option was also removed.

- The voxels refraction is not (re)implemented. Let me know if this should be a TODO.

NOTE:
All compute shaders (excepted voxels_light) come from WickedEngine, the licence is included on top of these files, even though they are converted from hlsl to glsl, and use armory's functions.
I have worked a long time with legacy conetracer before implementing WickedEngine's. From here, to have everything work together, it was better to implement (let's not say copy) the whole procedure.


e2002e:
Everything's fixed, I'm reopening now and will resolve conflicts tomorrow.

e2002e:
I Broke environment light when fixing the last bugs, I'm certain this is just some uniform or output value missing because it was working.
I'll reopen when everything works correctly (will give me time to check voxels shadows because they don't seem to work either).

e2002e:
Sorry about all these open/closed pull requests. I worked a bit disorganized and often forced pushed the branch while the pull request was closed. This is why I got to open a new one after.

e2002e:
Not working

e2002e:
@lubos this is working. I've made another change to iron.

e2002e:
This produces correct renders, sampling is consistent and seams are not abrupt, in the worst case we can see the blending between lod:
Capture du 2024-04-25 16-42-34
Environment light is supposed to be Nishita, hosek / wilkie has to be pushed up a lot and turns red under some angles.
This is a cycles / armory comparison with the smallest voxel's image size possible and aliasing dp set to 540:
office_cycles
Capture du 2024-04-25 16-44-50

e2002e:
There is an issue with normal being declared after it's used with some nodes setup. I'll look into it next week.

e2002e:
Ok the issues are fixed !
Last thing that I could see not working is normal map nodes with a color input. They must be bump nodes, otherwise the material iis all black. But i'm not sure whether this is an issue with the voxelizer.

e2002e:
@MoritzBrueckner can you review ? I'll ping lubos then.

e2002e:
There is an issue with emissive materials, they are inconsistently voxelized.
I'm working on that this week, afterward it should be done.

e2002e:
Hi @luboslenco, this is finished !
I spent the week testing it and there aren't any other bugs than those listed.
It's actually seamless now.
I'll still be trying to get voxels shadows to work. But I think you can merge.

e2002e:
@luboslenco don't forget to ship the updated Krom with this. It needs the commits you merged a few months ago (rgba8 3d images and compute shaders support).
I'll be looking for a way to implement transparent shadows now.

PR #3015, opened 2024-05-09
Move assets for published html5 games to /data folder

moisesjpelaez:
Move assets for published html5 games to /data folder instead of keeping them in the project's root folder.


luboslenco:
Hm, I think this was done as some sites prevented data folders when publishing html5 games, is that no longer the case?

moisesjpelaez:
I'm not sure but I can confirm have tested it on itch.io, with localhost and in my personal website (with paid web hosting/domain) and it has been working fine so far.

PR #3019, opened 2024-07-04
SSS fixes

e2002e:
This pull request implements the sss.glsl transmittance function into light.glsl so that not only config with
_Sun
_Spot && _SinglePoint
get a call to SSSTransmitance() function.

It also allows the refractive materials to have a materialID used for testing which material is processed by the SSS shader itself.

So refractive and translucent objects can have subsurface materials.

All the options from the principled shader regarding the subsurface are used in the pass excepted the anisotropy.  

The updated SSSBlur() function is enhanced by deepseek coder. Of course it's reviewed and tested.
The sss.glsl shader gets a cubeSSSSTransmittance function from deepseek-coder too, and it seems to work as the subsurface effect is much more powerful when the light is behind the object from the camera pov. (The video shared bellow uses a point light).

The rock is blue but the subsurface color is green, the refraction is enabled and the opacity is < 1.0.
Screenshot_20240704_074003

### DIFF
--- a/blender/arm/material/make.py
+++ b/blender/arm/material/make.py
@@ -137,6 +137,21 @@ def parse(material: Material, mat_data, mat_users: Dict[Material, List[Object]],
 
         elif rp == 'translucent' or rp == 'refraction':
             c['bind_constants'].append({'name': 'receiveShadow', 'boolValue': material.arm_receive_shadow})
+
+            if material.arm_material_id != 0:
+                c['bind_constants'].append({'name': 'materialID', 'intValue': material.arm_material_id})
+
+                if material.arm_material_id == 2:
+                    wrd.world_defs += '_Hair'
+
+            elif rpdat.rp_sss_state != 'Off':
+                const = {'name': 'materialID'}
+                if needs_sss:
+                    const['intValue'] = 2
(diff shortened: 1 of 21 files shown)


blender/arm/material/make.py
e2002e, line 1: The matid needs to be set into gbuffer0.a channel of two floats before sss is drawn

blender/arm/material/make_mesh.py
```diff
@@ -692,6 +708,15 @@ def make_forward_base(con_mesh, parse_opacity=False, transluc_pass=False):
         frag.write('}')
     frag.write('vec3 direct = vec3(0.0);')
 
+    is_shadeless = mat_state.emission_type == mat_state.EmissionType.SHADELESS
```
e2002e, line 711: materialID is read here and we set matid.
The reason is that make_forward_base is called before any other setups in make_refract and the matid is used for sampling light in make_forward_base (for transluc too)

e2002e:
You can see the module in action here:
https://www.youtube.com/watch?v=pfQho2K0rXI

I got a last TODO, removing the width option under renderpath since now the width is material dependent.

PR #3022, opened 2024-07-13
Gi clipmaps optimizations and crash fix.

e2002e:
- voxels resolve image are set to RGBA32.
- voxels light image uses offset coordinates to store rgb values in disctinct channels.
- imageAtomicAdd is used in place of imageAtomicMax for writing to voxelsLight.
- step size is used for diffuse conetracing to avoid missing voxels when resolution is high.
- moved the voxel size option to the voxels settings (not the ray panel).

### DIFF
--- a/Sources/armory/renderpath/Inc.hx
+++ b/Sources/armory/renderpath/Inc.hx
@@ -566,19 +566,12 @@ class Inc {
 		var res = iron.RenderPath.getVoxelRes();
 		var resZ =  iron.RenderPath.getVoxelResZ();
 
-		if (t.name == "voxels_diffuse" || t.name == "voxels_specular") {
+		if (t.name == "voxels_diffuse" || t.name == "voxels_specular" || t.name == "voxels_ao") {
 			t.width = 0;
 			t.height = 0;
 			t.displayp = getDisplayp();
 			t.scale = getSuperSampling();
-			t.format = getHdrFormat();
-		}
-		else if (t.name == "voxels_ao") {
-			t.width = 0;
-			t.height = 0;
-			t.displayp = getDisplayp();
-			t.scale = getSuperSampling();
-			t.format = "R8";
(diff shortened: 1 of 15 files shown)


Sources/armory/renderpath/Inc.hx
```diff
@@ -571,7 +571,7 @@ class Inc {
 			t.height = 0;
 			t.displayp = getDisplayp();
 			t.scale = getSuperSampling();
-			t.format = getHdrFormat();
+			t.format = "RGBA32";
```
e2002e, line 574: This can' t be set to RGBA64 and is not meant to

e2002e:
Now it's possible to get "clean" conetracing by letting the traceDiffuse function pass the step_size for conetracing, the original code I took would only pass the step_size for reflective cones, but then when increasing the resolution of the grid we have a lot of data that the conetracer misses. So that is fixed and everything should work correctly with correct voxels settings.
cornell

Sources/armory/renderpath/Inc.hx
luboslenco: Why is `voxels_ao` set to `RGBA32` instead of `R8` though?

e2002e: @luboslenco I was reworking the whole stuff and did this for debugging purpose, to see a proper black and white image of the AO, I am going to roll that back.

e2002e:
Here is a demo video:
https://www.youtube.com/watch?v=7xIixE-YJT4&t=316s&ab_channel=OweinDouard
Just be aware that all this asynchronous compute shaders process can't really happen in realtime on my laptop RTX 3050 with the settings I used (but it's possible by lowering), hence the instability.
I also have a last contribution to make for armory3d before I become an end-user again. I'll open the pull request when we'r done here, you'll be really glad.

PR #3024, opened 2024-07-22
fix voxel gi

e2002e:
This is a patch for the voxels module:
- Fix occlusion being multiplied by voxels specular instead of material's specular.
- Implements voxels shadows and also voxels refraction. Deals with material only illumination.
- Fix screen space refraction depth test that made background only appear through transparent objects.
- Removed the normalization of viewPos in the use of reflect / refract.
- Compute the environment light in the voxelizer and pass it in the "voxels" image for accuracy.

TODO:
- Voxels refraction only work with screen space refraction on, yet no hint is given in the UI.

Bugs:
- Forward renderer has wrong coordinates for sampling the voxels resolving images, so it results in strange projections.
The materials created with make_forward_base.py are working fine in deferred.

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -8,8 +8,13 @@
 #ifdef _Irr
 #include "std/shirr.glsl"
 #endif
-#ifdef _VoxelShadow
+#ifdef _VoxelRefract
 #include "std/conetrace.glsl"
+#endif
+#ifdef _VoxelShadow
+#ifndef _VoxelRefract
+#include "std/conetrace.glsl"
+#endif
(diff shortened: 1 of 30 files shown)


e2002e:
@luboslenco don't merge that yet, I'm in the middle of refactoring.
The voxels shadows work and everthing is setup but the forward materials probably don't have correct bindings for resolving textures so no forward renderpath and no voxels refraction with the deferred one.

Just so that you know I was able to make the forward renderer use the environment light for voxels and use the resolves shaders, this means pretty renders with sky and faster render time overall.

e2002e:
@lubos, you can merge than if you don't mind voxels not working in forward renderer.
Everything is functional in deferred,

PR #3027, opened 2024-08-06
bump overhaul !

e2002e:
This is a pull request with fixes for voxels and screen space refraction and a simple implementation for transparent shadows. Also fixes some bugs and add the possibility to render only lighting from an object for top down games elements like roofs.


e2002e:
Voxels shadows:
Capture d’écran du 2024-07-29 23-14-06

Material illumination:
Capture d’écran du 2024-08-02 10-30-07

Transparent shadows:
Capture d’écran du 2024-08-02 10-28-26

e2002e:
@lubos, I have been testing this during the week. So there's a few commits more and it shouldn't change.
I am currently trying to implement colored  voxels shadows, but you should merge this before because I don't think I will succeed.

e2002e:
@luboslenco I'm not touching this anymore. You can review if you want;

PR #3028, opened 2024-08-06
Upgrading and testing on Blender 4.3

Onek8:
Managed to keep support for 3.6 to 4.3 using 4.1 as the reference where I believe the node socket interface was removed. Also added a responsive canvas which should actually replace the previous commit adding so many lines for resizing which does not seem to even work on debian linux firefox as it was.


Onek8:
*** Bone layers missing for now

PR #3030, opened 2024-08-12
Blender version support

Onek8:
Clean up some node indexes and various features not available in different blender versions for better support. Also commented out QuantumCoderQC's bone animation support in the exporter for when he merges


Onek8:
I will remove the bone layers just in case it complicated Quantums merge into the animation branch. ¯\_(ツ)_/¯

PR #3031, opened 2024-08-17
Patch voxels

e2002e:
I'm reopening this pull request under another name because I got a little organisation issue.
It patches the voxels, implements transparent shadows, and adds a visible shadow option in case the user wants to have invisible ceilling that still casts shadows.

This is the forum's link where there are a few screenshots: https://forums.armory3d.org/t/patch-for-voxels-and-sreenspace-refraction/5763


blender/arm/exporter.py
```diff
@@ -770,7 +770,7 @@ def export_object(self, bobject: bpy.types.Object, out_parent: Dict = None) -> N
             else:
                 if not bobject.visible_camera:
                     out_object['visible_mesh'] = False
-                if not bobject.visible_shadow:
+                if not bobject.arm_visible_shadow:
```
e2002e, line 773: It's better that armory has it's own visible shadow option

PR #3016, opened 2024-05-15
Fix html5 canvas auto resize

moisesjpelaez:
Following Kha's [wiki](https://github.com/Kode/Kha/wiki/HTML5#full-page-canvas), adds a define in `khafile.js` and customizes `Main.hx` if target is 'html5' and set to full-screen.

Fixes #2515

PR #3034, opened 2024-08-27
Add optional arguments for Event listeners.

moisesjpelaez:
This is a proposal, opening a PR since it's relatively easy to implement.

The `send()` method in the `Event` class currently only calls functions with no arguments, with this PR is possible to add custom arguments when calling `Event.send()`.

I have tested it and worked, it shouldn't break compatibility with current Nodes and previous Projects. The current downside is that the console doesn't throw an error when a listener doesn't match the sender's arguments.

### DIFF
--- a/Sources/armory/network/WebSocket.hx
+++ b/Sources/armory/network/WebSocket.hx
@@ -109,7 +109,7 @@ class WebSocket {
 		_onmessage = value;
 		_ws.onmessage = function(message: Dynamic) {
 			if (_onmessage != null) {
-				if (Std.isOfType(message.data, JsBuffer)) {
+				if (message.data is JsBuffer) {
 					var buffer = new Buffer();
 					buffer.writeBytes(Bytes.ofData(message.data));
 					_onmessage(BytesMessage(buffer));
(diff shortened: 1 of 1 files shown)


Onek8:
I dont think this is correct/useful in any circumstance it is just adding overhead imports like your changes for the responsive canvas which you never reverted. You also did not update the nodes to go along with these changes in any case..

Onek8:
Reflect is slow as well to be noted for events

moisesjpelaez:
There are some cases I mentioned in the server but mentioning them here as well: collecting coins, ammo, health, transitioning between levels etc. this by passing different ids, names, amounts and so on as arguments. With the current implementation, as far as I'm aware, the user has to either create new events (which I'm sure is heavier) or think on architecture/data structures for these cases I mentioned.

The Nodes don't need to be changed at all, they keep working as they are intended to, this simply adds the possibility to add optional arguments to listeners when an event is sent using Haxe. The performance has been the same so far in my project and the change on the code has been minimal.

I have tested this in a Samsung Galaxy S6 lite tablet, which isn't that great for 3D games and have got ~30fps with and without this change.

The canvas resizing is another topic, I think we can move that conversation [there](https://github.com/armory3d/armory/pull/3016).

Onek8:
Architecture design as in I have all those features in game without the use of events and multiple ways to use events for those features without needing to pass arguments via the event.  ¯\_(ツ)_/¯

PR #3037, opened 2024-08-28
Add optional arguments for Event listeners.

moisesjpelaez:
This is a proposal. Moved the conversation from this [PR](https://github.com/armory3d/armory/pull/3034) and reopening it here since I messed up that previous branch.

Description:
The `send()` method in the `Event` class currently only calls functions with no arguments, with this PR is possible to add custom arguments when calling `Event.send()`.

I have tested it and worked, it shouldn't break compatibility with current Nodes and previous Projects. The current downside is that the console doesn't throw an error when a listener doesn't match the sender's arguments.

### DIFF
--- a/Sources/armory/system/Event.hx
+++ b/Sources/armory/system/Event.hx
@@ -1,7 +1,6 @@
 package armory.system;
 
 import haxe.Constraints.Function;
-import haxe.Rest;
 
 /**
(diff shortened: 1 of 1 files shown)


moisesjpelaez:
I'm closing this since I implemented a Signal class for what I needed.

PR #3039, opened 2024-08-30
Gi clipmaps

e2002e:
Hello @luboslenco,
This pull request fixes refraction and voxels.
There were a few inconsistencies in the voxelizing pipeline, and the screen space refraction wasn't still correct, so now it is, and also we can see the sky through refractive objects. We tested it and it was possible to maje a rifle scope.

Everything is tested and working, but voxels shadows can't be implemented yet for transparent objects in light.glsl because we have no way of knowing if the object is transparent and trace the cones directly instead of using resolve textures.

This will be implemented in my next pull request for transparent shadows, but now I'm leaving it to you to merge, I will work on another branch based on this one.

### DIFF
--- a/Shaders/deferred_light/deferred_light.frag.glsl
+++ b/Shaders/deferred_light/deferred_light.frag.glsl
@@ -8,9 +8,6 @@
 #ifdef _Irr
 #include "std/shirr.glsl"
 #endif
-#ifdef _VoxelShadow
-#include "std/conetrace.glsl"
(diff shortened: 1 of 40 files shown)


e2002e:
@luboslenco I don't know when you'r planning the next release but please merge this before.
I have taken a lot of time for refraction and voxels and as it's in main right now it's not usable, while the patch fixes everything.

PR #3051, opened 2024-09-22
Gi clipmaps patch 1.1

e2002e:
Hello @lubos,
here is a last fix for the voxels module.
I forgot to recompute half depth before resolving voxels for transparent objects.
So this is fixed here with the lattest changes I come with for ssrefr.

I'm waiting on you to merge this (possible before next release to avoid yet another broken feature) and then will open the pull request for transparent shadows which is ready.


e2002e:
mall
tokyo
house

PR #3056, opened 2024-10-07
Patch voxels 1.1

e2002e:
Hello @luboslenco,
This pull request fixes screen space refraction, and some inconsistencies with voxelization.


e2002e:
Capture d’écran du 2024-10-10 01-49-57
Capture d’écran du 2024-10-10 01-47-38

PR #3057, opened 2024-10-10
Patch voxels

e2002e:
This pull request fixes issues with screen space refraction, and fixes inconsistencies in voxelization.
I used imageatomic to pack voxels to uint and make use of RGBA64.


e2002e:
Capture d’écran du 2024-10-10 22-06-01

PR #3066, opened 2024-10-28
Add Signals

moisesjpelaez:
This adds a `Signal` class that works similarly to Godot's signals. This can be an alternative to the `Event` class for handling events that need arguments. Also helpful to maintain the code modular.

This is a first iteration but I plan to add new features if they are needed, for now I think it can be merged if you all think this is a good idea.

@luboslenco @QuantumCoderQC @MoritzBrueckner @Onek8


luboslenco:
Thanks!

PR #3100, opened 2024-12-08
OnTimerNode: add false to true case

t3du:
On Timer Node:
-- if repeat is set to false it just does one iteration
-- if repeat is set to true it repeats iterations
-- if repeat is changed from true to false during runtime it stops from doing new iterations
-- but if repeat is changed from false to true during runtime it does not continue to do new iterations
I modified the node with a reactivation function so in that case it can continue to do new iterations.

I think that way is more consistent.

image


luboslenco:
Thanks!

PR #3103, opened 2024-12-16
WriteJsonNode and WriteFileNode: add Html5 version

t3du:
Removed a trace from WriteJson

For html5 now json or file gets downloaded so its not longer lost.


luboslenco:
Thanks!

t3du:
thanks to you too, if could also update examples folder:

https://github.com/armory3d/armory_examples/pull/83

PR #3112, opened 2024-12-21
Html5 Alert, Prompt and Confirm nodes

t3du:
Basic nodes for html5 messages and interaction:

image

I thought these 3 elements could help in a basic way for based game behavior or debugging because: 

- `Alert` shows a message as an ui does (for people who don't know how to get to the console).
- `Prompt` helps to retrieve an user input by keyboard (which is useful in devices that don't have any).
- `Confirm` has a yes or no decision worflow  (interaction with user).

### DIFF
--- a/armory/Sources/armory/logicnode/AlertNode.hx
+++ b/armory/Sources/armory/logicnode/AlertNode.hx
@@ -1,0 +1,17 @@
+package armory.logicnode;
+
+class AlertNode extends LogicNode {
+
+	public function new(tree: LogicTree) {
+		super(tree);
+	}
+
+	override function run(from: Int) {
+
+		#if kha_html5
+		js.Browser.window.alert(inputs[1].get());
+		#end
+
+		runOutput(0);
+	}
+}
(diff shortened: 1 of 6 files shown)


Onek8:
@luboslenco I would like to chime in and ask if you would be interested in adding haxe targets as categories where these 3 nodes could be under "Javascript"? Adding a headless mode as well meaning no canvas in the html file write when targeting the browser, and likely removing all related functions to the canvas rendering with a flag. We would then have a new area of growth for armory and allow things like cli app development as well. An electron runtime would also be interesting, perhaps if the bundle would be too large as an option in the addon section to add/download the files. There is also the option of integrating tauri/dioxus if this is of interest to you as well. It would be nice to work directly with you on Discord if you ever decide to jump on it as it would help coordinate the development with everyone trying to contribute.

luboslenco:
Sounds good! Will try to also hop into discord more.

Onek8:
Oooooh snap!!! 2025 in trouble!!!

PR #3123, opened 2025-01-04
patch 1_2

e2002e:
Hello @luboslenco,
In this pull request:
- patch for voxels (lots of bug fixes and implementation of voxels shadows and voxels refraction)
- fix for screen space refraction (now we can see the sky refracting)
- transparent shadows.
- feature to make invisible objects contribute to the lighting (for top down games where you need invisible ceilings)
- SSGI (I took the freedom to remove the RTAO which is useless, so we go proper screen space effects).


e2002e:
@luboslenco,
This patch is important.
While I understand that it's not acceptable to break CMS, I haven't found a fix for it.
What I am going to do is to make transparent shadows an option, so that the user can choose to enable it (breaking clustered spots and cms) or not.

tindrew (approved):

e2002e:
@luboslenco I'm done here, you can merge if you want !

e2002e:
@luboslenco Here you can see the GI in action, if you look carefully, you can see the stairs reflection's changing level.

Capture d’écran du 2025-04-06 03-28-05

PR #3132, opened 2025-04-10
Patch voxels / refraction + SSGI + transparent shadows

e2002e:
Hello @luboslenco , in this pull request:
- patch for the voxels that had numerous bugs.
- implementation of transparent shadows (the technique comes from a Blizzard's note).
- SSGI. I droped the RTAO which is not working well.

@MoritzBrueckner for the purpose of reviewing, I will go into some details:

VOXELS:
The voxels are working with 4 different 3D images.
First we use "voxels" to write the voxels data (to the scale of the current clipmap) in a 3D grid.
This grid is extended on the X axis to store anisotropy (normal weights), and on the Z axis to store the needed datas (color, normal, environment, and emission).

Then we use the "voxelsOut" and "voxelsOutB" which are  ping-ponged between the  voxels_offset_prev and voxels_temporal, to write to the final voxels grid "voxelsOut", this is a technique to stabilize and clean the grids.

The process of putting together the different clipmaps and the precomputed directions is done in voxels temporal, which also unpacks the data stored on the "voxels" Z axis and the data coming from "voxelsLight" to light and color the image.
"voxelsLight" uses lights colors and shadowmap to modify the grids.
This result in a 3D image that has 16 + 6 times (precomputed directions + anisotropy that we keep for every other conetracing than idffuse and AO) the size of the voxels grid on the X axis, and "clipmapLevel" times the size of the grid on the Y axis.
Z axis is just the size the the voxels grid now that we have unpacked the data and processed it.

The conetracing is done in compute shaders, to optimize the render (and keep it asynchronous). The environment lighting is done here when tracing diffuse or ao. This avoids mixing resolution which results in visible seams.

For transparent objects, the conetracing is done directly, so voxels refraction is always a direct conetracing.

There is a data structure called "clipmaps" in iron now, it hasn't changed since the last pull request.

TRANSPARENT SHADOWS:
This comes from a paper I found on the WickedEngine website, it just has an issue is that it doubles the size taken by shadowmaps, so I got crashes with spots (which are included as an extra when _Spot is defined) and I think it is resource exhaustion (sigsev 11). To deal with that, I made transparent shadows and option. Anyways, it works with atlases for spots.

SSGI:
This is a change I've done lattely, I removed the RTAO because this isn't working well, it traces cones in screen space and unless the distance is one that can get these cones to cross each other without holes, they are visible, so it's not tunable, and we got SSAO working very fine.
SSGI works with basecolor, shadowmaps, and emission buffers. It seeks around the current pixel with some randomness, tests for the depths and applies a normal weight. This works out very nicely.


e2002e:
@MoritzBrueckner
This is done and will not change.
It's either for you to review and ask @luboslenco to merge or maybe you could also test it and merge ?
Either way I'm done working with armory, it's been great !

PR #3145, opened 2025-05-01
Patch 1 3

e2002e:
Hello @luboslenco,
This is a patch for voxels.
I also implemented SSGI and transparent shadows.

The changes regarding voxels are too extensive to list here, but they fix everything.
The SSGI comes from chatgpt.
The transparent shadows comes from a blizzard's note that was detailed by the dev of WickedEngine, so parts of the code come from here, but since I properly implemented it, no copyright is used.


e2002e:
@MoritzBrueckner could you review  ? Thanks.

PR #3151, opened 2025-05-13
Fix debug draw raycast on HL

moisesjpelaez:
This enables debug draw raycast on HL. Thanks to @Onek8 who tested and made the fix.


moisesjpelaez:
Accidentally closed this PR.

PR #3152, opened 2025-05-14
Patch 1 3

e2002e:
Hello @luboslenco,
In this pull request:
- implementation of transparent shadows.
- implementation of ssgi (replaces rtao).
- patch for voxels which had normalization issues (among else).

There is a crash with spot lights and gi because of transparent shadows. The user has to enable atlas.
I think this is a resource exhaustion.


e2002e:
@luboslenco, before you go any forward merging pull request that contain hundreds of commits, could you merge this ?
I am done, this is working.
I don't want to just repeteadly come here to synchronize and fix merge conflicts hoping that one day you'll merge.
After all, this is a gift from me (and remember I kinda stole it).

PR #3164, opened 2025-05-27
fix scene exposure setting

t3du:
Exposure setting was `bpy.data.scenes[0]` pointing always to scene 0 disregarding context scene or play or export scene options so I changed to `arm.utils.get_active_scene()`

The same happens a lot for `bpy.data.worlds['Arm']` is hardcoded world instead of using `arm.utils.get_active_scene().world`. 
@luboslenco is that ok?


luboslenco:
The `bpy.data.worlds['Arm']` is used like a single global storage for the whole project.

PR #3171, opened 2025-05-31
Gi clipmaps

e2002e:
Hello, this is a patch for the voxels.

You need Krom recompiled with my Kinc (which you point to in armorcore).
Kha should also point to my Kinc otherwise you cannot build with C and voxels.


e2002e:
hello @luboslenco,
I'm starting to wonder if you are keeping the broken voxels forever...
I am done here, you should consider merging this before things get too complicated.

PR #3183, opened 2025-06-12
Camera Render Filter

t3du:
`Camera Render Filter` allows to render objects from selected cameras which means only that cameras can see the object. It also affects the object particles.

If an object doesn't have a camera list assigned by default all cameras can render the object:

image

There is a get node for retrieving the camera list array and also a set node for adding or removing a camera from the list array:

image

Here is an example: only 1 scene and 3 cameras. Every camera sees a different combination of objects. But they are all present at the same time in the scene.

https://github.com/user-attachments/assets/ee9e4ce6-7332-474b-97f3-1001e10b1922

### DIFF
--- a/armory/Sources/armory/logicnode/GetCameraRenderFilterNode.hx
+++ b/armory/Sources/armory/logicnode/GetCameraRenderFilterNode.hx
@@ -1,0 +1,19 @@
+package armory.logicnode;
+
+import iron.object.MeshObject;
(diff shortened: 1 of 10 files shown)


t3du:
thanks, last question... the armory examples repo... could you update those too if I add new PRs? There is already one pending... but I could add more examples of the new features.

PR #3225, opened 2025-09-04
Gi clipmaps

e2002e:
Hello @luboslenco,
This pull request patches voxels.

- AABB mechanism
- weighted cosine cone directions aligned with tangent basis
- many bug fixes.


e2002e:
@luboslenco, I actually had to disable the AO in patch_1_3.
This is where i've been working, there is a new ssgi, and transparent shadows.
I think you should consider using the GI instead since everything is calibrated for that.
Nevertheless I will try to rollback the AO and open a pull request.

PR #3226, opened 2025-09-04
Blender 4.5 support

moisesjpelaez:
Support for Blender 4.5, this also supports external blend files and linked scenes.

Thanks to @Onek8 for most of these changes.


moisesjpelaez:
@luboslenco Some shader/material nodes may still need to get updated but I better call it ready for now, to not make this PR grow too much. I'm planning to open more PRs as I find issues/incompatibilities with shader/material nodes.

PR #3231, opened 2025-09-09
OverheadPersonController and minors changes

ObiNoWanKenobi:
This PR includes the bundle OverheadPersonController plus the LN which is similar to FPController, but "overheadCamera" to control the rotation of the "character" with the position of the mouse. 
example 
https://obinowankenobi.github.io/OverheadPersonController01.github.io
Press (v) to desable walk, jump, run and smoothTrack. (b) to reset

OverheadPersonController


luboslenco:
Thanks a lot! Can you also put the code comments into english?

PR #3254, opened 2025-11-29
Replace Object Node

ObiNoWanKenobi:
Hi
This node allows you to exchange the position, rotation and scaling between 2 objects in a simpler way.

Here is an example of the node next to probabilistic.
https://obinowankenobi.github.io/ReplaceObject.github.io/

Captura de pantalla 2025-11-29 180733

by the way @luboslenco
I don't know if it's an error in the latest armory compilation (2509). But it turns out that now I can't access 2 different objects from the same node. That is, if I use the (on volume, on contact, ect) or any node that requires calling 2 different objects that do not have the trait applied, they stop working.
This error does not happen with (24.12), the current version I use after trying with 25.09

The error occurs in both the scene trait and the object trait.

Captura de pantalla 2025-11-29 182130

### DIFF
--- a/armory/blender/arm/logicnode/transform/LN_replace_object.py
+++ b/armory/blender/arm/logicnode/transform/LN_replace_object.py
@@ -1,0 +1,15 @@
+from arm.logicnode.arm_nodes import *
+
+class ReplaceObjectNode(ArmLogicTreeNode):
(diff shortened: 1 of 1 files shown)


luboslenco:
> I don't know if it's an error in the latest armory compilation (2509)

Probably a new error then :(

